PR #2, opened 2024-01-29
Initial polishing of ModelSystem

### LINKED ISSUE #1: Initial polishing of ModelSystem
After our internal Area C meeting (some notes [here](https://github.com/FAIRmat-NFDI/AreaC-Meetings/discussions/26)) I want to clean up and polish the `ModelSystem` definition and the methods.

I will add more tasks, but initially:

- [x] Create a `base_classes_simulation.py` module and move `Computation` there --> I created `general.py`, but I wonder whether we can directly write this in `__init__.py`.
- [x] Rename `Computation` to `Simulation`.
- [x] Improve composition, especially for using methods of classes more freely by users.
- [x] ~~Define a basic metainfo and how to populate it, and create some docs or heuristics for this~~ --> issue #4

JosePizarro3:
Test branch in electronic-parsers: [test_basesections](https://github.com/nomad-coe/electronic-parsers/tree/test_basesections)

Branch in Gitlab to define `Program`, `GeometricalSpace` and `Simulation`: [1728-full-refactoring-of-run-into-data](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/compare/develop...1728-full-refactoring-of-run-into-data)


JosePizarro3:
Hi, these are the changes to be discussed in our meeting. The corresponding branch in nomad is [1728-full-refactoring-of-run-into-data](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/compare/develop...1728-full-refactoring-of-run-into-data).

JosePizarro3:
Feel free to drop a review when you want / can. Focus on the model_system.py please; there is an associate [branch for Wannier90](https://github.com/nomad-coe/electronic-parsers/blob/test_basesections/electronicparsers/wannier90/parser.py) and the corresponding [1721 in Gitlab](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/compare/develop...1728-full-refactoring-of-run-into-data).

ndaelman-hu:
> Feel free to drop a review when you want / can. Focus on the model_system.py please; there is an associate [branch for Wannier90](https://github.com/nomad-coe/electronic-parsers/blob/test_basesections/electronicparsers/wannier90/parser.py) and the corresponding [1721 in Gitlab](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/compare/develop...1728-full-refactoring-of-run-into-data).

I take it that this just an example of how the new schema may affect a parser, correct?
Or is the aim to have this merged asap as well?

JosePizarro3:
> > Feel free to drop a review when you want / can. Focus on the model_system.py please; there is an associate [branch for Wannier90](https://github.com/nomad-coe/electronic-parsers/blob/test_basesections/electronicparsers/wannier90/parser.py) and the corresponding [1721 in Gitlab](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/compare/develop...1728-full-refactoring-of-run-into-data).
> 
> I take it that this just an example of how the new schema may affect a parser, correct? Or is the aim to have this merged asap as well?

This is just an example. I will use some parsers to develop the schema, but we will not merge anything.

ndaelman-hu:
The only problem with reviewing, is that we should (at least once) have the option of reviewing the code in its entirety.
The changes in this particular PR come across as mostly changes in terminology and (at first glance) will likely go through without a problem.
The feedback we were giving in the last meeting (Tuesday 30.01) is directed at older code on `develop`, though.

From what I could find (https://github.com/orgs/community/discussions/9099), there's no such feature in Github?

**Edit**: there is apparently a workaround, but it entails adding comments outside the PR... Still, I think this is appropriate here.
- https://github.com/isaacs/github/issues/284
- https://stackoverflow.com/questions/12696158/comment-on-a-line-in-github-without-a-commit

JosePizarro3:
Cannot you add comments within old code then in a review process?

JosePizarro3:
~Ok, I changed the name of the file to `model_system2.py`. I think now it should be easier to review~

Ah damn, it is recognizing it as system.py anyways. I am going to create a `model_system_review.py` in parallel for you to drop it. **You can skip the `model_system.py` and focus on the _review one.**

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+    This is useful for other sub-sections under ModelSystem. It returns then the corresponding
+    AtomicCell section.
+
+    Args:
+        section (ArchiveSection): The section to check for its parent and AtomicCell.
+
+    Returns:
+        (AtomicCell): The AtomicCell section resolved from the parent.
+    """
+    if section.m_parent is None and section.m_parent.atomic_cell is None:
```
ndaelman-hu: Shouldn't this be `or`?
Moreover, if `m_parent` doesn't exist, the 2nd statement will result in an `AttributeError`.
I suggest splitting this into 2 nested `if`-statements.

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+from nomad.metainfo import Quantity, SubSection, SectionProxy, MEnum
+from nomad.datamodel.metainfo.basesections import System, GeometricSpace
+from nomad.datamodel.metainfo.annotations import ELNAnnotation
+
+
+def check_parent_and_atomic_cell(section, logger):
+    """
+    Checks if the parent of a section exists and whether it has a sub-section atomic_cell.
+    This is useful for other sub-sections under ModelSystem. It returns then the corresponding
+    AtomicCell section.
```
ndaelman-hu: Honestly, this sounds like it should be a function under `ModelSystem`.
Atm, I will get errors when simply applying this method to sections which aren't expected to have any `atomic_cell` whatsoever...

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+    Args:
+        section (ArchiveSection): The section to check for its parent and AtomicCell.
+
+    Returns:
+        (AtomicCell): The AtomicCell section resolved from the parent.
+    """
+    if section.m_parent is None and section.m_parent.atomic_cell is None:
+        logger.error(
+            "Could not find m_parent ModelSystem and its AtomicCell section for "
+            "Symmetry analyzer."
```
ndaelman-hu: Was this Ruff's doing? I'd use block comment here.

Moreover, I'd mark class names as `code`, for clarity.

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+    """
+    A base section used to specify the atomic cell quantities (labels, positions) of a system
+    at a given moment in time.
+    """
+
+    name = Quantity(
+        type=MEnum("original", "primitive", "standard"),
+        description="""
+        Name to identify the cell structure. It might be:
+            - 'original' as in orignally parsed,
```
ndaelman-hu: typo: orignally

I recommend _Code Spell Checker_ to spot these more easily.

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+
+
+class AtomicCell(GeometricSpace):
+    """
+    A base section used to specify the atomic cell quantities (labels, positions) of a system
+    at a given moment in time.
+    """
+
+    name = Quantity(
+        type=MEnum("original", "primitive", "standard"),
```
ndaelman-hu: I know that you didn't pick this naming, but I find it quite weird:

- `primitive` is clear
- does `standard` mean _conventional_? If so, I'd use that term. Tbh, it isn't quite clear from the description.
- what's the purpose of `original`, to perform a conversion to `standard`, afterwards? I feel like this could be removed.
- I'd also add `supercell`, as an alternative to `primitive` and `standard`

I'd also change `name`: this is more a representation type.

Btw, what will you do if we have only a primitive cell, but also want a conventional one? Store it here or create a separate hierarchy?

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+
+    labels = Quantity(
+        type=str,
+        shape=["*"],
+        description="""
+        List containing the labels of the atomic species in the system at the different positions
+        of the structure. It refers to a chemical element as defined in the periodic table,
+        e.g., 'H', 'O', 'Pt'. This quantity is equivalent to `atomic_numbers`.
+        """,
+    )
```
ndaelman-hu: If anything, I'd switch out this quantity for a reference to an `AtomType` section.
You're probably tired of hearing me say this, but I'd like for both `ModelSystem` and `ModelMethod` to rely on this section.

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+
+    atomic_numbers = Quantity(
+        type=np.int32,
+        shape=["*"],
+        description="""
+        List of atomic numbers Z. This quantity is equivalent to `labels`.
+        """,
+    )
```
ndaelman-hu: Same remark as the previous

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+    lattice_vectors_reciprocal = Quantity(
+        type=np.float64,
+        shape=[3, 3],
+        unit="1/meter",
+        description="""
+        Reciprocal lattice vectors of the simulated cell, in Cartesian coordinates and
+        including the $2 pi$ pre-factor. The first index runs over each lattice vector. The
+        second index runs over the $x, y, z$ Cartesian coordinates.
+        """,
+    )
```
ndaelman-hu: We wanted a section detailing the `KMesh`, right? How about moving this quantity there?
I don't really see any simulations that will be exclusively detailed in the k-space... Except maybe phonons?
I'm not sure how well `ModelSystem` fits those, though... We can see to brainstorm for them and (point-centered) vibrational modes.

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+
+    velocities = Quantity(
+        type=np.float64,
+        shape=["*", 3],
+        unit="meter / second",
+        description="""
+        Velocities of the atoms. It is the change in cartesian coordinates of the atom position
+        with time.
+        """,
+    )
```
ndaelman-hu: Would move `velocities` under `positions`.

I understand that `forces` / `acceleration`, meanwhile, fall under `output`, but my brain automatically completes the list...
Thinking from an outsider's perspective, I prefer to group those here.
Note that `velocities` are only input at the 1st position, the rest is technically also `output` @JFRudzinski ?

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+    supercell_matrix = Quantity(
+        type=np.int32,
+        shape=[3, 3],
+        description="""
+        Specifies the matrix that transforms the primitive unit cell into the supercell in
+        which the actual calculation is performed. In the easiest example, it is a diagonal
+        matrix whose elements multiply the `lattice_vectors`, e.g., [[3, 0, 0], [0, 3, 0], [0, 0, 3]]
+        is a $3 x 3 x 3$ superlattice.
+        """,
+    )
```
ndaelman-hu: This should be placed under `name`. When would you use this value?
Do you feel that we should always report the initial system? If we have interoperable forms, wouldn't that be also FAIR?

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+
+    wyckoff_letters = Quantity(
+        type=str,
+        shape=["*"],
+        # TODO improve description
+        description="""
+        Wyckoff letters associated with each atom position.
+        """,
```
ndaelman-hu: Aren't both rather part of the symmetry analysis?

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+    )
+
+    def to_ase_atoms(self, logger):
+        """
+        Generates a ASE Atoms object with the most basic information from the parsed AtomicCell
```
ndaelman-hu: type: generate **an** ASE [...]

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+                    ase.data.chemical_symbols[number] for number in atomic_numbers
+                ]
+            except IndexError:
+                logger.error(
+                    "The AtomicCell.atomic_numbers are out of range of the periodic table."
+                )
+                return
+        self.labels = atomic_labels
+
+        # We will use ASE Atoms functionalities to extract information about the AtomicCell
```
ndaelman-hu: Pet peeve, but I'd avoid personal forms for comments, since they come across as didactic. Imperatives will do.

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+                ModelSystem.atomic_cell, primitive_atomic_cell
+            )
+            self.m_parent.m_add_sub_section(
+                ModelSystem.atomic_cell, standard_atomic_cell
+            )
+            # Reference to the standarized cell, and if not, fallback to the originally parsed one
+            self.atomic_cell_ref = self.m_parent.atomic_cell[-1]
+
+
+class ChemicalFormula(ArchiveSection):
```
ndaelman-hu: Am happy with the shorter names, but... I'd like to support advanced text searches (also mentioned this in my FAIRmat Café, but you were on holidays back then).

Those need a list of the chemical "subunits", from which it can easily compose any kind of query. This serves only NOMAD GUI/API queries. Their advantage is that we could support a much wider variety of formulas, without the need of reprocessing, as well as substructure matching. The latter is already the standard in chemistry. **Composition queries are one of the most common types. We can't permit ourselves to use an outdated system here.**

I'm unsure if we should store them in `data`, though. With this quantity, the rest would become mostly redundant, but having them is fine.
The best approach for extending this to the JSON, would be to add the requested formulas when performing the API call, but this is a very advanced use case. @lauri-codes ?

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+        electronegativity as defined in the IUPAC nomenclature of inorganic chemistry (2005):
+
+            - https://en.wikipedia.org/wiki/List_of_inorganic_compounds
+
+        Contains reduced integer chemical proportion numbers where the proportion number
+        is omitted if it is 1.
+        """,
+    )
+
+    hill = Quantity(
```
ndaelman-hu: Capital letter?

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+
+        Returns:
+            ase.Atoms: The ASE Atoms object with the basic information from the AtomicCell.
+        """
+        # Initialize ase.Atoms object with labels
+        ase_atoms = ase.Atoms(symbols=self.labels)
+
+        # PBC
+        if self.periodic_boundary_conditions is None:
+            logger.info("Could not find AtomicCell.periodic_boundary_conditions.")
```
ndaelman-hu: Also state that it defaults to no PBC.

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+        type=np.int32,
+        shape=["*"],
+        description="""
+        List of atomic numbers Z. This quantity is equivalent to `labels`.
+        """,
+    )
+
+    positions = Quantity(
+        type=np.float64,
+        shape=["*", 3],
```
ndaelman-hu: So we decided on removing the sanity check?

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+        ase_atoms = ase.Atoms(symbols=self.labels)
+
+        # PBC
+        if self.periodic_boundary_conditions is None:
+            logger.info("Could not find AtomicCell.periodic_boundary_conditions.")
+            self.periodic_boundary_conditions = [False, False, False]
+        ase_atoms.set_pbc(self.periodic_boundary_conditions)
+
+        # Positions
+        if self.positions is not None:
```
ndaelman-hu: I think that `self.positions` should default to `[]`. If so, then you can remove this line.

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+                    2 * np.pi * ase_atoms.get_reciprocal_cell() / ureg.angstrom
+                )
+        else:
+            logger.info("Could not find AtomicCell.lattice_vectors.")
+
+        return ase_atoms
+
+    def normalize(self, archive, logger):
+        # Check if AtomicCell section exists
+        if self is None:
```
ndaelman-hu: Huh?

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,865 @@
...
+            logger.error(
+                "Could not find the basic ModelSystem.atomic_cell information."
+            )
+            return
+
+        # If the labels and atomic_numbers are not specified, we return with an error
+        if self.labels is None and self.atomic_numbers is None:
```
ndaelman-hu: I think that you can recover from this error if at least one of both is given...
If neither are, then you may return.

ndaelman-hu (commented):
Am posting an intermittent review (didn't get around to everything yet).

Overall the migration is quite a bit cleaner good work! I have a bunch of comments, but that's normal at this size.
Take special care with the normalizers: we will have to test them thoroughly.

2 specials notes:
1. Split big functions into smaller pieces.
2. Apply type annotation for the functions: it'll help in the long run.

simulationdataschema/model_system.py
```diff
@@ -76,16 +97,9 @@ class AtomicCell(RealSpace):
...
     )
 
-    n_atoms = Quantity(
-        type=np.int32,
-        description="""
-        The total number of atoms in the system.
-        """,
-    )
-
     labels = Quantity(
```
JFRudzinski, line 234: I think we should consider adding an additional quantity in terms of labeling atoms and then change the naming a bit. IMO there should be 2 identifying quantities:
1. elements: renaming what is now "labels". If this quantity refers to the chemical elements from the periodic table as the description states, we should name it that for clarity.
2. labels (new): this is an arbitrary/custom name given to each particle. Allows atoms of the same type to be labeled differently for appropriate applications. For standard atomic systems, this could either be identical to elements (so that we don't need to worry about downstream issues) or simply not populated and then functionalities which are using labels currently need to use elements by default and labels secondary (or some protocol). For non-standard systems or multi-scale descriptions, elements would simply not be populated ('X' is used currently for unknown elements), and labels can be set accordingly. Then there is some consistency between these 2 cases.

simulationdataschema/model_system.py
```diff
@@ -173,106 +187,95 @@ class AtomicCell(RealSpace):
...
+            )
             return
-        if len(atom_positions) != len(atom_labels):
+
+        # If the labels and atomic_numbers are not specified, we return with an error
+        if self.labels is None and self.atomic_numbers is None:
             logger.error(
-                "Length of System.atomic_cell.positions does not coincide with the length "
-                "of the System.atomic_cell.labels."
+                "Could not read parsed AtomicCell.labels or AtomicCell.atomic_positions."
```
JFRudzinski: "or AtomicCell.atomic_numbers"?

simulationdataschema/model_system.py
```diff
@@ -528,10 +571,134 @@ class ModelSystem(System):
...
+    """
+    Model system used as an input for the computation. It inherits from `System` where a set
+    of sub-sections for the `elemental_composition` is defined.
+
+    We also defined:
+        - `name` refers to all the verbose and user-dependent naming in ModelSystem,
+        - `type` refers to the type of the ModelSystem (atom, bulk, surface, etc.),
+        - `dimensionality` refers to the dimensionality of the ModelSystem (0D, 1D, 2D, 3D),
+
+    If the ModelSystem `is_representative`, the normalization occurs. The time evolution of
```
JFRudzinski, line 773: So did we decide that only 1 system can be representative? Do we need a check for this if so?

simulationdataschema/model_system.py
```diff
@@ -551,14 +733,15 @@ class ModelSystem(System):
         """,
     )
 
-    # TODO what about `branch_index`?
+    # TODO what about `branch_index` or `branch_depth`?
```
JFRudzinski: I like index better than depth, personally, as it implies more that it is an integer. Maybe in the description it should be stated that it is 0-indexed? (or what the 0-index is/means)

simulationdataschema/model_system_review.py
JFRudzinski: I could be wrong but I think original means whatever was stored in the raw file. So, there is necessarily a duplicate of the primitive or standard. I remember asking Laurie about this at some point cause it also seemed weird to me. Don't remember if it was completely necessary or there accidentally/historically

simulationdataschema/model_system_review.py
JFRudzinski: Yeah I agree, I have always felt this way. I understand why someone would place forces under output but in practice I think it makes more sense with positions. Output is more for collective properties. Are there any other properties besides forces and acceleration that are stored as a function on individual atoms in calculation?

For the MD parsers it is actually sort of a pain because forces are always kept with positions and velocities in the same file, while the other output is stored elsewhere or parsed separately.

JFRudzinski (commented):
@JosePizarro3 Looks good overall. Just a few minor comments. fyi - I accidentally looked at  the original ModelSystem file and put some comments there first. Sorry.

JosePizarro3:
Thanks a lot for the reviews! 😄 

@ndaelman-hu should I wait for you to finish? I am just curious if you will continue with it or do you prefer me to start with some changes from your suggestions.

ndaelman-hu:
> Thanks a lot for the reviews! 😄
> 
> @ndaelman-hu should I wait for you to finish? I am just curious if you will continue with it or do you prefer me to start with some changes from your suggestions.

@JosePizarro3 No, feel free to go ahead. I pretty much covered `AtomicCell`, the rest only partially.
Today, I likely won't get around to it, anyway, at best tomorrow. I can always add another review later on, but I think there's already enough, for now.

simulationdataschema/model_system_review.py
JosePizarro3: I think not, because this checks that whatever the `section` is, if it has a m_parent and this m_parent has an atomic_cell.

In any case, I think I can extend this and use it independently as an utility function (passing also the sub_section I want to resolve from the parent). It is a bit patchy, but might also be useful when going into some ladder of sections/sub-sections.

simulationdataschema/model_system_review.py
JosePizarro3: I tried that already, but it works very badly in the current nomad code. In any case, thanks for spotting the typo 🙂

simulationdataschema/model_system_review.py
JosePizarro3: Indeed, `"original"` refers to the originally parsed from the raw file. Then, it is duplicated by MatID.

I can change `"standard"` by `"conventional"`, no problem. I can also add `"supercell"`, but at this point we don't have examples. Maybe this is something to have in mind, as this quantity could be extended.

About `name` to `type`, I am not sure. I can change it, but we have to come with a consistent idea behind `name`, `type`, `label`,...

simulationdataschema/model_system_review.py
JosePizarro3: Ok, I will add a TODO.

simulationdataschema/model_system_review.py
JosePizarro3: > Would move velocities under positions.

Do you mean simply moving the quantity def up after `positions` or creating a new class `Positions` which contains both?

simulationdataschema/model_system_review.py
JosePizarro3: Ok, I see your point in `name` now.

I think one could decide to store this matrix as a way of defining their supercells, as an addition to positions and lattice_vectors. This is a very typical way of defining supercells.

This could also help in normalizing back from a given primitive cell into the original one, as maybe there is a way of finding the matrix. Maybe we should talk with Lauri about this.

simulationdataschema/model_system_review.py
ndaelman-hu: The former. It's because the `metainfo` / `data` will follow the order in which we declare the schema quantities, so I would group common kinds together.

simulationdataschema/model_system_review.py
JosePizarro3: They are, but they refer to the atomic cell. If you see the Symmetry normalization you will see that there these primitive and conventional cells are being generated.

simulationdataschema/model_system_review.py
ndaelman-hu: > Are there any other properties besides forces and acceleration that are stored as a function on individual atoms in calculation?

I can only think of derived tensors, e.g. stress and strain.

simulationdataschema/model_system_review.py
JosePizarro3: I don't think is necessary.

simulationdataschema/model_system_review.py
ndaelman-hu: True, a lot of function names or variables also get marked

simulationdataschema/model_system_review.py
JosePizarro3: I can recover it, but I don't see it very necessary for now.

simulationdataschema/model_system_review.py
JosePizarro3: I disagree: the positions have to be parsed

simulationdataschema/model_system_review.py
JosePizarro3: previous version code 😄 thanks for spotting it.

simulationdataschema/model_system_review.py
ndaelman-hu: Thinking further on it, and given the usage, I'd **replace this function with a `set_atomic_cell(self, logger) -> None`** that (iteratively or recursively) searches its parent tree until it finds and `atomic_cell` and then takes it over.
If necessary, we can also add fallbacks in case there are none.

Let me know if you want a hand with implementing this.

simulationdataschema/model_system_review.py
JosePizarro3: Yes, that's why the and. Later I check one quantity to resolve the other one.

simulationdataschema/model_system.py
JosePizarro3: We can work more in this. For now, the `ModelSystem.normalize()` (all the symmetry and chemical formula extraction) is only run if `is_representative` is true. And this is so far set to the last ModelSystem (in the `Simulation.normalize()`).

simulationdataschema/model_system.py
JosePizarro3: What about `tree` or `branch`?

simulationdataschema/model_system_review.py
JosePizarro3: Ok, I split them into some other if checks.

simulationdataschema/model_system.py
ndaelman-hu: Well what does it represent? When I hear

- _branch_ (or _tree_ when further splitting is incorporated), I think of the entity as a whole
- _index_, I think of horizontal positioning, either at that depth or wrt to the root reference
- _depth_ sounds like an `int` for vertical positioning

The description "Index refering to the depth of a branch in the system tree." matches mostly the last option.

simulationdataschema/model_system_review.py
JosePizarro3: I think not. The goal of this method is to do some seesaw movement in `section` to return `sub_section`, i.e., one section up, one section down. It does not require that we are using `AtomicCell`, but could be useful in the future when dealing with this entangled sections and normalizations.

I will define `utils/utils.py` and add this as `get_sub_section_from_section_parent(section, sub_section, logger)` (or `seesaw_section_to_sub_section(section, sub_section, logger)`). Feel free to suggest better naming.

ndaelman-hu: I'm not following... I don't think you mean depth navigation, but rather an order of normalizer execution?
My suggestion was based on its current usage. Maybe it's faster if you clarify in person...

For actual navigation (during construction), I'm building something out in a separate branch.

simulationdataschema/model_system.py
JosePizarro3: Resolving this in favor of #5

simulationdataschema/model_system_review.py
JosePizarro3: I added a TODO, and I am resolving in favor of #5

simulationdataschema/model_system_review.py
JosePizarro3: #5

JosePizarro3:
Hi @JFRudzinski @ndaelman-hu 

I added new changes. I am requesting a final review once more. Bear in mind the TODOs, and feel free to resolve your suggestions if you feel satisfied.

JFRudzinski (approved):
Yeah looks good to me for now :+1:

simulationdataschema/model_system_review.py
ndaelman-hu: I don't really care about a specific case, but more so the _policy_. For sure, we want / need sanity checks, simply by virtue of normalization, we just need to settle on guidelines. For the scope of this PR, you can leave it as is, and we continue this in a separate discussion: https://github.com/FAIRmat-NFDI/AreaC-Meetings/discussions/32

simulationdataschema/model_system_review.py
ndaelman-hu: @JosePizarro3 You can apply the move and resolve this comment. This should be pretty non-controversial.

Given that @JFRudzinski agrees with adding `forces`, you can go ahead there too.
We can hold off on strain and stress for now.

simulationdataschema/model_system_review.py
ndaelman-hu: Ok. Just add an in-line comment, e.g. `# ensure that positions are parsed`.

simulationdataschema/model_system_review.py
ndaelman-hu: Migrated this to a discussion: https://github.com/FAIRmat-NFDI/AreaC-Meetings/discussions/33

simulationdataschema/utils/utils.py
```diff
@@ -0,0 +1,56 @@
...
+    section: ArchiveSection, sub_section: str, logger
+) -> ArchiveSection:
+    """
+    Gets the sub_section of a section by performing a seesaw move, going to the parent of
+    section, and then going down to sub_section. Example, if section is `Symmetry`, and we
+    want to resolve `AtomicCell` (sub_section), this methods goes up to `ModelSystem` from `Symmetry`,
+    and then goes down to `AtomicCell`.
+
+    If the sub_section is a list, it returns the first element of the list. If the sub_section is
+    a single section, it returns the section.
```
ndaelman-hu: If it's a "seesaw" move, you don't end up with a `sub_section` of said section, but a `sibling_section`... This should be changed in the name.

I'd rewrite the docstring as follows:
```
Fetches a sibling section, i.e. subsection a common parent section.
If the sibling section is repeating, only the first element is returned.

Example section structure, where `sibling_section` is targeted:
parent_section
 |_ sister_section
 |_ starting_section
 |_ sibling_section
    |_ grandchild_section
 |_ brother_section
```

ndaelman-hu: Normally, one would expect the whole list to be returned. Can you explain the use cases you have in mind?
You could also add a `bool` flag.

simulationdataschema/utils/utils.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Args:
+        section (ArchiveSection): The section to check for its parent and retrieve the sub_section.
+        sub_section (str): The name of the sub_section to retrieve from the parent.
```
ndaelman-hu: Missing `logger`

simulationdataschema/utils/utils.py
```diff
@@ -0,0 +1,56 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from nomad.datamodel.data import ArchiveSection
+
+
+def get_sub_section_from_section_parent(
+    section: ArchiveSection, sub_section: str, logger
+) -> ArchiveSection:
```
ndaelman-hu: Python's type checking is optional (= retarded), so he didn't flag this:

```
def get_sub_section_from_section_parent(
    section: ArchiveSection, sub_section: str, logger: logging.Logger
) -> Optional[ArchiveSection]:
```

simulationdataschema/utils/utils.py
```diff
@@ -0,0 +1,56 @@
...
+    Args:
+        section (ArchiveSection): The section to check for its parent and retrieve the sub_section.
+        sub_section (str): The name of the sub_section to retrieve from the parent.
+
+    Returns:
+        sub_section_sec (ArchiveSection): The sub_section to be returned.
+    """
+    if section.m_parent is None:
+        logger.error("Could not find the parent of the section.")
+        return
```
ndaelman-hu: It's kind of strange to see an error be dismissed by a simple `return`. Since this non-fatal, I'd throw a `warning`.

simulationdataschema/utils/utils.py
ndaelman-hu: Don't forget to import `logging`

simulationdataschema/model_system_review.py
ndaelman-hu: Have added my review of your new implementation. You should be able to see it soon.

simulationdataschema/model_system_review.py
ndaelman-hu: > I think one could decide to store this matrix as a way of defining their supercells, as an addition to positions and lattice_vectors. This is a very typical way of defining supercells.

Indeed, but then we already need to have a primitive / conventional cell to be defined.

simulationdataschema/model_system_review.py
ndaelman-hu: Just checked it out. That's some very strange approach. 2 points:

1. Since you are accessing `AtomicCell` anyhow, you can just generate Wyckoff on the fly for `Symmetry`. That is to say: the execution does not determine the section structure here. You can just shove the part of 1 normalizer to the next. Thematically, Wyckoff is Symmetry, period.
2. I'm even more convinced now that we should leave out `original` cell, as it isn't used in other normalizers. We should just extract it during parsing, then have functions that convert it to `primitive` or `conventional`, which we do store. We can discuss the actual implementation in-person.

simulationdataschema/model_system.py
ndaelman-hu: Let's branch this out to its own discussion: https://github.com/FAIRmat-NFDI/AreaC-Meetings/discussions/34

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,864 @@
...
+        type=str,
+        description="""
+        Hall symbol for this system describing the minimum number of symmetry operations,
+        in the form of Seitz matrices, needed to uniquely define a space group. See
+        https://cci.lbl.gov/sginfo/hall_symbols.html. Examples:
+            - `F -4 2 3`,
+            - `-P 4 2`,
+            - `-F 4 2 3`.
+        """,
+    )
```
ndaelman-hu: > in the form of Seitz matrices

is a bit misleading, since it sounds like we're storing those matrices... I'd just remove it.

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,864 @@
...
+            - `4/mmm`,
+            - `m-3m`.
+        """,
+    )
+
+    space_group_number = Quantity(
+        type=np.int32,
+        description="""
+        Specifies the International Union of Crystallography (IUC) space group number of the 3D
+        space group of this system. See https://en.wikipedia.org/wiki/List_of_space_groups.
```
ndaelman-hu: Can we have a better reference than wikipedia?
How about one of the links there to the International Union of Crystallographers: https://it.iucr.org/

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,864 @@
...
+        type=str,
+        description="""
+        Specifies the International Union of Crystallography (IUC) space group symbol of the 3D
+        space group of this system. See https://en.wikipedia.org/wiki/List_of_space_groups.
+        Examples:
+            - `F-43m`,
+            - `P4/mmm`,
+            - `Fm-3m`.
+        """,
+    )
```
ndaelman-hu: Note: this will also benefit from a proper search system.

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,864 @@
...
+
+    atomic_cell_ref = Quantity(
+        type=AtomicCell,
+        description="""
+        Reference to the AtomicCell section that the symmetry refers to.
+        """,
+        a_eln=ELNAnnotation(component="ReferenceEditQuantity"),
+    )
```
ndaelman-hu: Is this still necessary in our new design?

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,864 @@
...
+        system = getattr(symmetry_analyzer, f"get_{cell_type}_system")()
+
+        positions = system.get_scaled_positions()
+        cell = system.get_cell()
+        atomic_numbers = system.get_atomic_numbers()
+        labels = system.get_chemical_symbols()
+
+        atomic_cell = AtomicCell(type=cell_type)
+        atomic_cell.lattice_vectors = cell * ureg.angstrom
+        atomic_cell.positions = positions * ureg.angstrom
```
ndaelman-hu: Why this conversion? Shouldn't the parser stipulate the units?

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,864 @@
...
+        return atomic_cell
+
+    def resolve_bulk_symmetry(
+        self, original_atomic_cell: AtomicCell, logger: BoundLogger
+    ) -> Tuple[Union[AtomicCell, None], Union[AtomicCell, None]]:
+        """
+        Resolves the symmetry of the material being simulated using MatID and the
+        originally parsed data under original_atomic_cell. It generates two other
+        `AtomicCell` sections (the primitive and standarized cells), as well as populating
+        the `Symmetry` section.
```
ndaelman-hu: Does this mean that I cannot set the `Symmetry` from the parser?
Many codes perform their own symmetry analysis. Are we just going to throw that out?

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,864 @@
...
+                conventional_num, conventional_wyckoff
+            )
+            aflow_prototype = search_aflow_prototype(
+                symmetry.get("space_group_number"), norm_wyckoff
+            )
+            strukturbericht = aflow_prototype.get("Strukturbericht Designation")
+            if strukturbericht == "None":
+                strukturbericht = None
+            else:
+                strukturbericht = re.sub("[$_{}]", "", strukturbericht)
```
ndaelman-hu: Isn't this the same as

```
if strukturbericht != None:
    strukturbericht = re.sub("[$_{}]", "", strukturbericht)
```

simulationdataschema/model_system_review.py
JosePizarro3: Ok, I moved `velocities` after `positions`, but before defining `forces`, let's think a bit more about `AtomType` or `Atoms`. I wrote a TODO

ndaelman-hu (commented):
Okay, I went through it till the end. Will still have to go through it with a finer comb.
I have the feeling that we are repeating a lot of data structures, e.g. various types of `AtomicCell` and too many reformulations of the same symmetries / formulas in our schema.
I think we can make this leaner, but we'd need to sit with @lauri-codes first.

@JosePizarro3 just resolve the open threads and you can merge. I think this is fine for now.
Several ideas have come up that we should investigate a bit before proceeding further here.
If you find other topics that require discussion, feel free to open a thread over at AreaCMeetings.

simulationdataschema/model_system_review.py
ndaelman-hu: Agreed. We can even have more, as the GUI representation (i.e. what you see when under a section listing in DATA, e.g. https://nomad-lab.eu/prod/v1/gui/search/entries/entry/id/0kWQ2bZTvWKlV_I-_KnC-ePkLJCc/data/results/material/elemental_composition/0) can be set via `label_quantity` in `m_def`.

simulationdataschema/model_system_review.py
JosePizarro3: I think we discussed this already. In the previous schema, we had Symmetry.system_original, Symmetry.system_primitive and Symmetry.system_std.

I decided to move it out and create a repeated list of AtomicCell instead, to avoid potential issues on where the representative system is pointing (some times it was Atoms[0], some other times it was Symmetry.system_std[0]).

Now, for 1, that's is the main reason I am generating new AtomicCell sections (outside of Symmetry); this is not strange at all (at least not wrt to the previous way). But still, we need the SymmetryAnalyzer for generating this extra cells, that's why the method is appearing under Symmetry.

In any case, I didn't delve into the details of the SymmetryAnalyzer, but I know it is generating 2 cells from the originally parsed one. But this is an optional extraction; it might be that the primitive and conventional cells cannot be extracted using that package... thus we need to keep `"original"` as a fallback. Otherwise yes, the original will be redundant.

simulationdataschema/utils/utils.py
JosePizarro3: Well, I would like to return one section. This is because otherwise we would have to deal with `isinstance` outside this method

JosePizarro3: Done. I changed the name to `get_sibling_section`

simulationdataschema/model_system.py
JosePizarro3: Using now `branch_label` and `branch_depth`

simulationdataschema/model_system_review.py
JosePizarro3: I think we should leave more things as ELN annotations, but I didn't tackle it fully yet.

simulationdataschema/model_system_review.py
JosePizarro3: This is coming from SymmetryAnalyzer

simulationdataschema/model_system_review.py
JosePizarro3: No, you can populate it if you want. Then the Symmetry.normalize() will not be run.

simulationdataschema/model_system_review.py
JosePizarro3: No, note that "None" is a string...

simulationdataschema/model_system_review.py
JosePizarro3: I think is pretty find the wikipedia here.

simulationdataschema/utils/utils.py
ndaelman-hu: But is that generic enough? It sounds like for now you just want data from something that's representative, but this might also be used to apply normalization to all siblings, in which case you do want the `list`.

Don't forget that functions in `utils` are supposed to be very generic. Too much specialization for the sake of a bit of convenience will kill the function's utility.

Anyhow, I think that for now you can just add an option parameter and have it default to your preference.
We can revisit the function scope later on.

simulationdataschema/model_system_review.py
ndaelman-hu: Sorry for the typo. My comment still stands, though, since `None` is the default for an empty `Quantity`.

simulationdataschema/model_system_review.py
ndaelman-hu: Normalization always runs, since NOMAD calls it explicitly.

simulationdataschema/utils/utils.py
JosePizarro3: > But is that generic enough? 

Yes

> It sounds like for now you just want data from something that's representative

No, I only want to retrieve one of the ArchiveSections. I can add another argument as the element of the list `i_sibling` which defaults to the first, 0.

simulationdataschema/model_system_review.py
ndaelman-hu: There's something wrong if the `SymmetryAnalyzer` requires a fixed set of units (I reckon its precision)... Please add a note that we investigate this.

simulationdataschema/model_system_review.py
ndaelman-hu: I'm completely fine with it being an ELN section. I more so meant the reference part...
This ties in with the multiple `AtomicCell`s. I think we can have a single one that explains its full `Symmetry`, but that's what I'd like to discuss with Lauri.

Just leave a note about the reference.

simulationdataschema/model_system_review.py
JosePizarro3: True, I fixed it

simulationdataschema/model_system_review.py
ndaelman-hu: ... While I don't have anything against Wikipedia in and of itself, it has no place in a scientific publication or product.
It just looks bad. Pls swap it with the other reference.

simulationdataschema/model_system_review.py
JosePizarro3: Then it is a matter of deciding how to normalize this more consistently. In my case, I defined by hand normalizations, which I don't know how it differs from the fact of having a `SystemNormalizer`, etc. I will talk with Alvin and Lauri about this.

JosePizarro3: > In my case, I defined by hand normalizations

(which means that only the line calling for normalize() under `ModelSystem.normalize()` is ran)

ndaelman-hu: Ow, I remember... No, **that's a flawed way of working**: that part of the code should never be modified by any plugin (dev), including those that we produce. The `normalize` interface is a contract between NOMAD and the plugins. You can't modify that contract, but you can refuse to sign it, i.e. we build our own alternative to `normalize`, e.g. `standardize`, if you need special control. Then we have `normalize` call `standardize`.

If you need more control over normalization, discuss it with @markus1978 .

ndaelman-hu: You see, by modifying the interface one-sidedly, you break the contract and subvert the expectations of all plugin devs.
I can no longer think about our own plugin, because of this, since somebody modified the contract "behind my back". Now imagine that everyone starts doing this...

simulationdataschema/model_system_review.py
ndaelman-hu: > I decided to move it out and create a repeated list of AtomicCell instead, to avoid potential issues on where the representative system is pointing (some times it was Atoms[0], some other times it was Symmetry.system_std[0]).

Yeah, I see that now (mentioned it also in another thread). For now, it's fine, but imho, we should just store the primitive (original in case of no symmetry) cell with denotations of the symmetry operations, and a supercell matrix if that applies to the original. Conventional cells or any other kind of supercell can be produced on the fly, that isn't so hard. That's what I want to look at with @lauri-codes 

> we need the SymmetryAnalyzer for generating this extra cells, that's why the method is appearing under Symmetry.

Sure, but Wyckoff can still go under Symmetry (where it belongs) and then be extracted via the `SymmetryAnalyzer` that you call regardless...

simulationdataschema/utils/utils.py
ndaelman-hu: Good addition. You're almost there:

`index_sibling: Optional[int] = 0,` where `None` stands for the full list.

Tbh, this solution is still very ad-hoc, but with this last addition, you can manage (for now).
This is a symptom of us lacking a proper generic search functionality. Let me see with Area D, how we can extend `MSection` with my `Hierearchy` design.

simulationdataschema/model_system_review.py
ndaelman-hu:
```
if section.m_def.name == 'Simulation' and isinstance(section, EntryData):
                self.normalize_section(section, logger)
                sub_sections = [sub_section for sub_section in section.m_contents()]
                for sub_section in sub_sections:
                    if sub_section.m_def.name == 'ModelSystem':
                        self.normalize_section(sub_section.atomic_cell[0], logger)
                        self.normalize_section(sub_section, logger)
                    elif sub_section.m_def.inherited_sections[2].name == 'TB':
                        self.normalize_section(sub_section, logger)
                        self.normalize_section(sub_section.k_mesh, logger)
```

This entire `if`-section in `metainfo.py` just recognizes our schema.
This logic has to be moved to `ModelSystem` and `TB` respectively, which then call subsections via `standardize`.

simulationdataschema/model_system_review.py
```diff
@@ -0,0 +1,873 @@
...
+                    cluster_threshold=config.normalize.cluster_threshold,
+                )
+                cls = classifier.classify(ase_atoms)
+            except Exception as e:
+                self.logger.warning(
+                    "MatID system classification failed.", exc_info=e, error=str(e)
+                )
+                return system_type, dimensionality
+
+            classification = type(cls)
```
ndaelman-hu: Just spotted this:

be careful with `type`, it's both very powerful and a bit of an unreliable tool. It holds a special place in Python (it's the only "real" class, etc.).
A very typical feedback from code checkers like Copilot would be to swap it with `isinstance`, which behaves more in line with what you want here. Anyhow, the best way is to check with a test.

Moreover, I'd avoid the use of the variable `cls`, since it is often used in a similar context as `self`.
A pydev's first instinct, when seeing it, is to check for `__new__()`. Just give it any other name (except `class` :p).

ndaelman-hu:
@JosePizarro3 is there a branch or commit with the up-to-date base section definitions?
The one you posted at the head (https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/compare/develop...1728-full-refactoring-of-run-into-data) is missing `Computation`, among others...

Also consider my advice on the git issue, to move our base classes for the time being into this project.
These are subject to change, as we are still in the process of designing and as such we will keep them at the same version.

simulationdataschema/utils/utils.py
```diff
@@ -0,0 +1,79 @@
...
+    Returns:
+        sibling_section (ArchiveSection): The sibling_section to be returned.
+    """
+    # Check if the parent of section exists
+    if section.m_parent is None:
+        logger.warning("Could not find the parent of the section.")
+        return
+
+    # Check if the sibling_section exists in the parent of section
+    if not section.m_parent.m_xpath(sibling_section_name):
```
ndaelman-hu: **Minor note**, but I'd use `hasattr` here for better **performance**: `m.x_path` converts the archive into a json which is then equipped with a powerful search interface. That's great for data scientists manipulating downloaded data, but here it's overkill. Don't forget that such conversions and queries are time-consuming.

JosePizarro3:
Ok, I am merging this after the last changes. 

Thanks a lot for the feedback! I will be working in `ModelMethod` in a new branch, and then we can revisit this section `ModelSystem`. Probably along the way we can solve #5

ndaelman-hu:
@JosePizarro3 Could we take a look at the Gitlab branch before moving on to `ModelSystem`?
I raised some suggestions that will facilitate us adding testing and generally working all 3 on this project.

JosePizarro3:
> @JosePizarro3 Could we take a look at the Gitlab branch before moving on to `ModelSystem`? I raised some suggestions that will facilitate us adding testing and generally working all 3 on this project.

Yeah, the branch is in gitlab.

PR #6, opened 2024-02-09
Fix testing

ndaelman-hu:
Minor changes to enable testing (when run in the dev setup from the NOMAD project folder `dependencies/schema/simulation/data`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,3 +3,17 @@
 This is a plugin for [NOMAD](https://nomad-lab.eu/nomad-lab/) which contains the base sections schema definitions for materials science simulations.
 
 <!--Add Getting started section to install the plugin independently of NOMAD for local and Oasis usage (see e.g. https://github.com/FAIRmat-NFDI/nomad-measurements?tab=readme-ov-file)-->
+
+## Importing this plug-in into NOMAD
+
+### For Developers
+
+In here we will use the term _local path_ to the schema to refer to the root of this project.
+Under the default naming convention, this would be `"dependencies/schema/simulation/data"` with respect to the NOMAD root.
+
+1. Update dependency files:
+   1. `pyproject.toml`: add the local path under `[tool.setuptools.packages.find]`.
+   2. `MANIFEST.in`: link the local path to this `README.md` file, as well as the `*.py` and `nomad_plugin.yaml` files in the source folder `simulationdataschema`.
+   3. `.gitmodules`: add the local path and this project's [URL](https://github.com/nomad-coe/nomad-schema-plugin-simulation-data) to check out the relevant commit.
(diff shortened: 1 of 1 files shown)


JosePizarro3:
What is the issue here? I think the relative imports, especially in `__init__.py` were correct. See simulation/run e.g.

@ladinesa, any idea of this?

ndaelman-hu:
> What is the issue here? I think the relative imports, especially in `__init__.py` were correct. See simulation/run e.g.
> 
> @ladinesa, any idea of this?

They weren't working for me. Even after fixing this, there's still a problem with `matid.classifications`. Am looking into it now.

Getting relative imports to work is a bit tricky, especially with nested projects. If you cross-reference with **other dependencies,** you'll see that **they also follow this kind of import statement**, i.e. relative to their own project root. That is like an absolute path from their perspective and thus robuster.

ladinesa:
> What is the issue here? I think the relative imports, especially in `__init__.py` were correct. See simulation/run e.g.
> 
> @ladinesa, any idea of this?

sorry I cannot follow the problem. In your simulationdata/init.py you have  ```from .system import ModelSystem```, so in principle your can do ```from simulationdata import ModelSystem```, ```from simulationdata.system import Model``` won't work unles you have ```from . import system```

ndaelman-hu:
@JosePizarro3 Fixed the matid import, but hit another dead end. I either use your branch 1728, but it hasn't been rebased to today's changes yet.

Moreover, I also suggest that for the time being we keep the classes that you introduce there in this plugin (for now, while we're generating a structure), unless it really breaks something.

JosePizarro3:
These relative imports should and do work. The imports in MatID that you implemented are anyways wrong.

In any case, please, wait for #1 to be merged. Now you are working on top of develop.

I will push the changes to gitlab next week.

ndaelman-hu:
Adding `__init__.py` to `tests` and re-building has solved the issue, while maintaining the relative paths.
Kudos goes to @ladinesa . He also pointed out that for dynamically installed Python envs, absolute paths are fine (better even), but not when dealing regular pip installs.

@JosePizarro3 I rebased to `1-initial-polishing-of-modelsystem`. The effective change is just adding an empty file, so you can also apply this directly to your branch before merging.

JosePizarro3:
> Adding `__init__.py` to `tests` and re-building has solved the issue, while maintaining the relative paths. Kudos goes to @ladinesa . He also pointed out that for dynamically installed Python envs, absolute paths are fine (better even), but not when dealing regular pip installs.
> 
> @JosePizarro3 I rebased to `1-initial-polishing-of-modelsystem`. The effective change is just adding an empty file, so you can also apply this directly to your branch before merging.

Can you target the branch 1-initial-polishing? :)

Issue #3, opened 2024-01-29
Adding ModelMethod for TB Wannier90


JosePizarro3:
As a test, I will transfer the `run.method.tb` section into the new `ModelMethod` base class. The idea is to create `TB(ModelMethod)` which then appears directly under `data`.

PR #7, opened 2024-02-19
Handling of labels in AtomicCell

### LINKED ISSUE #5: Handling of labels in AtomicCell
From the #1 review and after our [meeting on 08.02.2024](https://github.com/FAIRmat-NFDI/AreaC-Meetings/discussions/31), we have to polish the definitions `AtomicCell.labels` and `AtomicCell.atomic_numbers`.

- @ndaelman-hu commented that these quantities should live separately in some section called `AtomType`, possibly merged with the same concepts as in `run.method.atom_parameters`.
- @JFRudzinski commented that we should have 2 quantities to handle generic `labels` and specific `chemical_symbols`, specially to use in the context of models, otherwise it is very buggy when handling normalizations and using ASE.

In my opinion, `AtomicCell` should have a generic `labels`. Then, a couple of classes could be inheriting from it, and then used to be populated in `ModelSystem.atomic_cell`. These two classes (names? `CoarseGrainedCell`, `ElementsCell`?), will then overwrite the meaning of `labels`. Any other extra class (e.g., `LatticeModel`) could be used to overwrite again `labels`. 

I guess that if `labels` exist under some other class `AtomType`, then we can have multiple inheritance in our classes, but I am not 100% sure.

Let's discuss a bit, and I will tackle the issue.

JosePizarro3:
@JFRudzinski @ndaelman-hu please, review when you can this merge request.

I worked out a bit further on the `labels`, `atomic_numbers` and `Method.AtomParameters` ideas and came with the following schema. Just a few points to consider:

1. Now there is further abstraction by having a class `Cell`. This can be used to inherit other system cells, e.g., `CoarseGrainedCell` or `LatticeModelCell`, without the need of pasing real chemical elements information and normalization.
2. The concepts of `labels` and `atomic_numbers` have been moved under a new base class named `AtomsState`. This base class is a combination of these concepts together with the old schema for `AtomParameters`. Everything in this context appears then under `ModelSystem` from now on.
3. This new base class also contains: information about the orbital state, information about the atom charge (as a positive or negative integer, tho I am happy to hear feedback here in case this might be a float), information about CoreHole and HubbardInteractions.
4. The design idea is that `OrbitalsState` has the information of the relevant orbitals for each atom. This means that not all orbitals for an atom (i.e., 1s, 2px, 2py, etc.) need to be necessarily parsed, but rather the ones relevant for the use-case being treated. The example of use-cases could be `CoreHole` and `HubbardInteractions`.
5. For `HubbardInteractions`, now there is a list of `orbitals_ref` which then points to the specific orbitals used for the interactions. This means that `HubbardInteractions.umn` shape is defined by the length of these refs.
6. For `CoreHole`, I also added `orbital_ref`. It is now then a game of adding `n_excited_electrons` and checking the `orbital_ref.degeneracy` to resolve the occupation. Prior to this, `CoreHole` inherited from the state, but I think this is much more clean, as now `CoreHole` only contains info relevant for when creating a hole in a core orbital state.
7. I added `RussellSaundersState` under utils.py. I am still trying to debug and wrap my head around what are these methods using, but it shouldn't be too complicated. I would be ok too to move it to its own module, like you prefer.
8. I will deprecate the old method `MoleculeParameters`, but we can think if there are some use-cases how to implement it. It shouldn't be too complicated.
9. I am considering the old method `Pseudopotentials` to probably live under a `ModelMethod.BasisSet` class. But we (@ndaelman-hu and I) can discuss this a bit later. I am happy to hear more feedback tho.

I hope you like this. With this, I think ModelSystem is fully flexible, normalizations are properly ran, and I will continue working on the #3 `ModelMethod` schema.

_Edit_:
The corresponding [branch in Gitlab](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/commits/1728-full-refactoring-of-run-into-data), [electronic-parsers](https://github.com/nomad-coe/electronic-parsers/tree/test_basesections), and some [La2CuO4_test_basesections_Wannier90.zip](https://github.com/nomad-coe/nomad-schema-plugin-simulation-data/files/14331372/La2CuO4_test_basesections_Wannier90.zip).

### DIFF
--- a/simulationdataschema/atoms_state.py
+++ b/simulationdataschema/atoms_state.py
@@ -71,7 +71,7 @@ orbitals_map = {
 }
 
 
-class OrbitalState(ArchiveSection):
+class OrbitalsState(ArchiveSection):
     """
     A base section used to define the orbital state of an atom.
     """
@@ -259,14 +259,14 @@ class OrbitalState(ArchiveSection):
 
 class CoreHole(ArchiveSection):
     """
-    A base section used to define the core-hole state of an atom by referencing the `OrbitalState`
+    A base section used to define the core-hole state of an atom by referencing the `OrbitalsState`
     section where the core-hole was generated.
     """
 
     orbital_ref = Quantity(
-        type=OrbitalState,
+        type=OrbitalsState,
         description="""
-        Reference to the OrbitalState section that is used as a basis to obtain the `CoreHole` section.
+        Reference to the OrbitalsState section that is used as a basis to obtain the `CoreHole` section.
         """,
         a_eln=ELNAnnotation(component="ReferenceEditQuantity"),
     )
@@ -334,10 +334,10 @@ class HubbardInteractions(ArchiveSection):
     """
 
     orbitals_ref = Quantity(
-        type=OrbitalState,
+        type=OrbitalsState,
         shape=["*"],
         description="""
-        Reference to the OrbitalState sections that are used as a basis to obtain the Hubbard
+        Reference to the `OrbitalsState` sections that are used as a basis to obtain the Hubbard
         interaction matrices.
         """,
     )
@@ -510,19 +510,19 @@ class AtomsState(ArchiveSection):
         """,
     )
 
-    orbitals = SubSection(sub_section=OrbitalState.m_def, repeats=True)
+    orbitals_state = SubSection(sub_section=OrbitalsState.m_def, repeats=True)
 
     charge = Quantity(
         type=np.int32,
         default=0,
         description="""
         Charge of the atom. It is defined as the number of extra electrons or holes in the
-        atom. If the atom is neutral, charge = 0 and the summation of all (if available) the`OrbitalState.occupation`
+        atom. If the atom is neutral, charge = 0 and the summation of all (if available) the`OrbitalsState.occupation`
         coincides with the `atomic_number`. Otherwise, charge can be any positive integer (+1, +2...)
         for cations or any negative integer (-1, -2...) for anions.
 
         Note: for `CoreHole` systems we do not consider the charge of the atom even if
-        we do not store the final `OrbitalState` where the electron was excited to.
+        we do not store the final `OrbitalsState` where the electron was excited to.
         """,
         a_eln=ELNAnnotation(component="NumberEditQuantity"),
     )
--- a/simulationdataschema/atoms_state.py
+++ b/simulationdataschema/atoms_state.py
@@ -344,9 +344,16 @@ class HubbardInteractions(ArchiveSection):
     A base section to define the Hubbard interactions of the system.
     """
 
+    n_orbitals = Quantity(
+        type=np.int32,
+        description="""
+        Number of orbitals used to define the Hubbard interactions.
+        """,
+    )
+
     orbitals_ref = Quantity(
         type=OrbitalsState,
-        shape=["*"],
+        shape=["n_orbitals"],
         description="""
         Reference to the `OrbitalsState` sections that are used as a basis to obtain the Hubbard
         interaction matrices.
@@ -355,7 +362,7 @@ class HubbardInteractions(ArchiveSection):
 
     u_matrix = Quantity(
         type=np.float64,
-        shape=["*", "*"],
+        shape=["n_orbitals", "n_orbitals"],
         unit="joule",
         description="""
         Value of the local Hubbard interaction matrix. The order of the rows and columns coincide
--- a/simulationdataschema/atoms_state.py
+++ b/simulationdataschema/atoms_state.py
@@ -45,6 +45,8 @@ from nomad.metainfo import Quantity, SubSection, MEnum
 from nomad.datamodel.data import ArchiveSection
 from nomad.datamodel.metainfo.annotations import ELNAnnotation
 
+from .utils import RussellSaundersState
+
 
 orbitals = {
     -1: dict(zip(range(4), ("s", "p", "d", "f"))),
@@ -57,6 +59,15 @@ orbitals = {
             ("x(x^2-3y^2)", "xyz", "xz^2", "z^3", "yz^2", "z(x^2-y^2)", "y(3x^2-y^2)"),
         )
     ),
+}
+
+orbitals_map = {
+    "l_symbols": orbitals[-1],
+    "ml_symbols": {i: orbitals[i] for i in range(4)},
+    "ms_symbols": dict((zip((False, True), ("down", "up")))),
+    "l_numbers": {v: k for k, v in orbitals[-1].items()},
+    "ml_numbers": {k: {v: k for k, v in orbitals[k].items()} for k in range(4)},
+    "ms_numbers": dict((zip((False, True), (-0.5, 0.5)))),
 }
 
 
@@ -76,14 +87,30 @@ class OrbitalState(ArchiveSection):
     l_quantum_number = Quantity(
         type=np.int32,
         description="""
-        Azimuthal quantum number l of the orbital state.
+        Azimuthal quantum number l of the orbital state. This quantity is equivalent to `l_quantum_symbol`.
+        """,
+    )
+
+    l_quantum_symbol = Quantity(
+        type=np.int32,
+        description="""
+        Azimuthal quantum symbol l of the orbital state, "s", "p", "d", "f", etc. This
+        quantity is equivalent to `l_quantum_number`.
         """,
     )
 
     ml_quantum_number = Quantity(
         type=np.int32,
         description="""
-        Azimuthal projection of the l vector.
+        Azimuthal projection number of the l vector. This quantity is equivalent to `ml_quantum_symbol`.
+        """,
+    )
+
+    ml_quantum_symbol = Quantity(
+        type=np.int32,
+        description="""
+        Azimuthal projection symbol of the l vector, "x", "y", "z", etc. This quantity is equivalent
+        to `ml_quantum_number`.
         """,
     )
 
@@ -113,6 +140,14 @@ class OrbitalState(ArchiveSection):
         """,
     )
 
+    ms_quantum_symbol = Quantity(
+        type=np.int32,
+        description="""
+        Spin quantum symbol. Set to 'down' for spin down and 'up' for spin up. In non-collinear spin systems,
+        the projection axis $z$ should also be defined.
+        """,
+    )
+
     degeneracy = Quantity(
         type=np.int32,
         description="""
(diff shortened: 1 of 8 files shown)


JosePizarro3:
Btw, @ndaelman-hu, from where did you defined the orbital map? The dictionary that defines `l: ml`. I am curious to check the reference.

simulationdataschema/model_system.py
```diff
@@ -761,37 +740,44 @@ def normalize(self, archive, logger):
...
         - `type` refers to the type of the ModelSystem (atom, bulk, surface, etc.),
-        - `dimensionality` refers to the dimensionality of the ModelSystem (0D, 1D, 2D, 3D),
+        - `dimensionality` refers to the dimensionality of the ModelSystem (0, 1, 2, 3),
 
-    If the ModelSystem `is_representative`, the normalization occurs. The time evolution of
-    the system is encoded on the fact that ModelSystem is a list under Simulation, and for
-    each element of that list, `time_step` can be defined.
+    If the ModelSystem `is_representative`, normalization occurs. The time evolution of the
+    ModelSystem is encoded on the fact that it is a list under `Simulation`, and for each element
+    of that list, `time_step` can be defined.
```
JFRudzinski: Is `time_step` an integer step or a float physical time? Right now in calculation there is both `step` and `time`, which is also able to cover consecutive non-time-dependent workflows like geom. opts. I guess we should continue this approach and use `step` for the referencing between system and calculation (if this will still be relevant), and any other normalization that may be relevant. We will also need to store the physical time in some cases, but this will not always be available. Is this consistent with your view?

JFRudzinski (approved):
This looks great @JosePizarro3 , it appears to me to be more or less what we discussed in your office, and I think it makes a lot of sense to me. 

Just a small comment about the use of `time_step`.

simulationdataschema/atoms_state.py
```diff
@@ -0,0 +1,571 @@
...
+
+    umn = Quantity(
+        type=np.float64,
+        shape=["*", "*"],
+        unit="joule",
+        description="""
+        Value of the local Hubbard interaction matrix. The order of the rows and columns coincide
+        with the elements in `orbital_ref`.
+        """,
+    )
```
ndaelman-hu: I know that this is just a code migration, but I don't remember reviewing these extensions to the Hubbard parameters.

My only remark is about the naming. I really had no idea what `umn` was supposed to represent, until I read the description. Maybe you can make it clearer?
Moreover, are there any limitations on the matrix dimensions? Maybe the no. rows = l-degeneracy (at least in most DFT codes)?

simulationdataschema/atoms_state.py
```diff
@@ -0,0 +1,571 @@
...
+
+    jh = Quantity(
+        type=np.float64,
+        unit="joule",
+        description="""
+        Value of the (interorbital) Hund's coupling.
+        """,
+        a_eln=ELNAnnotation(component="NumberEditQuantity"),
+    )
```
ndaelman-hu: Same remark as above

simulationdataschema/atoms_state.py
```diff
@@ -0,0 +1,571 @@
...
+
+    up = Quantity(
+        type=np.float64,
+        unit="joule",
+        description="""
+        Value of the (interorbital) Coulomb interaction. In rotational invariant systems, up = u - 2 * jh.
+        """,
+        a_eln=ELNAnnotation(component="NumberEditQuantity"),
+    )
```
ndaelman-hu: Same remark as above: `up` is especially confusing. It will give a lot of false hits when performing text searches.
Even `u_h` would already be clearer.

simulationdataschema/atoms_state.py
```diff
@@ -0,0 +1,571 @@
...
+        Returns:
+            (Optional[np.float64]): The effective U parameter.
+        """
+        if self.u is None or self.j is None:
+            logger.warning(
+                "Could not find `HubbardInteractions.u` or `HubbardInteractions.j`."
+            )
+            return None
+        return self.u - self.j
+
```
ndaelman-hu: Nice! Do we have unit tests for these?

simulationdataschema/atoms_state.py
```diff
@@ -0,0 +1,571 @@
...
+                    "The length of `HubbardInteractions.umn` does not coincide with length of `HubbardInteractions.orbitals_ref`."
+                )
+
+
+class AtomsState(ArchiveSection):
+    """
+    A base section to define each atom state information.
+    """
+
+    # ? constraint to the normal chemical elements (no 'X' as defined in ASE included)
```
ndaelman-hu: Was this a remark from me or you?

simulationdataschema/model_system.py
JosePizarro3: Yeah, thanks for bringing it up, I am aware of these quantities; I will add a note to check back this again once we start with transfering the `Outputs`.

Is that OK? (an emoji " 👍🏻  " would do 😄 )

simulationdataschema/atoms_state.py
```diff
@@ -0,0 +1,571 @@
...
+                )
+                return
+        elif self.chemical_symbol is not None and self.atomic_number is None:
+            try:
+                self.atomic_number = ase.data.atomic_numbers[self.chemical_symbol]
+            except IndexError:
+                logger.error(
+                    "The `AtomsState.chemical_symbol` is not recognized in the periodic table."
+                )
+                return
```
ndaelman-hu: Slight suggestion to improve legibility (which is anyhow already pretty good):

```
        f = lambda x: tuple(map(bool, x))
        combo = (self.chemical_symbol, self.atomic_number)

        if f(combo) == f((None, not None)):
            try:
                self.chemical_symbol = ase.data.chemical_symbols[self.atomic_number]
            except IndexError:
                logger.error(
                    "The `AtomsState.atomic_number` is out of range of the periodic table."
                )
        elif f(combo) == f((not None, None)):
            try:
                self.atomic_number = ase.data.atomic_numbers[self.chemical_symbol]
            except IndexError:
                logger.error(
                    "The `AtomsState.chemical_symbol` is not recognized in the periodic table."
                )
        return None
```

You could similarly use other tuples (or a dictionary for better semantics) to condense the mapping and error message too.

You might also prefer to return a boolean for error handling. Isn't necessary atm, but pays off down the road

simulationdataschema/model_method.py
```diff
@@ -47,6 +47,6 @@ class ModelMethod(ArchiveSection):
 
     normalizer_level = 1
 
-    def normalize(self, archive, logger):
+    def normalize(self, archive, logger) -> None:
```
ndaelman-hu, line 50: Great! Don't forget the other types: `EntryArchive` and `BoundLogger`.

simulationdataschema/model_system.py
```diff
@@ -207,16 +209,20 @@ def get_geometric_space_for_atomic_cell(self, logger):
...
-        if ase_atoms is not None:
+        # Skip normalization for `Entity`
+        try:
+            ase_atoms = self.to_ase_atoms(logger)  # function defined in AtomicCell
             self.get_geometric_space_for_atomic_cell(logger)
+        except Exception as e:
+            logger.warning(
+                "Could not extract the geometric space information from ASE Atoms object.",
+            )
+            return
```
ndaelman-hu: `try-except` is indeed more Pythonic.

> except Exception as e:

Try to nail down the exception type as much as possible. Just leaving it at `Exception` is typically considered bad practice, since you might be catching genuine bugs too.
Moreover, if you don't use `e`, then don't extract it.

simulationdataschema/model_system.py
```diff
@@ -325,11 +322,11 @@ class AtomicCell(GeometricSpace):
         shape=["*"],
         # TODO improve description
         description="""
-        Wyckoff letters associated with each atom position.
+        Wyckoff letters associated with each atom.
```
ndaelman-hu, line 338: My comment about the symmetry endures.

simulationdataschema/model_system.py
```diff
@@ -551,19 +520,20 @@ def resolve_analyzed_atomic_cell(
 
         atomic_cell = AtomicCell(type=cell_type)
         atomic_cell.lattice_vectors = cell * ureg.angstrom
+        for label, atomic_number in zip(labels, atomic_numbers):
+            atom_state = AtomsState(chemical_symbol=label, atomic_number=atomic_number)
+            atomic_cell.m_add_sub_section(AtomicCell.atoms_state, atom_state)
```
ndaelman-hu, line 538: Maybe you can turn this into a 1-liner with `map`?

simulationdataschema/model_system.py
```diff
@@ -651,7 +630,7 @@ def resolve_bulk_symmetry(
 
         return primitive_atomic_cell, conventional_atomic_cell
 
-    def normalize(self, archive, logger):
+    def normalize(self, archive, logger) -> None:
```
ndaelman-hu, line 646: Add the remaining types.

simulationdataschema/model_system.py
```diff
@@ -740,7 +719,7 @@ def resolve_chemical_formulas(self, formula: Formula):
         self.hill = formula.format("hill")
         self.anonymous = formula.format("anonymous")
 
-    def normalize(self, archive, logger):
+    def normalize(self, archive, logger) -> None:
```
ndaelman-hu, line 735: Same

simulationdataschema/model_system.py
ndaelman-hu: > If the ModelSystem `is_representative`, normalization occurs.

If `ModelSystem` `is_representative`, proceed with normalization.

> The time evolution of the ModelSystem is encoded on the fact that it is a list under `Simulation`

The time evolution of `ModelSystem` is stored in a `list` format under `Simulation`.

simulationdataschema/model_system.py
```diff
@@ -761,37 +740,44 @@ def normalize(self, archive, logger):
...
+        2. `CoreHoleState.normalize()` in atoms_state.py under `AtomsState`
+        3. `HubbardInteractions.normalize()` in atoms_state.py under `AtomsState`
+        4. `AtomsState.normalize()` in atoms_state.py
+        5. `AtomicCell.normalize()` in atomic_cell.py
+        6. `Symmetry.normalize()` in this class
+        7. `ChemicalFormula.normalize()` in this class
+        8. `ModelSystem.normalize()` in this class
+
+    Note: `normalize()` can be called at any time for each of the classes without being re-triggered
+    by the NOMAD normalization.
```
ndaelman-hu: Very very nice

simulationdataschema/model_system.py
```diff
@@ -924,7 +910,7 @@ def resolve_system_type_and_dimensionality(
         self, ase_atoms: ase.Atoms
     ) -> Tuple[str, int]:
         """
-        Determine the ModelSystem.type and ModelSystem.dimensionality using MatID classification analyzer:
+        Resolves the `ModelSystem.type` and `ModelSystem.dimensionality` using MatID classification analyzer:
```
ndaelman-hu: `MatID`?

simulationdataschema/model_system.py
```diff
@@ -977,7 +963,7 @@ def resolve_system_type_and_dimensionality(
 
         return system_type, dimensionality
 
-    def normalize(self, archive, logger):
+    def normalize(self, archive, logger) -> None:
```
ndaelman-hu, line 978: other types?

simulationdataschema/utils/utils.py
ndaelman-hu, line 1: Yeah, I have to extend `RusselSaunders`. It might warrant its own file, but let's see down the line.

ndaelman-hu (commented):
All minor stuff. Just take a look, no need for additional review.

One question: what if we have basic building blocks larger than atoms, i.e. molecules or coarse-grained? They'll need their own `State`, similar to `AtomState`, I reckon...
Maybe add a generic base class too.

simulationdataschema/model_method.py
JosePizarro3: There are some circular import problems when trying to get `EntryArchive`. I opted for leaving out specifying types inputs in the `normalize()`.

ndaelman-hu: I see... Then never mind.

simulationdataschema/atoms_state.py
JosePizarro3: General information for all of us (included @JFRudzinski). Now `'X'` is excluded from the `chemical_symbol` and `atomic_number`.

I delete the comment.

simulationdataschema/atoms_state.py
JosePizarro3: I'll rename it `u_matrix`. I will also rename other quantities like `u_interaction`, `j_hunds_coupling`, `u_interorbital_interaction`, `j_local_exchange_interaction`. I think `u_effective` can remain as it is.

Yes, the shape of this matrix coincides with the length of `orbitals_ref`.

simulationdataschema/atoms_state.py
JosePizarro3: Not yet, maybe we can tackle testing soon in another branch. For now I am testing directly on a parser and checking that everything goes well.

JosePizarro3: Do you mind opening this as an issue? Maybe you can also share your ideas of what you are expecting for testing and we discuss there.

simulationdataschema/atoms_state.py
ndaelman-hu: > Yes, the shape of this matrix coincides with the length of `orbitals_ref`.

Maybe we should consider adding the shapes again? Let's at least keep it in the `description` for now.

simulationdataschema/atoms_state.py
ndaelman-hu: I can also take it after you merge this and add you as a reviewer?
I'll open the issue this evening. Just remind me if I forget.

simulationdataschema/atoms_state.py
JosePizarro3: Done

simulationdataschema/atoms_state.py
ndaelman-hu: Be careful: this might result in breaking changes with the parsers.

Luca Ghiringhelli had some calculations that were using ghost atoms (`X`).
Maybe you should ask him how we can accommodate him.

simulationdataschema/model_system.py
JosePizarro3: I added the comment saying that we have to clarify whether this belongs to `Symmetry`.

simulationdataschema/model_system.py
JosePizarro3: I prefer to keep it this way, I think it is more readable.

simulationdataschema/atoms_state.py
JosePizarro3: Ok... I can add them back, and add in the description that  `"... 'X' is used in the context of ghots atoms, see e.g., <href>"`.

JosePizarro3: I just wanted to separate a bit the real elements from other concepts, perhaps for these models it is easier to define `labels` as a list, or not defining them at all.

ndaelman-hu: I think that's more in line with its current use, yes.

ndaelman-hu: > I just wanted to separate a bit the real elements from other concepts, perhaps for these models it is easier to define labels as a list, or not defining them at all.

Yeah, they're a pain, since they also prevent the computation of physical quantities like mass density...
Idk, I don't like them either, but they are used by times...

JosePizarro3: I will keep them out for the time being. I have the feeling that I need to see an use-case before we make any decision. I'll add a comment.

simulationdataschema/atoms_state.py
JosePizarro3: Ok, I add them back. The checks are not being performed, but this is something we should discuss further.

PR #8, opened 2024-02-20
Adding ModelMethod for TB Wannier90

### LINKED ISSUE #3: Adding ModelMethod for TB Wannier90

JosePizarro3:
@ndaelman-hu @JFRudzinski 

_A brief intro before our meeting tomorrow 27.02.2024._

This is my initial version on defining `ModelMethod`, some base classes which will be contained within, and some usage in the specific case of `TB` models.

As you can see:
- `ModelMethod` serves as the base class from which to inherit to specific methodologies (e.g., `DFT` should also be inheriting from it, as well as `GW`, `BSE`, `ForceFields`(?), etc.). 
- It also contains info about meshes, self-consistency (this is the former schema `Scf` class), and proxy references to other ModelMethod sections already existing in the archive.

I deleted some quantities and sections which were not so clear or unused:
- `Smearing`
- `Electronic`
- `MoleculeParameters`
- `LatticeModelHamiltonian`
- `Method.stress_tensor_method`, and I defined `methods_ref` as a generic way of coverng all method references under a given `ModelMethod`.
- `Scf.native_tier`

I am still working out the details of `TB`, but there you can see the usage:
- Of referencing specific sections in `ModelSystem`
- Of using this class for further inheritance (there are different types of TB models, `Wannier`, `SlaterKoster`, `xTB`). 
- If you just check the old schema, you would see how complex `SlaterKoster` was to define, as Mohammad (who was the one doing this) was developing this it was very complex due to the problem of composing `TB` under `ModelMethod`. Now, I'll argue that it works much more nicely and it is easily usable by external people.

_____

Now, I will work a bit further on defining the core methodologies (DFT, GW, BSE, DMFT), and you might wanna take over to define your task-specific problems.


JosePizarro3:
@ndaelman-hu @JFRudzinski I updated the comment in the description and added you as reviewers. Let's chat tomorrow.

JosePizarro3:
@ndaelman-hu @JFRudzinski would you mind reviewing this PR? 

You can ignore the changes in `atoms_state.py`. This is done because of the unit testing showed some errors in that schema that I needed to fix.

Everything else in `model_method.py` is new. Like we talked, I left out everything related with `BasisSet` and `ForceField`. I am also skipping the implementation of some functions in `DMFT`, because first I need to see how the schema behaves as a whole.

simulationdataschema/atoms_state.py
```diff
@@ -110,15 +115,15 @@ class OrbitalsState(ArchiveSection):
...
     ms_quantum_number = Quantity(
-        type=np.int32,
+        type=np.float64,
         description="""
         Spin quantum number. Set to -1 for spin down and +1 for spin up. In non-collinear spin
         systems, the projection axis $z$ should also be defined.
         """,
     )
 
     ms_quantum_symbol = Quantity(
```
JFRudzinski, line 125: Just curious what the practicality is of storing the symbol as well as the number?

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+
+
+class NumericalSettings(ArchiveSection):
+    """
+    A base section used to define the numerical settings used in a simulation. These are meshes,
+    self-consistency parameters, and algebraic basis sets.
+    """
+
+    name = Quantity(
+        type=str,
```
JFRudzinski, line 68: We may have discussed this in our meeting, but why not make it an Enum here?

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+        description="""
+        Name of the numerical settings section. This is typically used to easy identification of the
+        `NumericalSettings` section. Possible values: "KMesh", "FrequencyMesh", "TimeMesh",
+        "SelfConsistency", "BasisSet".
+        """,
+        a_eln=ELNAnnotation(component="StringEditQuantity"),
```
JFRudzinski: I'm just trying to make note for my own future usage: Should the eln annotation always be included in these types of base/generic classes? Or what is the intended usage here?

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
+
+        # Set the name of the section
+        self.name = self.m_def.name if self.name is None else self.name
+
+
+class SelfConsistency(NumericalSettings):
+    """
+    A base section used to define the parameters used in the self-consistent field (SCF) simulations.
```
JFRudzinski: not a big deal, but is "simulation" the standard terminology? In my mind a simulation is some sort of modeling of reality, whereas a numerical optimization to converge upon a model or set of parameters I would call a calculation...but maybe this is not universal?

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+        super().normalize(archive, logger)
+
+        # Set the name of the section
+        self.name = self.m_def.name if self.name is None else self.name
 
 
 class ModelMethod(ArchiveSection):
-    """ """
+    """
+    Model method input parameters and numerical settings used to simulate the materials properties.
```
JFRudzinski: "used to simulate materials properties"

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+    # ! add more description once section is finished
+    """
 
     normalizer_level = 1
 
+    name = Quantity(
+        type=str,
+        description="""
+        Name of the model method. This is typically used to easy identification of the `ModelMethod` section.
+        Possible values: "DFT", "TB", "BSE", "GW", "DMFT".
```
JFRudzinski: Also for numerical settings type above: If it is not an Enum are these really "possible" values, or more values existing within this schema...again I am a little confused about when to use Enum

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+    )
+
+    n_method_references = Quantity(
+        type=np.int32,
+        description="""
+        Number of other `ModelMethod` section references of the current method.
+        """,
+    )
+
+    methods_ref = Quantity(
```
JFRudzinski: Something about this is a little unclear to me. I guess I don't understand the global strategy for referencing between workflows, methods, systems, and outputs:
1. I understand why you need to connect 2 or more methods, in cases where distinct methods are being applied simultaneously (in 1 step of a workflow e.g.), but I guess I don't understand how we will manage these references in relation to the other sections.
2. Will system also contain references to the corresponding method? If so, will it only ref 1 and then you connect the others through this ref somehow?
... etc, I have some other thoughts but I think it is more productive to just talk it out

JFRudzinski (approved):
This is great! Just to bring my main questions to the forefront:
1. When should we be using Enum values for method names?
2. When should we be adding eln annotations to quantities?
3. What is our overall strategy for referencing between methods and the other main sections (workflows, systems, outputs)?

ndaelman-hu:
@JosePizarro3 I'll do the review this evening, okay?

.github/workflows/actions.yaml
ndaelman-hu, line 1: I would add 2 ruff tests:
```
	ruff-linting:
		runs-on: ubuntu-latest
		steps:
			- uses: actions/checkout@v3
			- uses: chartboost/ruff-action@v1
			with:
				args: "check ."

	ruff-formatting:
		runs-on: ubuntu-latest
		steps:
			- uses: actions/checkout@v3
			- uses: chartboost/ruff-action@v1
			with:
				args: "format ."
```

simulationdataschema/atoms_state.py
ndaelman-hu: It's more so for search, but I agree that it's stupid.
There are several discussion points where I've brought it up that this kind of translation should be done at a higher level than the archive. Atm, there's no system in place, but I want to look into it with Lauri coming April.

simulationdataschema/atoms_state.py
```diff
@@ -163,63 +169,72 @@ def __init__(self):
...
-    ) -> None:
+        self, quantum_name: str, quantum_type: str, logger: BoundLogger
+    ) -> Optional[Union[str, int]]:
         """
-        Resolves the quantum number from the `self._orbitals_map` on the passed type. `type` can be either
-        'number' or 'symbol'. If the quantum type is not found, then the countertype
-        (e.g., type = 'number' => countertype = 'symbol') is used to resolve it.
+        Resolves the quantum number or symbol from the `self._orbitals_map` on the passed `quantum_type`.
+        `quantum_type` can be either 'number' or 'symbol'. If the quantum type is not found, then the countertype
+        (e.g., quantum_type = 'number' => countertype = 'symbol') is used to resolve it.
```
ndaelman-hu: Keep the style: (e.g., `quantum_type == 'number'` => `countertype == 'symbol'`).
Since the string values are also still code, I'd surround then with ` too.

simulationdataschema/atoms_state.py
```diff
@@ -163,63 +169,72 @@ def __init__(self):
...
+            (Optional[Union[str, int]]): The quantum number or symbol resolved from the orbitals_map.
         """
-        if quantum_number not in ["l", "ml", "ms"]:
+        if quantum_name not in ["l", "ml", "ms"]:
+            logger.warning("The quantum_name is not recognized. Try 'l', 'ml' or 'ms'.")
+            return None
+        if quantum_type not in ["number", "symbol"]:
             logger.warning(
-                "The quantum number is not recognized. Try 'l', 'ml' or 'ms'."
+                "The quantum_type is not recognized. Try 'number' or 'symbol'."
```
ndaelman-hu: I'd make this warning slightly more verbose, e.g.
```
f"The `quantum_type` {quantum_type} is not recognized. Expected `'number'` or `'symbol'`."
```

simulationdataschema/atoms_state.py
```diff
@@ -163,63 +169,72 @@ def __init__(self):
...
-            return
+            return None
 
         # If the counterpart exists, then resolve the quantity from the orbitals_map
-        quantity = self._orbitals_map.get(f"{quantum_number}_{type}s", {}).get(
-            other_quantity
-        )
-        setattr(self, f"{quantum_number}_quantum_{type}", quantity)
+        orbital_quantity = self._orbitals_map.get(f"{quantum_name}_{quantum_type}s", {})
+        if quantum_name == "l" or quantum_name == "ms":
```
ndaelman-hu:
```
if quantum_name in ("l", "ms"):
```

simulationdataschema/atoms_state.py
```diff
@@ -514,8 +544,11 @@ def normalize(self, archive, logger) -> None:
             ) = self.resolve_u_interactions(logger)
 
         # If u_effective is not available, calculate it
-        if self.u_effective is None:
-            self.u_effective = self.resolve_u_effective(logger)
+        self.u_effective = (
+            self.resolve_u_effective(logger)
+            if self.u_effective is None
+            else self.u_effective
+        )
```
ndaelman-hu: I liked the older version better. I remember reviewing a refactoring for Alvin some months ago, and he had removed all inline definitions.

simulationdataschema/model_system.py
```diff
@@ -753,11 +737,10 @@ def normalize(self, archive, logger) -> None:
 
 
 class ModelSystem(System):
-    # ! Work in this description with new `AtomState`
     """
-    Model system used as an input for the computation.
+    Model system used as an input for simulate the material.
```
ndaelman-hu: typo:
```
Model system used as an input for simulating the material.
```

simulationdataschema/model_system.py
```diff
@@ -66,7 +66,7 @@
 
 class GeometricSpace(Entity):
     """
-    A base section to define geometrical space-related entities.
+    A base section used to define geometrical space-related entities.
```
ndaelman-hu:
```
Base section defining geometrical spaces and their entities.
```

simulationdataschema/common.py
```diff
@@ -0,0 +1,89 @@
...
+import numpy as np
+
+from nomad.datamodel.data import ArchiveSection
+from nomad.metainfo import Quantity
+
+
+class HoppingMatrix(ArchiveSection):
+    """
+    Section containing the hopping/overlap matrix elements between N projected orbitals. This
+    is also the output of a TB calculation.
```
ndaelman-hu, line 49: The TB bit might be a bit too specific. Isn't this also part of the Hubbard model, and by extension DMFT?

simulationdataschema/common.py
```diff
@@ -0,0 +1,89 @@
...
+        """,
+    )
+
+    degeneracy_factors = Quantity(
+        type=np.int32,
+        shape=["n_wigner_seitz_points"],
+        description="""
+        Degeneracy of each Wigner-Seitz grid point.
+        """,
+    )
```
ndaelman-hu: Can this role be fulfilled by the atom state? How do you normally obtain the degeneracy here?

simulationdataschema/common.py
```diff
@@ -0,0 +1,89 @@
...
+        shape=["n_wigner_seitz_points", "n_orbitals * n_orbitals", 7],
+        description="""
+        Real space hopping matrix for each Wigner-Seitz grid point. The elements are
+        defined as follows:
+
+            n_x   n_y   n_z   orb_1   orb_2   real_part + j * imag_part
+
+        where (n_x, n_y, n_z) define the Wigner-Seitz cell vector in fractional coordinates,
+        (orb_1, orb_2) indicates the hopping amplitude between orb_1 and orb_2, and the
+        real and imaginary parts of the hopping in electron_volt.
```
ndaelman-hu: Personally, I prefer eV, but I think we still use J... I'd really like to change this.
Moreover, you should specify the units.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+)
+
+from .model_system import ModelSystem
+from .atoms_state import OrbitalsState, CoreHole
+
+
+class NumericalSettings(ArchiveSection):
+    """
+    A base section used to define the numerical settings used in a simulation. These are meshes,
+    self-consistency parameters, and algebraic basis sets.
```
ndaelman-hu: Why call them "algebraic" basis sets? Idk if algebra has much to do with those. Moreover, several basis sets can be construed as meshes, e.g. plane waves.

simulationdataschema/model_method.py
ndaelman-hu: If I remember the argument pro `str` well, there are simply too many possibilities here.
We want the community to use these sections without constraints, else we're back at them inheriting just to extend the Enum...
`results` seems like a more appropriate level at which to constrain.

simulationdataschema/model_method.py
ndaelman-hu: ELN can cover cases where we cannot determine a quantity by parsing. Here, the user might further refine the quantity.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+        type=MEnum(
+            "Gamma-centered",
+            "Monkhorst-Pack",
+            "Gamma-offcenter",
+            "Line-path",
+            "Equidistant",
+            "Logarithmic",
+            "Tan",
+            "Gauss-Legendre",
+            "Gauss-Laguerre" "Clenshaw-Curtis",
```
ndaelman-hu: missing comma?

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+        | `'Gamma-offcenter'` | Regular mesh with an offset that is neither `'Gamma-centered'`, nor `'Monkhorst-Pack'`. |
+        | `'Line-path'` | Line path along high-symmetry points. Typically employed for simualting band structures. |
+        | `'Equidistant'`  | Equidistant 1D grid (also known as 'Newton-Cotes') |
+        | `'Logarithmic'`  | log distance 1D grid |
+        | `'Tan'`  | Non-uniform tan mesh for 1D grids. More dense at low abs values of the points, while less dense for higher values |
+        | `'Gauss-Legendre'` | Quadrature rule for integration using Legendre polynomials |
+        | `'Gauss-Laguerre'` | Quadrature rule for integration using Laguerre polynomials |
+        | `'Clenshaw-Curtis'`  | Quadrature rule for integration using Chebyshev polynomials using discrete cosine transformations |
+        | `'Gauss-Hermite'`  | Quadrature rule for integration using Hermite polynomials |
+        """,
```
ndaelman-hu: I think this quantity is mixing 3 concepts: center, spacing, dimensionality. I would suggest splitting these into separate quantities.
I do think that the nomenclature is relevant for
1. search. Here, a dynamic layer would be helpful.
2. parsing:  some codes indeed use these terminologies directly. I would supply a conversion function here.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+    )
+
+    points = Quantity(
+        type=np.complex128,
+        shape=["n_points", "dimensionality"],
+        description="""
+        List of all the points in the mesh.
+        """,
+    )
```
ndaelman-hu: Doesn't this constrain the grid to having the same no. points along each axis?

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+
+    # ! is this description correct?
+    weights = Quantity(
+        type=np.float64,
+        shape=["n_points"],
+        description="""
+        The frequency of times the same point reappears. A value larger than 1, typically
+        indicates a symmtery operation that was applied to the mesh.
+        """,
+    )
```
ndaelman-hu: If I remember well, weights are the normalized version of `multiplicities`. This should be apparent from some normalizers / parsers.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+        Name of the high-symmetry starting point of the line path segment.
+        """,
+    )
+
+    end_point = Quantity(
+        type=str,
+        description="""
+        Name of the high-symmetry end point of the line path segment.
+        """,
+    )
```
ndaelman-hu: 1. Is it necessary to denote a traversal direction? I'd argue that you can flip it, so maybe both points could be contained in a quantity of shape `2`.
2.  While generically I agree that one should submit high-symmetry points, might this be too restrictive?

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+
+    points = Quantity(
+        type=np.float64,
+        shape=["n_line_points", 3],
+        description="""
+        List of all the points in the line path segment.
+        """,
+    )
```
ndaelman-hu: Maybe we can normalize this as equidistant when only `n_line_points` are submitted?

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+
+    all_points = Quantity(
+        type=np.float64,
+        shape=["*", 3],
+        description="""
+        Full list of the mesh points without any symmetry operations.
+        """,
+    )
```
ndaelman-hu: Shouldn't we also have a primitive counterpart?

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+            points = np.meshgrid(*[np.linspace(0, 1, n) for n in self.grid])
+        elif self.sampling_method == "Monkhorst-Pack":
+            try:
+                points += monkhorst_pack(self.grid)
+            except ValueError:
+                logger.warning(
+                    "Could not resolve `KMesh.points` from `KMesh.grid`. Monkhorst-Pack failed."
+                )
+                return None  # this is a quick workaround: k_mesh.grid should be symmetry reduced
+        return points
```
ndaelman-hu: I think you can also determine `offset` from Gamma-centered or Monkhorst.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+        for model_system in model_systems:
+            # General checks to proceed with normalization
+            if not model_system.is_representative:
+                logger.warning(
+                    "The last `ModelSystem` was not found to be representative."
+                )
+                return None
+            if model_system.type != "bulk":
+                logger.warning("`ModelSystem.type` is not describing a bulk system.")
+                return None
```
ndaelman-hu: You can add a TODO for me: this should be extended to all dimensions.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+        Returns:
+            (Optional[pint.Quantity]): The resolved `k_line_density` of the `KMesh`.
+        """
+        for model_system in model_systems:
+            # General checks to proceed with normalization
+            if not model_system.is_representative:
+                logger.warning(
+                    "The last `ModelSystem` was not found to be representative."
+                )
+                return None
```
ndaelman-hu: I have a feeling that these checks are so ubiquitous, we should have a special function for them in 1 of our root base sections.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+            self.resolve_points(logger) if self.points is None else self.points
+        )
+
+        # Calculate k_line_density for data quality measures
+        model_systems = self.m_xpath("m_parent.m_parent.model_system", dict=False)
+        self.k_line_density = (
+            self.resolve_k_line_density(model_systems, logger)
+            if self.k_line_density is None
+            else self.k_line_density
+        )
```
ndaelman-hu: Same comment as above about these kind of in-line options

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+
+
+class FrequencyMesh(Mesh):
+    """
+    A base section used to specify the settings of a sampling mesh in the frequency real or imaginary space.
+    """
```
ndaelman-hu: I think this really specific to QP, and maybe should be reflected in the name? There are other physical processes (e.g. glass transitions) where only real-valued frequency is relevant. @JFRudzinski can probably list more here.

simulationdataschema/model_method.py
ndaelman-hu: Agreed: I've mostly heard it as SCF calculation.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+
+        # Set the name of the section
+        self.name = self.m_def.name if self.name is None else self.name
+
+
+class SelfConsistency(NumericalSettings):
+    """
+    A base section used to define the parameters used in the self-consistent field (SCF) simulations.
+    This section is useful to determine if a simulation `is_converged` or not (see outputs.py) and if
+    the simulation fulfills:
```
ndaelman-hu: This description overrepresents variations on "use". Slight reformulation:
```
A base section defining the convergence settings of self-consistent field (SCF) calculations.
It determines the conditions for `is_converged` in `Output`. Convergence criteria covered are:
```

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+
+        1. The number of iterations is not larger or equal than `n_max_iterations`.
```
ndaelman-hu: No need to negate when there's an obvious positive:
```
1. The number of iterations is smaller than or equal to `n_max_iterations`.
OR
2. The number of iterations does not exceed `n_max_iterations`.
```

You could also use a more symbolic notation if you so wish.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
 
 class ModelMethod(ArchiveSection):
-    """ """
+    """
+    Model method input parameters and numerical settings used to simulate the materials properties.
+
+    # ! add more description once section is finished
+    """
 
     normalizer_level = 1
```
ndaelman-hu, line 539: This is an interesting solution. Given that this variable is for normalizers only, I'd hide it, i.e. `_normalizer_level`.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+        | --------------------- | ----------------------------------------- |
+        | `"TS"`  | http://dx.doi.org/10.1103/PhysRevLett.102.073005 |
+        | `"OBS"` | http://dx.doi.org/10.1103/PhysRevB.73.205101 |
+        | `"G06"` | http://dx.doi.org/10.1002/jcc.20495 |
+        | `"JCHS"` | http://dx.doi.org/10.1002/jcc.20570 |
+        | `"MDB"` | http://dx.doi.org/10.1103/PhysRevLett.108.236402 and http://dx.doi.org/10.1063/1.4865104 |
+        | `"XC"` | The method to calculate the Van der Waals energy uses a non-local functional |
+        """,
+        a_eln=ELNAnnotation(component="EnumEditQuantity"),
+    )
```
ndaelman-hu: This quantity goes under DFT: London dispersion is often unaccounted for in most DFs. They can then be corrected using external models. All of these are examples of that, with the exception of "XC". That one refers to DFs which are non-local.

I would also change the name to `vdw_correction`. I no way does this (or should we aim to) capture the full range of VdW interactions. Other methods like RPA, MP2, etc. all account for these internally. If we want to convey that, then we need an ontology.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+            "scalar_relativistic",
+            "pseudo_scalar_relativistic",
+            "scalar_relativistic_atomic_ZORA",
+        ),
+        description="""
+        Describes the relativistic treatment used for the calculation of the final energy
+        and related quantities. If `None`, no relativistic treatment is applied.
+        """,
+        a_eln=ELNAnnotation(component="EnumEditQuantity"),
+    )
```
ndaelman-hu: Hmmm... let me double-check, but I think these are general. Some (ZORA) may be specific to just pseudopotentials, though.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+        a_eln=ELNAnnotation(component="StringEditQuantity"),
+    )
+
+    name = Quantity(
+        type=MEnum("exchange", "correlation", "hybrid", "contribution"),
+        description="""
+        Name of the XC functional. It can be one of the following: 'exchange', 'correlation',
+        'hybrid', or 'contribution'.
+        """,
+    )
```
ndaelman-hu: You can ignore my feedback on Discord regarding the subsection. I'm not quite happy with this structure, but I'll compromise, till I find something better.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+        For example, if a calculations using a hybrid XC functional (e.g., HSE06)
+        specifies a user-given value of the mixing parameter between exact and GGA
+        exchange, then this non-default value is stored in this metadata.
+
+        The labels and units of these values may be defined in name.
+
+        If this metadata is not given, the default parameter values for the functional
+        are assumed.
+        """,
+    )
```
ndaelman-hu: Mmm... I don't like this quantity. Some parameters are vastly more used than others, e.g. exact exchange mixing, range-separation parameter.
Some codes do allow people to compose DFs out of others. My solution would be to make this a repeating subsection and give each a weight.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+
+    weight = Quantity(
```
ndaelman-hu, line 632: Haha, exaclty!

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+        Args:
+            weight (Optional[np.float64]): The weight of the functional.
+
+        Returns:
+            (Optional[str]): The weight as a string with a "*" added at the end.
+        """
+        weight_name = ""
+        if weight is not None:
+            weight_name = f"{str(weight)}*"
+        return weight_name
```
ndaelman-hu: See my feedback on Discord.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+    def __init__(self, m_def: Section = None, m_context: Context = None, **kwargs):
+        super().__init__(m_def, m_context, **kwargs)
+        self._jacobs_ladder_map = {
+            "lda": "LDA",
+            "gga": "GGA",
+            "mgg": "meta-GGA",
+            "hyb_mgg": "hyper-GGA",
+            "hyb": "hybrid",
+        }
+        self._jacobs_ladder_map_extended = {**self._jacobs_ladder_map, "hf_": "HF"}
```
ndaelman-hu: HF should get its own section...

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+
+    def resolve_libxc_names(
+        self, xc_functionals: List[XCFunctional]
+    ) -> Optional[List[str]]:
+        """
+        Resolves the `libxc_names` and sortes them from the list of `XCFunctional` sections.
```
ndaelman-hu: typo: sorts

Also, I guess you mean that it sorts the list.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+
+        Returns:
+            (Optional[List[str]]): The resolved and sorted `libxc_names`.
+        """
+        libxc_names = [
+            functional.libxc_name
+            for functional in xc_functionals
+            if functional.libxc_name is not None
+        ]
+        return sorted(libxc_names)
```
ndaelman-hu:
```
return sorted(
 [
            functional.libxc_name
            for functional in xc_functionals
            if functional.libxc_name is not None
        ]
)
```

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+                return 1 / 3
+            elif re.search("_PBE38$", xc_name):
+                return 3 / 8
+            elif re.search("_PBE50$", xc_name):
+                return 0.5
+            elif re.search("_M06_2X$", xc_name):
+                return 0.54
+            elif _scan_patterns(["M05_2X", "PBE_2X"], xc_name):
+                return 0.56
+        return None
```
ndaelman-hu: Ow God, this list has to have about 200 entries...

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+
+
+class TB(ModelMethodElectronic):
+    """
+    A base section containing the parameters pertaining to a tight-binding (TB) model calculation.
+    The type of tight-binding model is specified in the `type` quantity.
+    """
+
+    type = Quantity(
+        type=MEnum("DFTB", "xTB", "Wannier", "SlaterKoster", "unavailable"),
```
ndaelman-hu: I was also using SlaterKoster in reference to the atom state. Let me know if we should avoid the name clash.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+        if atomic_cell is None:
+            logger.warning("`AtomicCell` section was not found.")
+            return None
+
+        # If there is no child `ModelSystem`, the normalization will not run
+        atoms_state = atomic_cell.atoms_state
+        model_system_child = model_system.model_system
+        if model_system_child is None:
+            logger.warning("No child `ModelSystem` section was found.")
+            return None
```
ndaelman-hu: These are the same ubiquitous checks I mentioned before. These could go into separate functions into a higher base class.

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+            )
+            return
+        # This normalization only considers the last `ModelSystem` (default `model_index` argument set to -1)
+        orbitals_ref = self.resolve_orbital_references(model_systems, logger)
+        if orbitals_ref is not None and self.orbitals_ref is None:
+            self.n_orbitals = len(orbitals_ref)
+            self.orbitals_ref = orbitals_ref
+
+
+class Wannier(TB):
```
ndaelman-hu, line 1037: I also know of Wannier being used outside of TB, e.g. band structures: https://www.vasp.at/wiki/index.php/Si_bandstructure

simulationdataschema/model_method.py
```diff
@@ -34,19 +34,1646 @@
...
+        """,
+    )
+
+    magnetic_state = Quantity(
+        type=MEnum("paramagnetic", "ferromagnetic", "antiferromagnetic"),
+        description="""
+        Magnetic state in which the DMFT calculation is done. This quantity can be obtained from
+        `orbitals_ref` and their spin state.
+        """,
+    )
```
ndaelman-hu: Is this unique to DMFT? I think this should go with the non-collinear specifications.

ndaelman-hu (commented):
Hey! Very nice work. Mostly small stuff.

Tbh, I didn't check the methods after TB much, but I can see to help there with the testing of the normalizers.
I will brainstorm a bit on how to improve the DFT section too.

simulationdataschema/atoms_state.py
JosePizarro3: Well, there is also the question on whether some codes might store some information (e.g. the number), and others the equivalent one (e.g. the symbol).

In those cases, we would be enforcing people with one or another. But yes, I agree that might be the best.

simulationdataschema/atoms_state.py
JosePizarro3: I restored it

simulationdataschema/common.py
JosePizarro3: I defined them in "joule".

simulationdataschema/common.py
JosePizarro3: No, this is the degeneracy of Wigner-Seitz cells, meaning, it refers to neighboring atoms and their orbitals. The information in `AtomicCell.AtomsState` only refers to the (0, 0, 0) one.

What I have to do with Mohammad, is to merge this idea with the `SlaterKoster` one.

simulationdataschema/common.py
JosePizarro3: Well, this common.py is not very polished. I will revisit it later when defining outputs.py

simulationdataschema/model_method.py
JosePizarro3: Well, at the end you are constructing a basis set B = {|φG1>, |φG2>, ..., |φGN>} (e.g. for plane waves characterized by plane wave vectors Gi).

In any case, I deleted algebraic (which comes from "linear algebra").

simulationdataschema/model_method.py
JosePizarro3: Yes, I am with @ndaelman-hu here

simulationdataschema/model_method.py
JosePizarro3: indeed, thanks!

simulationdataschema/model_method.py
JosePizarro3: No. This is just the list of points. `n_points` is just the summation of all points independently of spacing.

simulationdataschema/model_method.py
JosePizarro3: Yes, it is a good idea.

simulationdataschema/model_method.py
JosePizarro3: Not sure if I understand point 1.

I guess we can simply save a shape=2 quantity called `high_symmetry_points` so that people can store them.

simulationdataschema/model_method.py
JosePizarro3: Is not `points` handling that if `all_points` is present?

simulationdataschema/model_method.py
JosePizarro3: I think here `FrequencyMesh` represents any discrete mesh for [ω0, ω1, ω2, ..., ωN]. It only represents a discrete mesh whose density might be relevant to know the precision of a calculation.

simulationdataschema/model_method.py
JosePizarro3: Ok, just changed it to "calculation". However, "simulation", "computation", "calculation"... are all synonyms in this schema

simulationdataschema/model_method.py
JosePizarro3: I think @ndaelman-hu replied well above.

Just to give a real case example, there was this user who wanted "kMC" covered as a label. With this, he can do it, while if we enforce MEnum, they still have to pass through us.

I will anyways add `suggestions`. I think this is possible in the ELNs, let me check.

simulationdataschema/model_method.py
JosePizarro3: 1. I think in my mind, `method[-1]` will be always the one used for reference (e.g., in an output), while the others will be part of `method[-1].methods_ref`.
2. Cannot imagine an example where this can occur. But maybe we should simply work on referencing in outputs.py, and then go back to these kind of quantities? 

As I see it, referencing in these sections is purely to make life easier when trying to find the provenance of a simulation. E.g., in the case of adding `TB.orbitals_ref`, this kind of info can be anyways accessed via other methods, but I just decided to add it there for convenience. This does not mean that it is the correct way of doing it.

simulationdataschema/model_method.py
JosePizarro3: Ok, I will move it down and change the name.

simulationdataschema/model_method.py
JosePizarro3: I just copied whatever was in the MethodNormalizer :-)

simulationdataschema/model_method.py
JosePizarro3: Ok, it does not work. So I can improve a bit the description and talk to Area D to implement this, perhaps.

simulationdataschema/model_method.py
JosePizarro3: Sorry, Github-lens messed with the position of the comments...

JosePizarro3: So this schema is pretty much based on what was in the old one. Feel free to modify it at will (even merging XCFunctional and DFT like you suggested).

simulationdataschema/model_method.py
JosePizarro3: My question here: is this anything else than `exact_exchange_mixing_factor`? If not, then I will use the quantity defined in DFT and forget about this.

JosePizarro3: Ok, I saw your comment on the weight. I then delete this, and we keep `weight`.

simulationdataschema/model_method.py
JosePizarro3: I will keep `weight`, delete `parameters`.

simulationdataschema/model_method.py
JosePizarro3: this mapping is not being used anyways, so I delete it.

simulationdataschema/model_method.py
JosePizarro3: Where?

simulationdataschema/model_method.py
JosePizarro3: That is also the same as here. VASP2WANNIER90 actually calculates first the Wannier tight binding model, and then diagonalizes it along the same path as the DFT bands.

What I mean is that there is no sense in "Wannier being used outside of TB, e.g. band structures". The bands are calculated from a Wannier TB model.

simulationdataschema/model_method.py
JosePizarro3: I added a comment, but yes. This should be obtained by other means (in fact, you see below something defined like `resolve_magnetic_state`, but I believe this should go into `ModelSystem` by checking what are the `AtomsState` in the atomic cell.

JosePizarro3: I removed the function and also added something there to remember we have to pass this into ModelSystem

simulationdataschema/atoms_state.py
ndaelman-hu: You're right, but that could be mitigated by having a conversion function: storing is different from parsing or serving.
Regardless, I think it's fine for now.

JosePizarro3:
Thanks a lot for your reviews!! 👏🏻 

@ndaelman-hu @JFRudzinski would you mind to quickly checking again? Focus on your comments, close them if you feel satisfied, and if everything is good, I will merge.

ndaelman-hu:
> Thanks a lot for your reviews!! 👏🏻
> 
> @ndaelman-hu @JFRudzinski would you mind to quickly checking again? Focus on your comments, close them if you feel satisfied, and if everything is good, I will merge.

I was just about to head home. Do you mind if I do it tomorrow morning?

simulationdataschema/atoms_state.py
ndaelman-hu: I'm fine with this. @JFRudzinski ?

simulationdataschema/model_method.py
ndaelman-hu: Yeah, I'll open an issue for me to look into this.
libxc has a Python module, but I don't think it serves this kind of information.

simulationdataschema/common.py
ndaelman-hu: So you're saying that this degeneracy doesn't stem from the onsite electronic structure, nor the lattice symmetry, but rather interaction between neighboring cells?
 
Anyhow, I'll leave up to the both of you to straighten this out. Would be quite interested in knowing what you eventually settle on.

simulationdataschema/model_method.py
ndaelman-hu: In the `CoreHole`s. Slater-Koster just refers to the J, M notation of the atom state, so it's normal that it pops up at various places.

While I think that we should support this and other common notations (the hydrogen quantum numbers, crystal field, spectroscopic notation, etc.), I think we should aim for a unified storage based on their mathematical commonality, i.e. point groups and character theory.

simulationdataschema/model_method.py
ndaelman-hu: Interesting! Thank you for clarifying that.

simulationdataschema/model_method.py
ndaelman-hu: Personally, I don't have a use case for this (yet). At best maybe basis set extrapolations, but I was thinking of using a workflow there. I kind of get @JFRudzinski 's questions, but let me reformulate them in my own words:

@JosePizarro3 Workflows are already tools for provenance. Now, I'm fine adding more, but the distinction on when to use each should be clear (and documented).
Also, how should information be retrieved here? Is the idea that we only show data from the last method? When should a user traverse the method graph? What kind of tools should we offer there?

simulationdataschema/model_method.py
ndaelman-hu: @JosePizarro3 Can elaborate on what didn't work?
If it's just the ELN suggestions, I think we can look into that separately with Area D. This can be merged.

I didn't find this piece of code (marked as outdated) anymore. Did you end up removing it?

simulationdataschema/model_method.py
ndaelman-hu: Okay, that's better. While most QM theory indeed focuses on basis sets to a vector space, there are also grid-based approaches (so no linear operations) and wavelets (I'm not an expert in this theory, but I understand them to be a generalization of the Fourier basis set, but using a group operation?).

simulationdataschema/model_method.py
ndaelman-hu: @JosePizarro3 ?

simulationdataschema/model_method.py
ndaelman-hu: Indeed. The naming now suggests that there's a specific direction.
Physically, this is irrelevant. It only matters for visualization. I would go with your suggestion of combining both into a single quantity, saying that the order only determines plotting.

simulationdataschema/model_method.py
ndaelman-hu: Yes, but I was more so referring to the physical interpretation. There still is one, since you are using energy units for `points`. In the context of time, frequency is Hz, not energy. There is of course a deep connection between both, as time-symmetry leads to energy conservation. QP (or at least Green's functions) implicitly uses this connection. For statistical systems (glass transitions, chemical reactions), you could recover a different time-energy relation via ergodicity, but there it's safer to just speak of s or Hz without the energy connotation.
In the case of FFT, then this mesh units could be anything really, e.g. inverted electron density in VASP.

My comment here is about whether we should reserve a generic name like `Frequency` for solely QP? I think you can retain this section, but that it should be renamed to better clarify its semantics.

ndaelman-hu (approved):
Okay, all done! I closed all that is resolved from my side. Just check the open comments.

I'm really eager to get hands-on with this schema, as now we will have much clearer business logic.

simulationdataschema/model_method.py
JosePizarro3: I tried adding:
```python
    ...
    a_eln=ELNAnnotation("StringEditQuantity", props=dict(suggestions=["DFT", "TB", "GW", "BSE", "DMFT", "kMC", "NMR"]))
    ...
```

But the develop is a bit buggy, so I couldn't see whether this was a real problem or something else going on.

simulationdataschema/model_method.py
JosePizarro3: Yeah, I agree this might not be very clear...

I was trying to come up with an example, but I cannot imagine one right now which will not require a workflow instead.

However, we might find cases where it is easier to simply define 2 ModelMethod sections, and hence use the last one to ref the first one; probably because the first one does not contain enough information to be an Entry itself. However, this is what I found problems supporting, as indeed, we could anyways break this into 2 Entries.

This is then making me wonder: should we enforce that `ModelMethod` **does not** repeat, and then whenever anyone needs to create another ModelMethod instance, it should be in another Entry?

simulationdataschema/model_method.py
JosePizarro3: Then I can improve on the description. There is no normalization anywhere (I didn't check the parsers tho)

simulationdataschema/model_method.py
JosePizarro3: I see, interesting. Ok, thanks for the info, I will rename it to `QuasiparticlesFrequencyMesh`.

simulationdataschema/model_method.py
JosePizarro3: I couldn't find anything about that.

The "Slater-Koster" method is a method to obtain the two-center (meaning, two atoms) overlapping integrals between their orbitals. The [old paper](Slater J C and Koster G F 1954 Phys. Rev. 94 1498) is surprisingly easy to read, and it mainly contains tables and tables of integrals allowed by symmetry in different crystal environments.

For core-holes, my guess is that they might be doing some two-center integration as well. But in that case, I would call it "TwoCenterIntegrals", not "SlaterKoster".

If you share some reference I can take a look.

simulationdataschema/model_method.py
JosePizarro3: Indeed, I added it (and change the name of the method to `resolve_points_and_offset`).

simulationdataschema/model_method.py
JosePizarro3: I just decided to take it out for now.

We can decide later along the way when transferring parsers into this new schema.

simulationdataschema/model_method.py
JosePizarro3: Changed to:
```python
    multiplicities = Quantity(
        type=np.float64,
        shape=['n_points'],
        description="""
        The amount of times the same point reappears. A value larger than 1, typically indicates
        a symmtery operation that was applied to the `Mesh`. This quantity is equivalent to `weights`:

            multiplicities = 1 / weights
        """,
    )

    weights = Quantity(
        type=np.float64,
        shape=['n_points'],
        description="""
        Weight of each point. A value smaller than 1, typically indicates a symmtery operation that was
        applied to the mesh. This quantity is equivalent to `multiplicities`:

            weights = 1 / multiplicities
        """,
    )
```

I am resolving the conversation too.

simulationdataschema/model_method.py
JosePizarro3: Ok, I added some normalization.

I also changed `KMesh.high_symmetry_points` to be of JSON type. There, we could store all the high symmetry points of a given lattice, and use it here. I left it for later, as there is some normalization in the `BandStructureNormalizer`.

simulationdataschema/model_method.py
JosePizarro3: It is indeed.

simulationdataschema/model_method.py
ndaelman-hu: Haven't used too much our JSON types yet. So essentially, we interact with it like a dictionary with a JSON conversion?

simulationdataschema/model_method.py
ndaelman-hu: Okay, let me check this tomorrow. Feel free to remind me if I forget.
For now, you can resolve this discussion, it's fine.

simulationdataschema/model_method.py
JosePizarro3: Yes, something like:
```
high_symmetry_points = {
    'Gamma': [0, 0, 0],
    'X': [0.5, 0, 0],
    ...
}
```

simulationdataschema/model_method.py
JosePizarro3: Ok, I defined (just copy-pasting it here for simplicity):
```python
    spacing = Quantity(
        type=MEnum('Equidistant', 'Logarithmic', 'Tan'),
        default='Equidistant',
        description="""
        Identifier for the spacing of the Mesh. Defaults to 'Equidistant' if not defined. It can take the values:

        | Name      | Description                      |
        | --------- | -------------------------------- |
        | `'Equidistant'`  | Equidistant grid (also known as 'Newton-Cotes') |
        | `'Logarithmic'`  | log distance grid |
        | `'Tan'`  | Non-uniform tan mesh for grids. More dense at low abs values of the points, while less dense for higher values |
        """,
    )

    center = Quantity(
        type=MEnum('Gamma-centered', 'Monkhorst-Pack', 'Gamma-offcenter'),
        description="""
        Identifier for the center of the Mesh:

        | Name      | Description                      |
        | --------- | -------------------------------- |
        | `'Gamma-centered'` | Regular mesh is centered around Gamma. No offset. |
        | `'Monkhorst-Pack'` | Regular mesh with an offset of half the reciprocal lattice vector. |
        | `'Gamma-offcenter'` | Regular mesh with an offset that is neither `'Gamma-centered'`, nor `'Monkhorst-Pack'`. |
        """,
    )

    quadrature = Quantity(
        type=MEnum(
            'Gauss-Legendre',
            'Gauss-Laguerre',
            'Clenshaw-Curtis',
            'Newton-Cotes',
            'Gauss-Hermite',
        ),
        description="""
        Quadrature rule used for integration of the Mesh. This quantity is relevant for 1D meshes:

        | Name      | Description                      |
        | --------- | -------------------------------- |
        | `'Gauss-Legendre'` | Quadrature rule for integration using Legendre polynomials |
        | `'Gauss-Laguerre'` | Quadrature rule for integration using Laguerre polynomials |
        | `'Clenshaw-Curtis'`  | Quadrature rule for integration using Chebyshev polynomials using discrete cosine transformations |
        | `'Gauss-Hermite'`  | Quadrature rule for integration using Hermite polynomials |
        """,
    )
```

ndaelman-hu: Looking nice!

JosePizarro3:
Thanks a lot for the reviews @JFRudzinski @ndaelman-hu !! 🥳 

I merge this and work on other minor topics of the repo.

Issue #14, opened 2024-03-04
Plugin distribution as pypi package


JosePizarro3:
We need to align the way of developing a plugin as an independent package. I will check the [nomad-analysis repo](https://github.com/FAIRmat-NFDI/nomad-analysis) and try to reproduce the same settings here.

PR #15, opened 2024-03-04
Adding unit testing for ModelSystem and AtomsState

### LINKED ISSUE #9: Adding unit testing
Following @ndaelman-hu and Alvin's suggestion, we will add unit testing for the different classes and functions.

We can follow Alvin's implementation in the [test of the simulation workflow schema](https://github.com/nomad-coe/nomad-schema-plugin-simulation-workflow/blob/develop/tests/test_simulationworkflowschema.py).

This will be tackled separately, adding unit testing for:
- [x] `model_system.py`, `atoms_state.py` (#15)
- [x] `model_method.py` (@ndaelman-hu needs to tackle the DFT schema and add testing #13)
- [x] `numerical_settings.py` (to be tackled separately together with the work on `BasisSet`, @ndaelman-hu)
- [x] `general.py`
- [x] `outputs.py`

This testing will probably polish the current functionalities, and perhaps add new ones that are missing.

JosePizarro3:
Adding unit testing for the `model_system.py` and `atoms_state.py`


JosePizarro3:
@ladinesa @ndaelman-hu before continuing adding tests, would you mind checking these I prepared? Do they align with the idea you have for doing unit testing?

When preparing them, I wanted to test each of the functions defined for the base classes (including `normalize()`).

tests/test_atoms_state.py
```diff
@@ -0,0 +1,386 @@
...
+    @staticmethod
+    def add_u_j(hubbard_interactions, u, j) -> None:
+        """Adds `u_interaction` and `j_local_exchange_interaction` (in eV) to the `HubbardInteractions` section."""
+        if u is not None:
+            hubbard_interactions.u_interaction = u * ureg('eV')
+        if j is not None:
+            hubbard_interactions.j_local_exchange_interaction = j * ureg('eV')
+
+    @pytest.fixture(autouse=True)
+    def hubbard_interactions(self) -> HubbardInteractions:
```
ladinesa: not sure what this fixture is all about

tests/test_atoms_state.py
```diff
@@ -0,0 +1,386 @@
...
+    """
+
+    logger = logging.getLogger(__name__)
+
+    @staticmethod
+    def add_element_information(atom_state, quantity_name, value) -> None:
+        setattr(atom_state, quantity_name, value)
+
+    @pytest.fixture(autouse=True)
+    def atom_state(self) -> AtomsState:
```
ladinesa: also this.

tests/test_atoms_state.py
```diff
@@ -0,0 +1,386 @@
...
+        assert np.isclose(hubbard_interactions.u_interaction.to('eV').magnitude, 3.0)
+        assert hubbard_interactions.slater_integrals is None
+
+
+class TestAtomsState:
+    """
+    Tests the `AtomsState` class defined in atoms_state.py.
+    """
+
+    logger = logging.getLogger(__name__)
```
ladinesa: maybe put this at the header since __name__ will be the same

tests/test_atoms_state.py
```diff
@@ -0,0 +1,386 @@
...
+
+    @staticmethod
+    def add_element_information(atom_state, quantity_name, value) -> None:
+        setattr(atom_state, quantity_name, value)
```
ladinesa: not sure why you would define this one liner method, also are you if you not supposed to use m_set

tests/test_atoms_state.py
```diff
@@ -0,0 +1,386 @@
...
+
+class TestHubbardInteractions:
+    """
+    Test the `HubbardInteractions` class defined in atoms_state.py.
+    """
+
+    logger = logging.getLogger(__name__)
+
+    @staticmethod
+    def add_slater_interactions(hubbard_interactions, slater_integrals) -> None:
```
ladinesa: this one liner static methods are a bit off for me.

tests/test_atoms_state.py
```diff
@@ -0,0 +1,386 @@
...
+                0.5,
+                None,
+                (0.5, None, 1.5),
+            ),
+            (OrbitalsState(l_quantum_number=1), 0.5, 'initial', (None, 1, None)),
+            (OrbitalsState(l_quantum_number=1), 0.5, 'final', (0.5, None, 5.5)),
+            (None, 0.5, None, (0.5, None, None)),
+        ],
+    )
+    def test_normalize(
```
ladinesa, line 226: the naming is a bit general

tests/test_atoms_state.py
```diff
@@ -0,0 +1,386 @@
...
+
+class TestCoreHole:
+    """
+    Test the `CoreHole` class defined in atoms_state.py.
+    """
+
+    logger = logging.getLogger(__name__)
+
+    @pytest.fixture(autouse=True)
+    def core_hole(self) -> CoreHole:
```
ladinesa: this also

tests/test_atoms_state.py
```diff
@@ -0,0 +1,386 @@
...
+            l_quantum_number,
+            ml_quantum_number,
+            ms_quantum_number,
+            j_quantum_number,
+            mj_quantum_number,
+        )
+        resolved_degeneracy = orbital_state.resolve_degeneracy()
+        assert resolved_degeneracy == degeneracy
+
+    def test_normalize(self, orbital_state):
```
ladinesa: see other comment

ladinesa (commented):
the tests are written in the way i expected, did not check the coverage so please have a look,

JosePizarro3:
> the tests are written in the way i expected, did not check the coverage so please have a look,

Great, thanks a lot!

Indeed, I am aligning the repo with the other plugins #14, so I will check the coverage.

tests/test_atoms_state.py
JosePizarro3: Well, I was thinking these could be used to add certain numbers directly to the section passed. To test in other testing units.

tests/test_atoms_state.py
JosePizarro3: Deleted here and elsewhere.

Indeed, it is a pretty dumb fixture.

tests/test_atoms_state.py
JosePizarro3: Done

tests/test_atoms_state.py
JosePizarro3: Well, any suggestions?

tests/test_atoms_state.py
JosePizarro3: (just resolving in favor of the other comment)

tests/test_atoms_state.py
JosePizarro3: I ended up adding it to `tests/__init__.py`

tests/test_atoms_state.py
JosePizarro3: Ok, I saw I cannot do in another module something like:
```python
from .test_atoms_state import TestHubbardInteractions
...
TestHubbardInteractions.add_slater_interactions()
...
```

as this will run the tests of that class.

tests/test_atoms_state.py
JosePizarro3: Also, there was some issue by using `logging` to the get the name. I ended up using the `get_logger` utils function, which seems to be compatible with typing `BoundLogger`, and more safe.

tests/test_atoms_state.py
JosePizarro3: I deleted the ones that were used once.

JosePizarro3:
@ndaelman-hu would you mind having a look at this? I prepared some testing for `atoms_state.py` and `model_system.py`. I will initially merge this, then continue adding testing for `model_method`.

JosePizarro3:
@ndaelman-hu would you mind taking a look on this? Just a quick check, I have to still rebase to your conftest, and keep writing tests for ModelSystem, it is more the general structure, if you agree or not.

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+    @pytest.mark.parametrize(
+        'number, values, results',
+        [
+            ('n_quantum_number', [-1, 0, 1, 2], [False, False, True, True]),
+            ('l_quantum_number', [-2, 0, 1, 2], [False, False, True, True]),
+            # l_quantum_number == 2 when testing 'ml_quantum_number'
+            ('ml_quantum_number', [-3, 5, -2, 1], [False, False, True, True]),
+            ('ms_quantum_number', [0, 10, -0.5, 0.5], [False, False, True, True]),
+        ],
+    )
```
ndaelman-hu: I would extend the coverage here, since you can be exhaustive (at least for the `True` responses: the `False` are a pain).
Maybe a series of nested `for`-loops are more appropriate here than the `pytest.mark.parametrize` decorator?
Question: _what about states that contain multiple electrons?_

ndaelman-hu: I mean now you've only covered `n=2`.

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+        """
+        Test the quantum number check for the `OrbitalsState` section.
+        """
+        orbital_state = OrbitalsState(n_quantum_number=2)
+        for val, res in zip(values, results):
+            if number == 'ml_quantum_number':
+                orbital_state.l_quantum_number = 2
+            setattr(orbital_state, number, val)
+            check = orbital_state._check_quantum_numbers(logger)
+            assert check == res
```
ndaelman-hu: Smaller notes:
- maybe add typing. Especially, having the variable name `number` for a `str` is kinda confusing.
- I would be a bit more verbose here, regarding the explanation. In the description of the function you say that it checks "physicality", but given the origin of the spherical harmonics, I'd say that _inconsistent_ is the better word.

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+            ('ml', 'number', -1, 'symbol', 'x'),
+            ('ml', 'number', 0, 'symbol', 'z'),
+            ('ml', 'number', 1, 'symbol', 'y'),
+            ('ml', 'number', -2, 'symbol', None),
+            ('ms', 'number', -0.5, 'symbol', 'down'),
+            ('ms', 'number', 0.5, 'symbol', 'up'),
+            ('ms', 'number', -0.75, 'symbol', None),
+            ('no_attribute', 'number', None, 'symbol', None),
+        ],
+    )
```
ndaelman-hu: If `number` and `symbol` never change, then why bother listing them?

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+    def test_number_and_symbol(
+        self,
+        quantum_name,
+        quantum_type,
+        value,
+        countertype,
```
ndaelman-hu: underscore as a separator? or better even, remove this along with `quantum_type`

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+        self,
+        l_quantum_number,
+        ml_quantum_number,
+        j_quantum_number,
+        mj_quantum_number,
+        ms_quantum_number,
+        degeneracy,
+    ):
+        """
+        Test the degeneracy of each of orbital states defined in the parametrization.
```
ndaelman-hu: typo: too many "of"

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+        """
+        orbital_state = OrbitalsState(n_quantum_number=2)
+        self.add_state(orbital_state, 2, -2, None, None, None)
+        orbital_state.normalize(None, logger)
+        assert orbital_state.n_quantum_number == 2
+        assert orbital_state.l_quantum_number == 2
+        assert orbital_state.l_quantum_symbol == 'd'
+        assert orbital_state.ml_quantum_number == -2
+        assert orbital_state.ml_quantum_symbol == 'xy'
+        assert orbital_state.degeneracy == 2
```
ndaelman-hu: This is redundant given the tests above. The only thing you could (but don't have to) check is whether the states with `None` are properly normalized. In that case, use `pytest.mark.parametrize`

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+        'l_quantum_number, ml_quantum_number, j_quantum_number, mj_quantum_number, ms_quantum_number, degeneracy',
+        [
+            (1, None, None, None, 0.5, 3),
+            (1, None, None, None, None, 6),
+            (1, -1, None, None, 0.5, 1),
+            (1, -1, None, None, None, 2),
+            # ! these J and MJ tests are unphysical; check what is going on
+            (1, -1, [1 / 2, 3 / 2], None, None, 6),
+            (1, -1, [1 / 2, 3 / 2], [-3 / 2, 1 / 2, 1 / 2, 3 / 2], None, 2),
+        ],
```
ndaelman-hu: > these J and MJ tests are unphysical; check what is going on

Pls elaborate.

I'd say that one can compute `j` and `mj` given `l` and `ms`, and I would add this to the normalizer + tests.
Conversely, why are the tested `j` and `mj` arrays rather than numbers? I'd say it's more so the reverse. More so, I had a much more sophisticated normalization in the `CoreHole`. Why discard that?

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+    )
+    def test_occupation(self, orbital_ref, degeneracy, n_excited_electrons, occupation):
+        """
+        Test the occupation of a core hole for a given set of orbital reference and degeneracy.
+        """
+        core_hole = CoreHole()
+        core_hole.orbital_ref = orbital_ref
+        if orbital_ref is not None:
+            assert orbital_ref.resolve_degeneracy() == degeneracy
+        core_hole.n_excited_electrons = n_excited_electrons
```
ndaelman-hu: Why not perform all the setting at instantiation?

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+        ],
+    )
+    def test_normalize(self, orbital_ref, n_excited_electrons, dscf_state, results):
+        """
+        Test the normalization of the `CoreHole`. Inputs are defined as the quantities of the `CoreHole` section.
+        """
+        core_hole = CoreHole()
+        core_hole.orbital_ref = orbital_ref
+        core_hole.n_excited_electrons = n_excited_electrons
+        core_hole.dscf_state = dscf_state
```
ndaelman-hu: Why not perform all the setting at instantiation?

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+        resolved_occupation = core_hole.resolve_occupation(logger)
+        if resolved_occupation is not None:
+            assert np.isclose(resolved_occupation, occupation)
+        else:
+            assert resolved_occupation == occupation
+
+    @pytest.mark.parametrize(
+        'orbital_ref, n_excited_electrons, dscf_state, results',
+        [
+            (OrbitalsState(l_quantum_number=1), -0.5, None, (-0.5, None, None)),
```
ndaelman-hu, line 213: Whut? This option should not be allowed.

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+            (OrbitalsState(l_quantum_number=1), 0.5, None, (0.5, None, 5.5)),
+            (
+                OrbitalsState(l_quantum_number=1, ml_quantum_number=-1),
+                0.5,
+                None,
+                (0.5, None, 1.5),
+            ),
```
ndaelman-hu, line 220: I think that this case is already sufficiently covered by the previous test.

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+            ('Cu', 29),
+            ('O', 8),
+        ],
+    )
+    def test_chemical_symbol_and_atomic_number(self, chemical_symbol, atomic_number):
+        """
+        Test the `chemical_symbol` and `atomic_number` resolution for the `AtomsState` section.
+        """
+        # Testing `chemical_symbol`
+        atom_state = AtomsState()
```
ndaelman-hu: You can write this test in 2 lines.

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+    Tests the `AtomsState` class defined in atoms_state.py.
+    """
+
+    @pytest.mark.parametrize(
+        'chemical_symbol, atomic_number',
+        [
+            ('Fe', 26),
+            ('H', 1),
+            ('Cu', 29),
+            ('O', 8),
```
ndaelman-hu, line 353: Ordered by atomic number?

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+        if None not in (u_interaction, u_interorbital_interaction, j_hunds_coupling):
+            assert np.isclose(u_interaction.to('eV').magnitude, results[0])
+            assert np.isclose(u_interorbital_interaction.to('eV').magnitude, results[1])
+            assert np.isclose(j_hunds_coupling.to('eV').magnitude, results[2])
+        else:
+            assert (
+                u_interaction,
+                u_interorbital_interaction,
+                j_hunds_coupling,
+            ) == results
```
ndaelman-hu, line 290: How about you define your own comparison that combines `np.isclose` and `==`?
Alternatively, I remember older tests using `approximate` from `pytest`.

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+            hubbard_interactions.u_interaction = u * ureg('eV')
+        if j is not None:
+            hubbard_interactions.j_local_exchange_interaction = j * ureg('eV')
+
+    @pytest.mark.parametrize(
+        'slater_integrals, results',
+        [
+            ([3.0, 2.0, 1.0], (0.1429146, -0.0357286, 0.0893216)),
+            (None, (None, None, None)),
+            ([3.0, 2.0, 1.0, 0.5], (None, None, None)),
```
ndaelman-hu, line 258: What does this last example test? An incorrect no. passed integrals by whom?
I'd rather see more integrals tested..?

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+                u_interaction,
+                u_interorbital_interaction,
+                j_hunds_coupling,
+            ) == results
+
+    @pytest.mark.parametrize(
+        'u_interaction, j_local_exchange_interaction, u_effective',
+        [
+            (3.0, 1.0, 2.0),
+            (-3.0, 1.0, -4.0),
```
ndaelman-hu: At least in DFT+U, a negative U should not be an option.

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+            (3.0, None, None),
```
ndaelman-hu: here, `Ueff == 3.0`

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+        ],
+    )
+    def test_u_effective(
+        self,
+        u_interaction,
+        j_local_exchange_interaction,
+        u_effective,
+    ):
+        """
+        Test the effective Hubbard interaction `Ueff` for a given set of Hubbard interactions `U` and `J`.
```
ndaelman-hu: `U_eff`

tests/test_atoms_state.py
```diff
@@ -0,0 +1,350 @@
...
+        # ? Is this enough for testing? Can we do more?
+        hubbard_interactions = HubbardInteractions()
+        self.add_u_j(hubbard_interactions, 3.0, 2.0)
+        hubbard_interactions.u_interorbital_interaction = 1.0 * ureg('eV')
+        hubbard_interactions.j_hunds_coupling = 2.0 * ureg('eV')
+
+        hubbard_interactions.normalize(None, logger)
+        assert np.isclose(hubbard_interactions.u_effective.to('eV').magnitude, 1.0)
+        assert np.isclose(hubbard_interactions.u_interaction.to('eV').magnitude, 3.0)
+        assert hubbard_interactions.slater_integrals is None
```
ndaelman-hu: The only extra step on the normalizer's side is checking the length of `u_matrix`.
Checking logs isn't that easy (maybe via the logger object?), and is kind of redundant here.

I think this can be removed.

ndaelman-hu (commented):
(Disclaimer: only checked the `AtomState` tester. Will do `ModelSystem` some other time.)
Overall, it lies in-line with my understanding of unit testing, yes. Good work.

Please double-check the test values I marked. I think those need to be tweaked.

tests/test_atoms_state.py
JosePizarro3: > Question: what about states that contain multiple electrons?

Here I am only defining orbital states, not the electronic one. I think the electronic could be define a posteriori using some normalization, that's why I open a discussion #22, so that we can safely define electronic state **from** `OrbitalsState`

tests/test_atoms_state.py
JosePizarro3: indeed :) At some point I had some other testing and this changed and now `number` and `symbol` are useless

tests/test_atoms_state.py
JosePizarro3: The whole calculation using `RussellSanders` work on the electronic state, but not for single orbitals. I would rework a bit how we plan to do this, we can work on #22.

For now, I will improve the comment.

tests/test_atoms_state.py
JosePizarro3: Well, I wanted to check the execution of `normalize`, right? Just with a controlled case to see if the logic runs as expected.

tests/test_atoms_state.py
JosePizarro3: True, I moved it.

tests/test_atoms_state.py
JosePizarro3: True, I moved it

tests/test_atoms_state.py
JosePizarro3: Well, the logger captures an error in `normalize`, but for testing I kept the degeneracy and occupation to None, so it does not run (line 375 in `atoms_state.py`).

tests/test_atoms_state.py
JosePizarro3: I just want to illustrate the difference in occupation when breaking the ml degeneracy

tests/test_atoms_state.py
JosePizarro3: well, not so important I think 😄

tests/test_atoms_state.py
JosePizarro3: well, but the functionality does not support it, there is a warning in the function.

JosePizarro3:
Ok, this is ready for merging I think. I added typing, and improved a bit on the `atoms_state.py`.

@ndaelman-hu did you have time to check on `test_model_system.py`? If not no worries, I can wait.

JosePizarro3:
@ndaelman-hu I will double check the conflicts, but this thing is ready and it might influence a bit what you are doing with ModelSystem, just fyi.

Maybe I can double check and merge directly?

ndaelman-hu:
> @ndaelman-hu I will double check the conflicts, but this thing is ready and it might influence a bit what you are doing with ModelSystem, just fyi.
> 
> Maybe I can double check and merge directly?

Thank you for pointing this out. Overall, the only potentially conflicting files are the src code (I'll check when opening my PR).
Overall, there's little pressure from my side, as I will be out for a week. We can coordinate both PRs (if necessary) when I'm back.
Of course, feel free to close this PR with the help of @ladinesa, no prob.

JosePizarro3:
@ladinesa would you mind giving a last final quick look? I understand this pr is big so it will take time, but I am just aiming for a quick double-check (the tests are already thought of).

I can then continue adding test to the ModelMethod.

ladinesa (approved):

PR #16, opened 2024-03-05
Plugin distribution as pypi package

### LINKED ISSUE #14: Plugin distribution as pypi package

JosePizarro3:
Closes #14


JosePizarro3:
In order to speed up the plugin dev, I am merging this myself.

@JFRudzinski maybe this interests you. The way of defining the plugin have changed a bit. I added some small instructions in the README.md as well, which should be enough to use imports from this kind of plugin.

@ladinesa this is coming from talking with Area A people. I am still keeping this as a submodule for launching the nomad worker, but perhaps there is a way around in which this will work (without the need of adding this repo to `.gitmodules`).

PR #29, opened 2024-04-02
 Add `Outputs` base class

### LINKED ISSUE #27: Add `Outputs` base class
@ndaelman-hu @JFRudzinski 

This is the issue to add the basic `Outputs` section. I will call for review once it is ready.

Some basic ideas:
- It should contain `model_system_ref` (repeats=False), `atoms_state_ref` (repeats=True), `orbitals_state_ref` (repeats=True).
- It should contain `is_derived` and `outputs_ref` for when a property is derived during normalization.
- It should contain `is_converged` and be connected with `numerical_settings.SelfConsistency`

I leave out handling times for another issue.

JosePizarro3:
Adding base class for outputs and the unit testing. Testing is a bit empty, but I guess we will see more once we implement other cases.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ license = { text = "Apache-2.0" }
 requires-python = ">=3.9"
 dependencies = [
     "nomad-lab>=1.2.0",
-    'matid>=2.0.0.dev2'
+    "matid>=2.0.0.dev2",
 ]
 
 [project.urls]
@@ -35,6 +35,7 @@ dev = [
     'pytest-cov==2.7.1',
     'ruff==0.1.8',
     "structlog==22.3.0",
+    "lxml_html_clean>=0.1.0",
 ]
 
 [tool.ruff]
--- a/src/nomad_simulations/outputs.py
+++ b/src/nomad_simulations/outputs.py
@@ -15,38 +15,139 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-#
-# Copyright The NOMAD Authors.
-#
-# This file is part of NOMAD.
-# See https://nomad-lab.eu for further info.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-#
 
-import re
 import numpy as np
+from typing import Optional
+from structlog.stdlib import BoundLogger
 
-from nomad.units import ureg
 from nomad.datamodel.data import ArchiveSection
-from nomad.metainfo import Quantity, SubSection, SectionProxy, MEnum
+from nomad.metainfo import Quantity, SubSection, SectionProxy
+
+from .atoms_state import AtomsState, OrbitalsState
+from .model_system import ModelSystem
+from .numerical_settings import SelfConsistency
 
 
-class Outputs(ArchiveSection):
-    """ """
+class BaseOutputs(ArchiveSection):
+    """
+    Base section to define the output properties of a simulation. This is used as a placeholder for both
+    final `Outputs` properties and the self-consistent (SCF) steps, see `Outputs` base section definition.
+    """
 
     normalizer_level = 2
 
+    orbitals_state_ref = Quantity(
+        type=OrbitalsState,
+        description="""
+        Reference to the `OrbitalsState` section on which the simulation is performed and the
+        output property is calculated.
+        """,
+    )
+
+    atoms_state_ref = Quantity(
+        type=AtomsState,
+        description="""
+        Reference to the `AtomsState` section on which the simulation is performed and the
+        output property is calculated.
+        """,
+    )
+
+    model_system_ref = Quantity(
+        type=ModelSystem,
+        description="""
+        Reference to the `ModelSystem` section on which the simulation is performed and the
+        output property is calculated.
+        """,
+    )
+
+    is_derived = Quantity(
+        type=bool,
+        default=False,
+        description="""
+        Flag indicating whether the output property is derived from other output properties. We make
+        the distinction between directly parsed, derived, and post-processing output properties:
+            - Directly parsed: the output property is directly parsed from the simulation output files.
+            - Derived: the output property is derived from other output properties. No extra numerical settings
+                are required to calculate the output property.
+            - Post-processing: the output property is derived from other output properties. Extra numerical settings
+                are required to calculate the output property.
+        """,
+    )
+
+    outputs_ref = Quantity(
+        type=SectionProxy('BaseOutputs'),
+        description="""
+        Reference to the `BaseOutputs` section from which the output property was derived. This is only
+        relevant if `is_derived` is set to True.
+        """,
+    )
+
+    def check_is_derived(self, is_derived: bool, outputs_ref) -> Optional[bool]:
+        if not is_derived and outputs_ref is not None:
+            return True
+        if is_derived and outputs_ref is None:
+            return None
+        return False
+
     def normalize(self, archive, logger) -> None:
         super().normalize(archive, logger)
-        self.logger = logger
+
+        check_derived = self.check_is_derived(self.is_derived, self.outputs_ref)
+        if check_derived is not None:
+            self.is_derived = check_derived
+        else:
+            logger.error(
+                'A derived output property must have a reference to another `Outputs` section.'
+            )
+            return
+
+
+class Outputs(BaseOutputs):
+    """
+    Output properties of a simulation.
+
+    # ! add more description once we defined the output properties
+    """
+
+    n_scf_steps = Quantity(
+        type=np.int32,
+        description="""
+        Number of self-consistent steps to converge the output property.
+        """,
+    )
+
+    scf_step = SubSection(
+        sub_section=BaseOutputs.m_def,
+        repeats=True,
+        description="""
+        Self-consistent (SCF) steps performed for converging a given output property.
+        """,
+    )
+
+    is_converged = Quantity(
+        type=bool,
+        description="""
+        Flag indicating whether the output property is converged or not. This quantity is connected
+        with `SelfConsistency` defined in the `numerical_settings.py` module.
+        """,
+    )
+
+    self_consistency_ref = Quantity(
+        type=SelfConsistency,
+        description="""
+        Reference to the `SelfConsistency` section that defines the numerical settings to converge the
+        output property.
+        """,
+    )
+
+    # ? Can we add more functionality to automatically check convergence from `self_consistency_ref` and the last `scf_step[-1]`
+    def check_is_converged(self, is_converged: bool, logger: BoundLogger) -> bool:
+        if not is_converged:
+            logger.info('The output property is not converged.')
+            return False
+        return True
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
+
+        self.is_converged = self.check_is_converged(self.is_converged, logger)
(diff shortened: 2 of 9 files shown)


JosePizarro3:
(please, disregard the failing pipeline; tests are passing locally)

ndaelman-hu:
will review after lunch

src/nomad_simulations/outputs.py
```diff
@@ -15,38 +15,153 @@
...
+        else:
+            logger.error(
+                'A derived output property must have a reference to another `Outputs` section.'
+            )
+            return
+
+
+class Outputs(BaseOutputs):
+    """
+    Output properties of a simulation.
```
JFRudzinski: Can we be clearer here and throughout the new schema what we mean by simulation? I guess different people (even between us) may have different interpretations, e.g., what you call "simulation" I would use the generic word "calculation" I think...not sure may depend on the context

src/nomad_simulations/outputs.py
```diff
@@ -15,38 +15,153 @@
...
+
+    # ! add more description once we defined the output properties
+    """
+
+    n_scf_steps = Quantity(
```
JFRudzinski, line 146: with all these scf-specific quantities, this seems like a more specific output property class. Additionally, it's also a little confusing to me how something like "Number of self-consistent steps to converge the output property" is not a workflow property. This is dependent on the output property value of several individual model systems, correct?

JFRudzinski (approved):

src/nomad_simulations/outputs.py
JosePizarro3: It was decided before joining FAIRmat, I think, that all scf steps will belong to the same `SinglePoint` concept, just on the same basis that breaking System into individual entries does not make too much sense from the computational side.

But yes, being very strict, scf steps are a workflow. I will add more description into these quantities and link this with the `SinglePoint` concept in NOMAD. Eventually, these things should be carefully described in the documentation page.

ndaelman-hu: I would keep SCF with the same entry, since we consider single-point calculations as an elementary task.

> this seems like a more specific output property class

Agreed. If it uses the same concepts, you can inherit, else keep them separate.

src/nomad_simulations/outputs.py
```diff
@@ -15,38 +15,153 @@
...
+        description="""
+        Flag indicating whether the output property is derived from other output properties. We make
+        the distinction between directly parsed, derived, and post-processing output properties:
+            - Directly parsed: the output property is directly parsed from the simulation output files.
+            - Derived: the output property is derived from other output properties. No extra numerical settings
+                are required to calculate the output property.
+            - Post-processing: the output property is derived from other output properties. Extra numerical settings
+                are required to calculate the output property.
+        """,
+    )
```
ndaelman-hu: 1. I think that a generic reference may be appropriate here, or can you guarantee that the option will always be binary?
2. I would not mention `post-processing` here, since it's part of a separate plugin. We can, however, use this comparison in our area-specific docs.

src/nomad_simulations/outputs.py
JosePizarro3: Well, I think all these are synonyms: "simulation", "calculation", "computation". For me it is easier to think of "Calculation" as any generic mathematical operation which produces a result, while "Simulation" implies an idealization for modelling purposes of some real world case.

src/nomad_simulations/outputs.py
ndaelman-hu: Btw, on point 1, this is an interesting concept:
we all agree that FAIR data needs provenance information, but here we're actually logging internal provenance steps.

src/nomad_simulations/outputs.py
JosePizarro3: I will dig into the ontologies being developed, but I saw [MDO actually uses Calculation](https://liusemweb.github.io/mdo/core/1.1/index.html#Calculation) (as it appears in our `BaseSimulation(Activity)` class in the general.py module).

JFRudzinski: Sorry, if I wasn't clear. I am totally fine with "simulation", I just wanted to know how you define it (which you did above). We should just state this somewhere in one of the base class descriptions I think.

ndaelman-hu: > I think all these are synonyms

You're right, but I think that towards the users we should provide a clear lingo. The discussion on where each term falls is quite ontological and might bring us too far. I'd just remain consistent and provide some kind of definition.

Personally, I use _calculation_ and _computation_ interchangeably, though in this context I take some offense with calling any mathematical operation (e.g. simple addition) a _calculation_. Anything that uses ab initio data, as well as MD, is what I end up calling _simulation_.

JosePizarro3: Yeah, I'll add it under `Simulation`, and improve this description for `Outputs` anyways to make the SCF thingy more clear.

src/nomad_simulations/outputs.py
```diff
@@ -15,38 +15,153 @@
...
+
+    def check_is_derived(self, is_derived: bool, outputs_ref) -> Optional[bool]:
+        if not is_derived:
+            if outputs_ref is not None:
+                return True
+            return False
+        elif is_derived and outputs_ref is not None:
+            return True
+        return None
+
```
ndaelman-hu: Not sure if I follow the logic here (maybe also add a small pydocs):

This function is used to set `is_derived`, correct?
If `is_derived` is already set (i.e. `is not None`), only then it should trigger.

Moreover, if `outputs_ref` fully covers `is_derived`, then you don't need the latter.

src/nomad_simulations/outputs.py
```diff
@@ -15,38 +15,153 @@
...
+        """,
+    )
+
+    self_consistency_ref = Quantity(
+        type=SelfConsistency,
+        description="""
+        Reference to the `SelfConsistency` section that defines the numerical settings to converge the
+        output property.
+        """,
+    )
```
ndaelman-hu: Call it `is_scf_converged`.

I understand your approach here. Probably, it's clearer for people extending the schema to rigorously enforce the distinction _method_ and _output_.
Still, it feels kind of weird to have `is_converged` exist separately from its specification...

src/nomad_simulations/outputs.py
JFRudzinski: I wasn't arguing to split into different entries, but now I am questioning a bit my understanding of the plan. We are planning to keep the current basic structure for Outputs right? Where Outputs is a list with each member corresponding to a single configuration? If this is the case I don't see how `n_scf_steps` makes sense here. Doesn't it need to be stored at a higher level?

My bigger concern is about the specificity of this class. The Outputs class sounds general, but these quantities are specifically referring to some scf workflow (whether it is stored in workflow or not).

src/nomad_simulations/outputs.py
```diff
@@ -15,38 +15,153 @@
...
+
+    # ? Can we add more functionality to automatically check convergence from `self_consistency_ref` and the last `scf_step[-1]`
```
ndaelman-hu: Yes, but you'd need access to the SCF steps.

src/nomad_simulations/outputs.py
```diff
@@ -15,38 +15,153 @@
...
+    def check_is_converged(self, is_converged: bool, logger: BoundLogger) -> bool:
+        if not is_converged:
+            logger.info('The output property is not converged.')
```
ndaelman-hu: Lots of people ask me for these kinds of warnings, so I'm glad you added them.
I just fear that it might end up getting buried in `info`. We should look into a better semantic category.

src/nomad_simulations/outputs.py
JosePizarro3: You mean defining another level to contain the outputs which have required some sort of SCF process? Something like `SCFOutputs(Outputs)` and not defining the scf steps here in the `Outputs` level?

JosePizarro3: Ok, my bad, that does not make any sense: `Outputs` and `BaseOutputs` already handle that situation, maybe I should rename them to `SCFOutputs` and `Outputs`, respectively?

src/nomad_simulations/outputs.py
ndaelman-hu: One more point, I'm going to ask Area D for "write-protected" quantities, i.e. if the value has been set, it can't be updated. This would help make the normalization leaner.

src/nomad_simulations/outputs.py
JFRudzinski: Yeah, something like that. I am just wondering what Class I am going to use for my data, I don't think these scf quantities are relevant to me, so if it makes sense to you, I would like to use a higher-level class that doesn't include them

tests/test_outputs.py
```diff
@@ -0,0 +1,73 @@
...
+    )
+    def test_normalize(self, is_converged, result):
+        """
+        Test the `normalize` method.
+        """
+        outputs = Outputs()
+        assert outputs.check_is_converged(is_converged, logger) == result
+        outputs.is_converged = is_converged
+        outputs.normalize(None, logger)
+        assert outputs.is_converged == result
```
ndaelman-hu: These tests appear very similar. Can they be combined?

ndaelman-hu (approved):
Just some minor points, but I agree with most of the approach.

src/nomad_simulations/outputs.py
JosePizarro3: Yes, I change the naming then, and you can use `Outputs` (the current `BaseOutputs`) directly

src/nomad_simulations/outputs.py
JosePizarro3: No, because it might be missing `outputs_ref`, hence the `None` return describes the "missing refs" case. I added a better description after you reviewed it, sorry.

src/nomad_simulations/outputs.py
JosePizarro3: Yes, I leave this function defined but also with a TODO to add these checks in a future when we have a clear use-case

src/nomad_simulations/outputs.py
JosePizarro3: Well, `is_scf_converged` will help on this 😉

tests/test_outputs.py
JosePizarro3: Well, I am checking two different methods, but yes, I agree they are pretty meh. I am imagining that in a future `check_is_converged` can be separated in its own method (with more complex logic) with more complex logic. I'll add a comment here

src/nomad_simulations/outputs.py
JosePizarro3: >  can you guarantee that the option will always be binary?

Yes, it is just that we need to clarify that there is a third option. But my argument: let's see how the `nomad-analysis` plugin evolves and how we fit everything altogether.

src/nomad_simulations/outputs.py
ndaelman-hu: No, the requests are more generic than just SCF. Kind of, raise a flag when some setting is inappropriate.

JosePizarro3: Yeah, indeed. We should use some other category, or some functionality to extract and use logger messages (like the one Sarthak did for some testing).

Anyways, I leave a question mark here for the time being.

JosePizarro3:
Just finished some first version, feel free to do another review.

You can also resolve threads when you feel you satisfied with the resolutions.

pyproject.toml
```diff
@@ -35,6 +35,7 @@ dev = [
     'pytest-cov==2.7.1',
     'ruff==0.1.8',
     "structlog==22.3.0",
+    "lxml_html_clean>=0.1.0",
```
ndaelman-hu, line 38: Does this solve the issue then?

ndaelman-hu (approved):

src/nomad_simulations/outputs.py
ndaelman-hu: Okay, but let's run over the possible scenarios:
1. both `outputs_ref` and `is_derived` are set, no issue
2. only `outputs_ref` is set: you can set `is_derived = True`, but no additional information is conveyed
3. `is_derived = True` and `outputs_ref = None`: **when would this happen?**

JosePizarro3: For number 3., maybe some wrong setting of `is_derived` or some missing `outputs_ref` when normalizing? In this case, `return None` is happening

pyproject.toml
JosePizarro3: Yeah

PR #35, opened 2024-04-02
26 add post hf methods

ndaelman-hu:
Add a 1st version for perturbation theory and coupled cluster.
For future extensions, I'll reopen the original issue.

### DIFF
--- a/src/nomad_simulations/model_method.py
+++ b/src/nomad_simulations/model_method.py
@@ -438,6 +438,54 @@ class CoupledCluster(ModelMethodElectronic):
     A base section used to define the parameters of a Coupled Cluster calculation.
     """
 
+    def check_orders(self, logger) -> bool:
+        """Perform a sanity check on the excitation and perturbation order.
+        Raise a logging error if any inconsistency is found.
+        """
+        if self.excitation_order is None:
+            logger.warning('`CoupledCluster.excitation_order` is undefined.')
+            return False
+        if len(self.excitation_order) > 1:
+            if 2 not in self.excitation_order:
+                logger.error('Coupled Cluster typically starts from doubles.')
+                return False
+        for order in (3, 4):
+            if order in self.excitation_order and order in self.perturbative_order:
+                logger.error(
+                    f'Order {order} is defined as both excitation and perturbative.'
+                )
+                return False
+        return True
+
+    def type_cc(self) -> str:
+        """Produce an educated guess based on the other parameters."""
+        name = 'CC'
+        # cover the basic cases
+        if 2 in self.excitation_order:
+            if 1 in self.excitation_order:
+                name += 'SD'
+            name += 'D'
+        # cover extended excitations
+        for order, abbrev in {3: 'T', 4: 'Q'}.items():
+            if order in self.excitation_order:
+                name += abbrev
+            elif order in self.perturbative_order:
+                name += f'({abbrev})'
+        # cover explicit correlation
+        if self.explicit_correlation is not None:
+            name += self.explicit_correlation
+        # cover specific solver approaches
+        if self.solver == 'quasi-variational':
+            name = 'QV' + name
+        elif self.solver == 'Brueckner':
+            name = 'B' + name
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
+        if self.check_excitation_order(logger):
+            if type is None:
+                self.type = self.type_cc()
+
     type = Quantity(
         type=MEnum(
             *[
@@ -487,7 +535,7 @@ class CoupledCluster(ModelMethodElectronic):
         ),
         default='variational',
         description="""
-        Solvers of the coupled cluster equations.
+        Solvers of / approaches to the coupled cluster equations.
         """,
         a_eln=ELNAnnotation(component='EnumEditQuantity'),
     )
@@ -509,7 +557,7 @@ class CoupledCluster(ModelMethodElectronic):
         which denotes restrictions on the ground state.
         """,
         a_eln=ELNAnnotation(component='BoolEditQuantity'),
-    )  # this should go into a knowledge base
+    )
 
     is_size_extensive = Quantity(
         type=bool,
--- a/src/nomad_simulations/model_method.py
+++ b/src/nomad_simulations/model_method.py
@@ -436,7 +436,12 @@ class PerturbationMethod(ModelMethodElectronic):
 class CoupledCluster(ModelMethodElectronic):
     """
     A base section used to define the parameters of a Coupled Cluster calculation.
+    A standard schema is defined, though the most common cases can be summarized in the `type` quantity.
+    Conversion between both is provided.
     """
+
+    order_map = {k: v for k, v in enumerate(('S', 'D', 'T', 'Q'))}
+    solver_map = {'quasi-variational': 'QV', 'Brueckner': 'B'}
 
     def check_orders(self, logger) -> bool:
         """Perform a sanity check on the excitation and perturbation order.
@@ -457,7 +462,7 @@ class CoupledCluster(ModelMethodElectronic):
                 return False
         return True
 
-    def type_cc(self) -> str:
+    def type_to_cc(self) -> None:
         """Produce an educated guess based on the other parameters."""
         name = 'CC'
         # cover the basic cases
@@ -475,21 +480,42 @@ class CoupledCluster(ModelMethodElectronic):
         if self.explicit_correlation is not None:
             name += self.explicit_correlation
         # cover specific solver approaches
-        if self.solver == 'quasi-variational':
-            name = 'QV' + name
-        elif self.solver == 'Brueckner':
-            name = 'B' + name
+        if self.solver in self.solver_map:
+            name = self.solver_map[self.solver] + name
+
+    def cc_to_type(self) -> None:
+        """Try to extract the excitation and perturbation orders from the type."""
+        self.type = self.type.upper()
+        match = re.match(
+            r'(?:(QV|B))CC(S)?(D)?(?:(T|\(T\)))(?:(Q|\(Q\)))(?:-(F12|R12))', self.type
+        )
+        if match is None:
+            return
+        self.excitation_order = []
+        self.perturbative_order = []
+        for i in range(2, 6):
+            if abbrev := match.group(i):
+                if abbrev[0] == '(':
+                    self.perturbative_order.append(i)
+                else:
+                    self.excitation_order.append(i)
+        if match.group(1) in self.solver_map:
+            self.solver = self.solver_map[match.group(1)]
+        if match.group(6):
+            self.explicit_correlation = match.group(6)
 
     def normalize(self, archive, logger) -> None:
         super().normalize(archive, logger)
         if self.check_excitation_order(logger):
             if type is None:
-                self.type = self.type_cc()
+                self.type_to_cc()
+            else:
+                self.cc_to_type()
 
     type = Quantity(
         type=MEnum(
             *[
-                f'{prefix}CC{base_order}{ext_order}{corr}'
+                f'{prefix}CC{base_order}{ext_order}-{corr}'
                 for prefix in ('', 'QV', 'B')
                 for base_order in ('D', 'SD')
                 for ext_order in ('', '(T)', '(Q)', 'T', 'Q')
@@ -530,8 +556,7 @@ class CoupledCluster(ModelMethodElectronic):
     solver = Quantity(
         type=MEnum(
             'variational',
-            'quasi-variational',
-            'Brueckner',
+            *solver_map.keys(),
         ),
         default='variational',
         description="""
--- a/src/nomad_simulations/model_method.py
+++ b/src/nomad_simulations/model_method.py
@@ -444,85 +444,6 @@ class CoupledCluster(ModelMethodElectronic):
     solver_map = {'QV': 'quasi-variational', 'B': 'Brueckner'}
     map_solver = {v: k for k, v in solver_map.items()}
 
-    def check_orders(self, logger) -> bool:
-        """Perform a sanity check on the excitation and perturbation order.
-        Raise a logging error if any inconsistency is found.
-        """
-        if self.excitation_order is None:
-            logger.warning('`CoupledCluster.excitation_order` is undefined.')
-            return False
-        if len(self.excitation_order) > 1:
-            if 2 not in self.excitation_order:
-                logger.error('Coupled Cluster typically starts from doubles.')
-                return False
-        for order in (3, 4):
-            if order in self.excitation_order and order in self.perturbative_order:
-                logger.error(
-                    f'Order {order} is defined as both excitation and perturbative.'
-                )
-                return False
-        return True
-
-    def cc_to_type(self) -> None:
-        """Produce an educated guess based on the other parameters."""
-        name = 'CC'
-        # cover the basic cases
-        if 2 in self.excitation_order:
-            if 1 in self.excitation_order:
-                name += 'SD'
-            name += 'D'
-        # cover extended excitations
-        for order, abbrev in {3: 'T', 4: 'Q'}.items():
-            if order in self.excitation_order:
-                name += abbrev
-            elif order in self.perturbative_order:
-                name += f'({abbrev})'
-        # cover explicit correlation
-        if self.explicit_correlation is not None:
-            name += self.explicit_correlation
-        # cover specific solver approaches
-        if self.solver in self.map_solver:
-            name = self.solver_map[self.solver] + name
-
-    def type_to_cc(self) -> None:
-        """Try to extract the excitation and perturbation orders from the type."""
-        match = re.match(
-            r'(QV|B)?CC(S)?(D)?(T|\(T\))?(Q|\(Q\))?(-F12|-R12)?', self.type
-        )
-        if match is None:
-            return
-
-        ptb_initialized, exc_initialized = False, False
-        for i in range(2, 6):
-            if abbrev := match.group(i):
-                order = i - 1
-                if abbrev[0] == '(':
-                    if not ptb_initialized:
-                        self.perturbative_order = []
-                        ptb_initialized = True
-                    self.perturbative_order = np.append(self.perturbative_order, order)
-                else:
-                    if not exc_initialized:
-                        self.excitation_order = []
-                        exc_initialized = True
-                    self.excitation_order = np.append(self.excitation_order, order)
-        if match.group(1) in self.solver_map:
-            self.solver = self.solver_map[match.group(1)]
-        if match.group(6):
-            self.explicit_correlation = match.group(6)[1:]  # remove the dash
-
-    def normalize(self, archive, logger) -> None:
-        super().normalize(archive, logger)
-        if self.type is None:
-            if self.check_orders(logger):
-                self.cc_to_type()
-        else:
-            self.type_to_cc()
-        if isinstance(self.excitation_order, list):
-            self.excitation_order = np.sort(self.excitation_order)
-        if isinstance(self.perturbative_order, list):
-            self.perturbative_order = np.sort(self.perturbative_order)
-
     type = Quantity(
         type=MEnum(
             *[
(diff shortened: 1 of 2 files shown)


JosePizarro3:
I will review this next week.

Just a small detail: would you mind keeping the Quantities defined before the methods? Only for readability, it is a bit confusing going down first and up then to check the functionalities. 
Also for mappings, can we use `__init__` as the place to define those (see [atoms_state.py](https://github.com/nomad-coe/nomad-schema-plugin-simulation-data/blob/develop/src/nomad_simulations/atoms_state.py#L132))? So it is not confusing that these are not quantities. If you decide to go in this direction, don't forget to do `super().__init__(m_def, m_context, **kwargs)`.

These are minor things, just trying to be a bit consistent across modules.

ndaelman-hu:
> I will review this next week.

Fine, as long as we don't introduce any breaking changes.

> Just a small detail: would you mind keeping the Quantities defined before the methods? Only for readability, it is a bit confusing going down first and up then to check the functionalities.

Sure, I'll update it.

> Also for mappings, can we use `__init__` as the place to define those (see [atoms_state.py](https://github.com/nomad-coe/nomad-schema-plugin-simulation-data/blob/develop/src/nomad_simulations/atoms_state.py#L132))? So it is not confusing that these are not quantities. If you decide to go in this direction, don't forget to do `super().__init__(m_def, m_context, **kwargs)`.

I'm not entirely sure that I got the last part. Essentially, I don't care too much, though this changes them from class to object variables. It's minor, but they are rather the former. **I suggest that I define them in `__new__` instead.**

ndaelman-hu:
> I'm not entirely sure that I got the last part. Essentially, I don't care too much, though this changes them from class to object variables. It's minor, but they are rather the former. I suggest that I define them in __new__ instead.

Scratch that last part: for referencing them, it's more appropriate to keep them internally.
Btw, I disagree with the confusion with `Quantity`: this is set in the definition.

ndaelman-hu:
@JosePizarro3 Given that the sprint is being extended, I'd appreciate it if you could review this sooner.

ndaelman-hu:
@JosePizarro3 can I go ahead and merge?

JosePizarro3:
> @JosePizarro3 can I go ahead and merge?

No, I am working on other stuff, but I need some time to review this. I will review this once we finish all the issues related to the sprint (probably while I am waiting for review on my stuff, I can manage to review before everything is merged then).

My question would be why you are not working on outputs, and rather spending the time on this tho. **This is far from being a priority**. We already agreed on having a task distribution between us, but I haven't seen anything from your side on output properties.

ndaelman-hu:
> My question would be why you are not working on outputs, and rather spending the time on this tho. **This is far from being a priority**. We already agreed on having a task distribution between us, but I haven't seen anything from your side on output properties.

1. This was done before we started the sprint.
2. As for the outputs, I'm waiting on the `PropertyBaseClass` implementation, as discussed in our last meeting.

JosePizarro3:
> 1. This was done before we started the sprint.

No, I meant the review. It was done last Tuesday **DURING** the sprint.

> 2. As for the outputs, I'm waiting on the PropertyBaseClass implementation, as discussed in our last meeting.

Sure, this is what we talked. You can wait for us to finish this, and work on these outputs. But that does not mean that Joseph or I have time to review this PR. Furthermore, you could try it out now to see if you find the implementation sounding or not, instead of having to wait and then stop the development process for weeks until you grasp the ideas.

Once more (just to make sure we are in the same page): when I finish the `PhysicalProperty` and the outputs related with the ElectronicDOS, Spectra, and GreensFunctions, I will look at this.

src/nomad_simulations/model_method.py
```diff
@@ -87,6 +87,7 @@ class ModelMethodElectronic(ModelMethod):
     """
 
     # ? Is this necessary or will it be defined in another way?
+    ## ND: yes, this is a very necessary parameter, but sometimes the total spin is also restricted
```
JosePizarro3, line 90: Can you add a TODO to prepare some utility function to handle spin-polarized properties and their connection with ModelMethodElectronic?

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
 
 
+class PerturbationMethod(ModelMethodElectronic):
+    type = Quantity(
+        type=MEnum('MP', 'RS', 'BW'),
+        description="""
+        Perturbation approach. The abbreviations stand for:
+        | Abbreviation | Description |
+        | ------------ | ----------- |
+        | `'MP'`       | Moller-Plesset |
```
JosePizarro3, line 412: Would you mind adding a new column with "Reference"?

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
                 )
 
 
+class PerturbationMethod(ModelMethodElectronic):
```
JosePizarro3, line 405: Add description please.

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+        description="""
+        Order up to which the perturbation is expanded.
+        """,
+        a_eln=ELNAnnotation(component='NumberEditQuantity'),
+    )
+
+    is_size_extensive = Quantity(
+        type=bool,
+        description="""
+        If the perturbation method is size extensive.
```
JosePizarro3, line 430: can you explain more?

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+        """,
+        a_eln=ELNAnnotation(component='BoolEditQuantity'),
+    )  # this should go into a knowledge base
+
+
+class CoupledCluster(ModelMethodElectronic):
+    """
+    A base section used to define the parameters of a Coupled Cluster calculation.
```
JosePizarro3, line 438: Improve the description, please.

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+    A standard schema is defined, though the most common cases can be summarized in the `type` quantity.
+    Conversion between both is provided.
+    """
+
+    order_map = {k: v for k, v in enumerate(('S', 'D', 'T', 'Q'))}
```
JosePizarro3, line 443: Move to the init function, please. Just a minor detail, but that increases readability.

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+    solver_map = {'QV': 'quasi-variational', 'B': 'Brueckner'}
+    map_solver = {v: k for k, v in solver_map.items()}
+
+    type = Quantity(
+        type=MEnum(
+            *[
```
JosePizarro3, line 449: This is unreadable, but I understand for efficiency you want to do it like this. However, like we are humans, I'd like to understand: 1) the full list of labels, 2) what each label means (with refs probably). Furthermore, you have some "non-sensical combinations"?

An alternative way is to consider `type` as a `str` and then do the check later in some function. I'd actually go with this option.

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+
+    excitation_order = Quantity(
+        type=np.int32,
+        shape=['*'],
+        description="""
+        Orders at which the excitation are used.
+        1 = single, 2 = double, 3 = triple, 4 = quadruple, etc.
+
+        Note that coupled cluster typically start from doubles.
+        Singles excitations in a Koopman-compliant scheme only make sense as a response to a perturbation.
```
JosePizarro3, line 472: Could you improve this phrase and explain a bit more? Also, the note is not clear why or how.

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+        """,
+        a_eln=ELNAnnotation(component='NumberEditQuantity'),
+    )
+
+    perturbative_order = Quantity(
+        type=np.int32,
+        shape=['*'],
+        description="""
+        Excitation order at which the perturbative correction is used.
```
JosePizarro3, line 481: Maybe explain the diff wrt to the `excitation_order`?

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+        1 = single, 2 = double, 3 = triple, 4 = quadruple, etc.
+        """,
+        a_eln=ELNAnnotation(component='NumberEditQuantity'),
+    )
+
+    perturbative_method = SubSection(sub_section=PerturbationMethod.m_def)
+
+    solver = Quantity(
+        type=MEnum(
```
JosePizarro3, line 490: Again, I see MEnums as a way of us, to understand which values are covered. Right now it is complicated, and it would be much simpler to add the full list and not `solver_map.values()` (which are only 2 options)

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+            'variational',
+            *solver_map.values(),
+        ),
+        default='variational',
+        description="""
+        Solvers of / approaches to the coupled cluster equations.
```
JosePizarro3, line 496: Improve description please.

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+        """,
+        a_eln=ELNAnnotation(component='EnumEditQuantity'),
+    )
+
+    explicit_correlation = Quantity(
+        type=MEnum('F12', 'R12', ''),
+        default='',
+        description="""
+        Explicit correlation treatment.
```
JosePizarro3, line 505: Improve description, I don't understand what `explicit` means here.

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+        ),
+        default='variational',
+        description="""
+        Solvers of / approaches to the coupled cluster equations.
+        """,
+        a_eln=ELNAnnotation(component='EnumEditQuantity'),
+    )
+
+    explicit_correlation = Quantity(
+        type=MEnum('F12', 'R12', ''),
```
JosePizarro3, line 502: MEnum empty string? No, delete it and also the default.

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+        Explicit correlation treatment.
+        """,
+        a_eln=ELNAnnotation(component='EnumEditQuantity'),
+    )
+
+    is_unrestricted = Quantity(
+        type=bool,
+        description="""
+        Denotes whether the excitations are unrestricted or not.
+        It should be confused with `is_spin_polarized`,
```
JosePizarro3, line 514: Improve description. And explain a bit more, I don't know what it means that the excitations are restricted or not

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+        which denotes restrictions on the ground state.
+        """,
+        a_eln=ELNAnnotation(component='BoolEditQuantity'),
+    )
+
+    is_size_extensive = Quantity(
```
JosePizarro3, line 520: Not sure what this quantity is. It also appears in `PerturbationMethod`?

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+    def check_orders(self, logger) -> bool:
+        """Perform a sanity check on the excitation and perturbation order.
+        Raise a logging error if any inconsistency is found.
+        """
+        if self.excitation_order is None:
+            logger.warning('`CoupledCluster.excitation_order` is undefined.')
+            return False
+        if len(self.excitation_order) > 1:
+            if 2 not in self.excitation_order:
+                logger.error('Coupled Cluster typically starts from doubles.')
```
JosePizarro3, line 537: It looks to me that this could be a `warning` instead?

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+            elif order in self.perturbative_order:
+                name += f'({abbrev})'
+        # cover explicit correlation
+        if self.explicit_correlation is not None:
+            name += self.explicit_correlation
+        # cover specific solver approaches
+        if self.solver in self.map_solver:
+            name = self.solver_map[self.solver] + name
+
+    def type_to_cc(self) -> None:
```
JosePizarro3, line 568: This kind of reaffirms my idea that `type` should be a str and here you do the checks.

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+        for order in (3, 4):
+            if order in self.excitation_order and order in self.perturbative_order:
+                logger.error(
+                    f'Order {order} is defined as both excitation and perturbative.'
+                )
+                return False
+        return True
+
+    def cc_to_type(self) -> None:
+        """Produce an educated guess based on the other parameters."""
```
JosePizarro3, line 548: Can you improve the description?

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+                return False
+        for order in (3, 4):
+            if order in self.excitation_order and order in self.perturbative_order:
+                logger.error(
+                    f'Order {order} is defined as both excitation and perturbative.'
+                )
+                return False
+        return True
+
+    def cc_to_type(self) -> None:
```
JosePizarro3, line 547: You forgot to `return name` and store it in `self.type` in line 599?

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+                    if not ptb_initialized:
+                        self.perturbative_order = []
+                        ptb_initialized = True
+                    self.perturbative_order = np.append(self.perturbative_order, order)
+                else:
+                    if not exc_initialized:
+                        self.excitation_order = []
+                        exc_initialized = True
+                    self.excitation_order = np.append(self.excitation_order, order)
+        if match.group(1) in self.solver_map:
```
JosePizarro3, line 590: You could simply say:
```python
self.solver = self.solver_map.get(match.group(1))
```
Probably a try should be added, in case `match` does not have `group(1)`.

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+            self.solver = self.solver_map[match.group(1)]
+        if match.group(6):
```
JosePizarro3, line 592: Is this a list? If so, I think you should also include a check on the length.

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+            self.explicit_correlation = match.group(6)[1:]  # remove the dash
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
+        if self.type is None:
+            if self.check_orders(logger):
+                self.cc_to_type()
```
JosePizarro3, line 599: I like to return type in this methods, just to clarify when we are setting `self.type` using them, but also it helps when testing.

tests/test_model_method.py
```diff
@@ -0,0 +1,144 @@
...
+
+from nomad.utils import get_logger
+from nomad_simulations.model_method import CoupledCluster
+
+
+def approx(value, abs=0, rel=1e-6):
+    return pytest.approx(value, abs=abs, rel=rel)
+
+
+LOGGER = get_logger(__name__)
```
JosePizarro3, line 29: you can now:
```python
from . import logger
```

tests/test_model_method.py
```diff
@@ -0,0 +1,144 @@
...
+                'ptb_order': None,
+                'slv': 'variational',
+                'corr': 'R12',
+            },
+        ),
+    ],
+)
+def test_cc(name, reference):
+    method = CoupledCluster(type=name.upper())
+    method.normalize(None, LOGGER)
```
JosePizarro3, line 138: actually, I think here we should use `EntryArchive()` as an input, otherwise, super normalizations might break (e.g., for `Entity`)

tests/test_model_method.py
```diff
@@ -0,0 +1,144 @@
...
+            {
+                'exc_order': [2],
+                'ptb_order': None,
+                'slv': 'variational',
+                'corr': 'R12',
+            },
+        ),
+    ],
+)
+def test_cc(name, reference):
```
JosePizarro3, line 136: Add typing.

Also, you don't need to define `reference` as a dictionary, you can simply pass each quantity individually. This would help understanding the testing much more.

Also, could you be more verbose: `cc_type`, `excitation_order`, `perturbation_order`, `solver`, `explicit_correlation`? Also for the name of the testing: `test_coupled_cluster()`

tests/test_model_method.py
```diff
@@ -0,0 +1,144 @@
...
+    method = CoupledCluster(type=name.upper())
```
JosePizarro3, line 137: Can you pass `name` directly capitalized?

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+        Orders at which the excitation are used.
+        1 = single, 2 = double, 3 = triple, 4 = quadruple, etc.
+
+        Note that coupled cluster typically start from doubles.
+        Singles excitations in a Koopman-compliant scheme only make sense as a response to a perturbation.
+        """,
+        a_eln=ELNAnnotation(component='NumberEditQuantity'),
+    )
+
+    perturbative_order = Quantity(
```
JosePizarro3, line 477: `perturbation_order` I think it should be better.

src/nomad_simulations/model_method.py
```diff
@@ -392,6 +402,209 @@ def normalize(self, archive, logger) -> None:
...
+        type=np.int32,
+        shape=['*'],
+        description="""
+        Excitation order at which the perturbative correction is used.
+        1 = single, 2 = double, 3 = triple, 4 = quadruple, etc.
+        """,
+        a_eln=ELNAnnotation(component='NumberEditQuantity'),
+    )
+
+    perturbative_method = SubSection(sub_section=PerturbationMethod.m_def)
```
JosePizarro3, line 487: You are also mixing here "Perturbative" and "Perturbation". Here I'll use also `Perturbation`.

tests/test_model_method.py
```diff
@@ -0,0 +1,144 @@
...
+    method.normalize(None, LOGGER)
+    for ref, res in {'exc': 'excitation', 'ptb': 'perturbative'}.items():
+        res_order = getattr(method, f'{res}_order')
+        res_order = list(res_order) if res_order is not None else None
+        assert res_order == reference[f'{ref}_order']
```
JosePizarro3, line 142: I would actually go testing individual methods instead of the full `normalize` function.

JosePizarro3 (commented):
Good, but I think there are some style changes you were diverging a bit too much. Also, I doubt some MEnums are working as you intend (also, setting a MENum to an empty string is very strange), so please check.

I also like the idea of having a string `CC.type` or the explicit parameters and some methods handling these situations, but I think these should be improved a bit more. Also, I'd suggest to return the quantities and then set them in the `normalize()` function; once you suggested that and I totally agreed with your idea.

Finally, we need better descriptions overall. I think this is a good moment where I can be pushy a bit with the descriptions, especially because I am not an expert and do not fully understand your design decisions. Do you have some nice review that I could take a look to have an idea? Is your thesis a good starting point? 😄

src/nomad_simulations/model_method.py
EBB2675: why do we need this?

src/nomad_simulations/model_method.py
EBB2675: it looks like only ground-state & canonical coupled cluster

ndaelman-hu:
C4 will implement post-HF. This can be closed.

PR #43, opened 2024-04-05
`PhysicalProperty` definition

### LINKED ISSUE #40: `PhysicalProperty` definition

JosePizarro3:
@JFRudzinski I was a bit inspired today, and I managed to make this work. I will anyways keep working for next week. **You don't need to review it today**

Some remarks (more as a compendium of things to not forget):
- I don't know how to set up `Variables` as a `DataType`. This is because we want to also `m_set` the n_bins, bins, bins_errors..., so I think we just need to be `MSection` (as this already has the desired functionality) but with the `normalize` capabilities, i.e., using `ArchiveSection`.
- Instead, `Variables` appear as sub-sections within each PhysicalProperty. `name` sets the specific name of the Variable, and people can easily add custom stuff by simply using `Variables` (check the end of the `outputs.py` module).
- Together,  `variables[i].n_bins` and `shape` set the shape of the `value` of a PhysicalProperty. This is done in `__setattr__`, but I need to check and do testing still.
- `entity_ref` covers the cases of having several contributions to the same physical property, but I want to play a bit more with it.
- The basic `normalize` from `PhysicalProperty` checks if a property is derived or not from another property.
- `Outputs` contains then a list of properties + the `model_system_ref` configuration from which they were obtained. Descriptions need refinement.
- We keep the concept of `SCFOutputs`.

I also created a simplified diagram to understand the relations:
Base Sections for Area C

Closes #40


src/nomad_simulations/outputs.py
```diff
@@ -17,89 +17,214 @@
...
+    # bins_error = Quantity()
 
-    # TODO add time quantities
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
 
-    normalizer_level = 2
 
-    name = Quantity(
+class Temperatures(Variables):
```
JFRudzinski: Just a small thing: I think that the convention should be that variables (at least 1-D variables should be named as a singular, i.e., Temperature, Energy) since they correspond to a singular variable of the function (property). At least that is my intuition but we could discuss of course.

src/nomad_simulations/outputs.py
```diff
@@ -17,89 +17,214 @@
...
+class Energies(Variables):
+    def __init__(self, m_def: Section = None, m_context: Context = None, **kwargs):
+        super().__init__(m_def, m_context, **kwargs)
+        self.name = self.m_def.name
+
+
+class PhysicalProperty(ArchiveSection):
+    """ """
+
+    source = Quantity(
```
JFRudzinski: This is nice! :raised_hands: I was just talking to Julia about something that would need this functionality in her case

src/nomad_simulations/outputs.py
```diff
@@ -17,89 +17,214 @@
...
 
-    orbitals_state_ref = Quantity(
-        type=OrbitalsState,
+    label = Quantity(
+        type=str,
         description="""
-        Reference to the `OrbitalsState` section to which the output property references to and on
-        on which the simulation is performed.
+        Label for additional classification of the physical property. Example: an `ElectronicBandGap`
+        can be labeled as `'DFT'` or `'GW'` depending on the methodology used to calculate it.
```
JFRudzinski: I imagine that the usage of label is very flexible, we should add a bunch of examples in the description (and docs of course) eventually to help guide people. I would use this to specify the types of atoms, molecules, etc. that this property is describing. E.g., the center of mass molecular mean squared displacement for waters maybe would have type = 'Dynamic' or 'Correlation Function', and then label = 'H20'. Is this consistent with what you are thinking?

src/nomad_simulations/outputs.py
```diff
@@ -17,89 +17,214 @@
...
         """,
-        a_eln=ELNAnnotation(component='ReferenceEditQuantity'),
     )
 
-    atoms_state_ref = Quantity(
-        type=AtomsState,
+    shape = DirectQuantity(
```
JFRudzinski: Did this end up being a correct use of DirectQuantity? I was just copying Quantity when I wrote this in my class but I never tested it

src/nomad_simulations/outputs.py
```diff
@@ -17,89 +17,214 @@
...
     )
 
-    model_system_ref = Quantity(
-        type=ModelSystem,
+    variables = SubSection(
+        type=Variables.m_def,
         description="""
-        Reference to the `ModelSystem` section to which the output property references to and on
-        on which the simulation is performed.
+        Variables over which the physical property varies. These are defined as binned, i.e., discretized
```
JFRudzinski: what happens if the function is a scalar in this variable? I don't have a very clear example besides the Cartesian coordinates, which we discussed is a special case that is not covered by this. But I am pretty sure there will be other examples. It is sort of like cases where the variable has a single bin, but that bin cannot be described numerically...it is still unclear to me. I mean, there is definitely a case with bins that are not floats but integers or booleans, like you have a setting which is simply on or off.

When a property is a scalar all together it means that you've effectively integrated over all variables, ... anyway didn't mean to ramble, I need to think about this more :thinking:

src/nomad_simulations/outputs.py
```diff
@@ -160,3 +345,21 @@ class SCFOutputs(Outputs):
...
         super().normalize(archive, logger)
+
+
+# Playing with `PhysicalProperty`
+band_gap = ElectronicBandGap(source='simulation', type='direct', label='DFT')
+n_bins = 3
+temperature = Temperatures(n_bins=n_bins, bins=np.linspace(0, 100, n_bins))
+band_gap.variables.append(temperature)
+n_bins = 6
+custom_bins = Variables(n_bins=n_bins, bins=np.linspace(0, 100, n_bins))
```
JFRudzinski: what does this correspond to? Or is it just a dummy test? You can pass a name, right?

src/nomad_simulations/outputs.py
```diff
@@ -160,3 +345,21 @@ class SCFOutputs(Outputs):
...
+band_gap.variables.append(custom_bins)
+# band_gap.value_unit = 'joule'
+band_gap.value = [
+    [[1, 2, 3], [1, 2, 3]],
+    [[1, 2, 3], [1, 2, 3]],
+    [[1, 2, 3], [1, 2, 3]],
+] * ureg.eV
+# band_gap.value = [1, 2, 3] * ureg.eV
+print(band_gap)
```
JFRudzinski: :raised_hands:

JFRudzinski (approved):
Just a quick/soft review to document some initial thoughts. I really like the direction this is going, thanks a lot for all your effort on it!

My plan is to try to look more carefully Mon morning (although I have 1 meeting unfortunately). I mean in principle I would like to try it out and implement my own examples before we meet, but I am not sure that it is necessary. I think it will take quite a while to even discuss the overview, motivations, i.e., bigger picture stuff. So, we might not even get that deep into the implementation on Mon. 

In any case, think a bit if you feel ready to discuss with the others, and then we decide Mon morning if we keep the meeting or push it 1-2 days.

src/nomad_simulations/outputs.py
JosePizarro3: Good thing is that, if `variables` and shape does not coincide with the stored `value`, then a `ValueError` is raised 🙂

src/nomad_simulations/outputs.py
JosePizarro3: Dummy test, but indeed, one can pass a `name` to make clear which variables these are.

src/nomad_simulations/outputs.py
JosePizarro3: Yeah, sounds good. I am thinking on plural because these are an array of temperatures, but it is true that this would be encoded in the plural of `bin`.

src/nomad_simulations/outputs.py
JosePizarro3: Yeah, I think labels are some "additional" classification, but there is not real rule for this, I'd say. We can put some examples like you said, so users can use this properly into their situations.

src/nomad_simulations/outputs.py
JosePizarro3: Yes, otherwise `Dimension` is not properly working 🤷🏻

src/nomad_simulations/outputs.py
JosePizarro3: Ok, I see. Quickly answering to the fact that bins can be anything, then I'd suggest go with `Any` as type, meaning, it can be any booleans, integers, floats. I am not 100% sure this is the correct typing, but let me ask.

JosePizarro3:
Ok, I added a few minor fixes and normalization for `n_bins`. I am going to work on other things before our meeting, but we can go back to this afterwards and merge today.

JosePizarro3:
Ok I added some changes after our meeting + testing for variables.py and physical_property.py.

Tomorrow morning I will add some final testing for outputs.py and this will be ready for us to quickly chat about it, and merge.

JosePizarro3:
Hi there everybody,

Would you mind dropping a review? I am still working on testing for outputs.py, but you can already see the idea of the schema.

src/nomad_simulations/variables.py
```diff
@@ -0,0 +1,126 @@
...
+        self.n_grid_points = self.get_n_grid_points(self.grid_points, logger)
+
+
+class Temperature(Variables):
+    """ """
+
+    grid_points = Quantity(
+        type=np.float64,
+        unit='kelvin',
+        shape=['n_bins'],
```
JFRudzinski: n_grid_points

src/nomad_simulations/variables.py
```diff
@@ -0,0 +1,126 @@
...
+        super().normalize(archive, logger)
+
+
+class Energy(Variables):
+    """ """
+
+    grid_points = Quantity(
+        type=np.float64,
+        unit='joule',
+        shape=['n_bins'],
```
JFRudzinski: n_grid_points

JFRudzinski (approved):
Looks good to me :+1:

src/nomad_simulations/outputs.py
JosePizarro3: I added a couple of warning comments (`# !`) for adding more examples in `label` and `type` in a future.

JosePizarro3:
Ok, I improved some functionalities to extract if a physical property `is_scf_converged` or not, and added tests.

If nobody has any objection, by tomorrow I will merge this.

src/nomad_simulations/outputs.py
JosePizarro3: I added a couple of warning comments (`# !`) for adding more examples in `label` and `type` in a future.

JosePizarro3:
@JFRudzinski one final thing: after talking with Andrea I had some feedback and it makes sense to change the `PhysicalProperty.shape` naming to something more related with the rank of the tensor, i.e., we can call it `rank` or `tensor_rank`. What do you think?

JosePizarro3:
Ok, after our meeting, I will be merging this tonight

@JFRudzinski do you mind doing a final pass? 
@ndaelman-hu can you please review this? I will consider your comments in the other pr, but please, bear in mind you were reviewing some old stuff so the comments won't apply anymore :/

JosePizarro3:
@lauri-codes you can also review this

JFRudzinski:
> @JFRudzinski one final thing: after talking with Andrea I had some feedback and it makes sense to change the `PhysicalProperty.shape` naming to something more related with the rank of the tensor, i.e., we can call it `rank` or `tensor_rank`. What do you think?

I think rank is ok, I don't really like tensor_rank, it's a bit much.

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+            - scalars (tensor rank 0) have `rank=[]` (`len(rank) = 0`),
+            - vectors (tensor rank 1) have `rank=[a]` (`len(rank) = 1`),
+            - matrices (tensor rank 2), have `rank=[a, b]` (`len(rank) = 2`),
+            - etc.
+        """,
+    )
+
+    variables = SubSection(sub_section=Variables.m_def, repeats=True)
+
+    value: Quantity = _placeholder_quantity
```
JFRudzinski, line 103: was just wondering how this works for the custom property? I guess it doesn't matter if value can be any datatype. I was thinking at first that it was always a float, but now I'm not sure :thinking:

JFRudzinski (approved):
It looks good to me! I don't think I'll have time to test it out today, but I already tested the overall functionality before and it seemed good. I think you should go ahead and merge it when you are ready. Thanks!

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+# We add `logger` for the `PhysicalProperty.variables_shape` method
+logger = utils.get_logger(__name__)
+
+
+class PhysicalProperty(ArchiveSection):
+    """
+    A base section used to define the physical properties obtained in a simulation, experiment, or in a post-processing
+    analysis. Besides its `value`, it also contains:
```
ndaelman-hu: I would add a bit more explanation about the intent here, else that'll be lost:
if I understood you well, you want something `Quantity`-like that can "contain" other `Quantity`-like attributes in a number of prespecified ways.

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+        Name of the physical property. Example: `'ElectronicBandGap'`.
+        """,
+    )
+
+    source = Quantity(
+        type=MEnum('simulation', 'measurement', 'analysis'),
+        default='simulation',
+        description="""
+        Source of the physical property. Example: an `ElectronicBandGap` can be obtained from a `'simulation'`
+        or in an `'measurement'`.
```
ndaelman-hu: typo: "an"

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+
+    name = Quantity(
+        type=str,
+        description="""
+        Name of the physical property. Example: `'ElectronicBandGap'`.
+        """,
+    )
+
+    source = Quantity(
+        type=MEnum('simulation', 'measurement', 'analysis'),
```
ndaelman-hu, line 60: As mentioned before, I wouldn't use enumeration here, since it's too restrictive and then has to be amended via inheritance.
2 example categories that should be added: synthesis and machine-learned models.

Rather, think about this in terms of what could be a source:
- plugins?
- other archives, or entries, etc.?

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+    )
+
+    type = Quantity(
+        type=str,
+        description="""
+        Type categorization of the physical property. Example: an `ElectronicBandGap` can be `'direct'`
+        or `'indirect'`.
+        """,
+        # ! add more examples in the description
+    )
```
ndaelman-hu: This definition does not explain anything. As denoted in your comment, I think you're picking up on this.
Rather, try and specify when something would classify as a `type`. If you can't, then remove this.

Personally, I think `type` is just lazy naming: in `ElectronicBandGap` it means the `momentum_vector`, but since people sometimes use the term `type` (that is, just distinguishing non-zero cases) it's fine there.

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+
+    label = Quantity(
+        type=str,
+        description="""
+        Label for additional classification of the physical property. Example: an `ElectronicBandGap`
+        can be labeled as `'DFT'` or `'GW'` depending on the methodology used to calculate it.
+        """,
+        # ! add more examples in the description
+    )
```
ndaelman-hu: Same remark as with `type`. If a property needs a `label`, let it introduce it and name it itself.

Let me play devil's advocate and ask you: _why stop with `type` and `label`? There are also `kind`, `category`, `classification`, etc._
Used as now, these are all just vague indicators, not actual abstractions...

src/nomad_simulations/physical_property.py
JosePizarro3: Yeah, I think in any case `value` has to be set up in the custom property you want to define.

ndaelman-hu: Shame that you can't add a description here.
Though, `property()` upon which `_placeholder_quantity` is based does allow for `doc`

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+
+    rank = DirectQuantity(
```
ndaelman-hu, line 86: Is the use of `DirectQuantity` instead of `Quantity` necessary here? The docs specify that it resolves _bootstrapping_.
Or is this just a hack to circumvent the typing system of `Quantity`?

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+    variables = SubSection(sub_section=Variables.m_def, repeats=True)
+
+    value: Quantity = _placeholder_quantity
+
+    entity_ref = Quantity(
+        type=Entity,
+        description="""
+        Reference to the entity that the physical property refers to.
+        """,
+    )
```
ndaelman-hu: This is again quite vague, but here I think you might have a shot at concretizing.
Within this plugin, you use `Entity` for defining a `GeometricSpace`. Are there any use cases that you're expecting?
I thought that this was rather meant to refer to the proper part in `model_system`.

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+
+    physical_property_ref = Quantity(
+        type=Reference(SectionProxy('PhysicalProperty')),
+        description="""
+        Reference to the `PhysicalProperty` section from which the physical property was derived. If `physical_property_ref`
+        is populated, the quantity `is_derived` is set to True via normalization.
+        """,
+    )
```
ndaelman-hu, line 121: So, in the end, you decided never to nest other `PhysicalProperty`s?

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+        description="""
+        Flag indicating whether the physical property is derived from other physical properties. We make
+        the distinction between directly parsed and derived physical properties:
+            - Directly parsed: the physical property is directly parsed from the simulation output files.
+            - Derived: the physical property is derived from other physical properties. No extra numerical settings
+                are required to calculate the physical property.
+        """,
+    )
+
+    is_scf_converged = Quantity(
```
ndaelman-hu, line 135: Where is the actual SCF progression stored? Wasn't this supposed to be nested?

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+        type=bool,
+        description="""
+        Flag indicating whether the physical property is converged or not after a SCF process. This quantity is connected
+        with `SelfConsistency` defined in the `numerical_settings.py` module.
```
ndaelman-hu, line 139: I would mention `numerical_settings.py` in the description of `self_consistency_ref`

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+
+    self_consistency_ref = Quantity(
+        type=SelfConsistency,
+        description="""
+        Reference to the `SelfConsistency` section that defines the numerical settings to converge the
+        physical property.
+        """,
+    )
+
+    @property
```
ndaelman-hu, line 151: One important note about the `property` decorator: indexing them is a nightmare.
Only use this if you don't want your quantity to be searchable.

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+        description="""
+        Rank of the tensor describing the physical property. This quantity is stored as a Dimension:
+            - scalars (tensor rank 0) have `rank=[]` (`len(rank) = 0`),
+            - vectors (tensor rank 1) have `rank=[a]` (`len(rank) = 1`),
+            - matrices (tensor rank 2), have `rank=[a, b]` (`len(rank) = 2`),
+            - etc.
+        """,
+    )
+
+    variables = SubSection(sub_section=Variables.m_def, repeats=True)
```
ndaelman-hu, line 100: `dependent_variables`

ndaelman-hu: Btw, why can't these be other properties in their own right?

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+        """
+        if self.variables is not None:
+            return [v.get_n_grid_points(v.grid_points, logger) for v in self.variables]
+        return []
+
+    @property
+    def full_shape(self) -> list:
+        """
+        Full shape of the physical property. This quantity is calculated as:
+            `full_shape = variables_shape + shape`
```
ndaelman-hu: Be careful, cause this sounds as if you are talking about numbers.
The proper term for lists is _concatenation_. Python calls it `__add__()` so they can use the built-in symbol.

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+        where `shape` is passed as an attribute of the `PhysicalProperty` and is related with the order of
+        the tensor of `value`, and `variables_shape` is obtained from `variables_shape` and is
+        related with the shapes of the `variables` over which the physical property varies.
+
+        Example: a physical property which is a 3D vector and varies with `variables=[Temperature, ElectricField]`
+        will have `shape = [3]`, `variables_shape=[n_temperatures, n_electric_fields]`, and thus
+        `full_shape=[n_temperatures, n_electric_fields, 3]`.
```
ndaelman-hu: Now _property_ will be responsible for containing its own definition along with dependent variables.
This leads to the awkward case where extracting `Energy` data, I have to plow through all its variables...
So, every time a data scientist wants to manipulate any `PhysicalPropertyBaseclass`, they first have to access `full_shape` and `variable_shape` to extract the underlying property itself...

How about `PhysicalPropertyBaseclass` actually stores `value` and `dependent_variables` separately, and if the user wants the manifold, they access both?

ndaelman-hu (commented):
Here's a first review, more focused on the concepts. Haven't gotten around to everything yet.
To dig into the technical side, I'd definitely need more time. Or you get somebody from Area D to sign off on that part.

ndaelman-hu:
> * I don't know how to set up `Variables` as a `DataType`. This is because we want to also `m_set` the n_bins, bins, bins_errors..., so I think we just need to be `MSection` (as this already has the desired functionality) but with the `normalize` capabilities, i.e., using `ArchiveSection`.

Let's see to look into this, on Friday maybe?

src/nomad_simulations/physical_property.py
JosePizarro3: After our meeting with the interoperability TF, I will leave this to talk with Lauri, Markus and Theodore next week.

My problem here was that, when instantiating `value` as a Quantity with type, description, etc, then I cannot assign shape in a dynamic way (and similarly, unit will not be possible to be done like this), and rather it gets assign when initiating for the first time (even if it is a None, I cannot assign it a posteriori). I think this could be fixed if these attrs could be assigned `Any` for any potential value, but I will hear from Area D first about this.

So if you all agree, I keep this as it is now, and next week we will take care of it.

src/nomad_simulations/outputs.py
```diff
@@ -17,18 +17,48 @@
...
+    # TODO add more functionalities here
+
+    def __init__(self, m_def: Section = None, m_context: Context = None, **kwargs):
+        super().__init__(m_def, m_context, **kwargs)
+        self.name = self.m_def.name
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
+
+
```
lauri-codes, line 61: I guess this is a good example of what defining a new property looks like when you use `PhysicalProperty` as a base class. There are things here that need some thinking:

- We don't currently have a concept for what you are doing here with `rank`: section definitions only consist of fields for quantities and subsections, whereas `rank` here is some kind of a default value for all instances of this section. This may cause all kinds of problems in the current code and in the YAML schemas.
- Are the functions needed here? Remember also that if someone uses `PhysicalProperty` as a base section in YAML schemas, they cannot define functions.
- `value` is also kind of an interesting case: you need to re-refine it for each subclass (because the units or data type change) and should get an error if you don't define it, as it is part of the interface that all PhysicalProperties should follow.

We should discuss this in the meeting next Wednesday. You could present this code as an example of what defining a new `PhysicalProperty` looks like as this is ultimately what users would need to use a lot in custom schemas.

src/nomad_simulations/physical_property.py
```diff
@@ -0,0 +1,231 @@
...
+
+from .variables import Variables
+from .numerical_settings import SelfConsistency
+
+
+# We add `logger` for the `PhysicalProperty.variables_shape` method
+logger = utils.get_logger(__name__)
+
+
+class PhysicalProperty(ArchiveSection):
```
lauri-codes, line 43: As of now, `PhysicalProperty` has many Area C specific fields that don't need to be a part of the underlying base class. I would say that if we want to add such a base class to the core package, it should only contain `value`, `rank` and `variables` in the first version. This will simplify the discussions a lot, but also does not stop you from extending the base class to fit your needs.

src/nomad_simulations/outputs.py
ndaelman-hu: > value is also kind of an interesting case: you need to re-refine it for each subclass (because the units or data type change) and should get an error if you don't define it, as it is part of the interface that all PhysicalProperties should follow.

very good point! hadn't considered this potential pitfall yet

src/nomad_simulations/physical_property.py
ndaelman-hu: Sure, I indeed think it's advisable to consult Area D here.

Whether this should already be merged depends on any deadlines. From my side, at least, don't feel pressured.
This approach is foundational enough, that it warrants a proper implementation: we don't want a plugin built on "good-enough-for-now" code.

src/nomad_simulations/physical_property.py
JosePizarro3: Well, all these strings are simple quantities for easy identification of what is going on. This one, `source`, pretty much relate with the "type of Activity" being performed. I am not sure `'synthesis'` deserves its own label (I meant, everything will fall into `'measurement'`), and we can talk with Sarthak and others if `'ml_model'` deserves its own source label or it can fall under the category of `'analysis'`.

As you see, I am talking here about the **activity** responsible of the property. I will change the description to make it more clear 👍🏻

src/nomad_simulations/physical_property.py
JosePizarro3: I will answer up, in the comment on type.

src/nomad_simulations/physical_property.py
JosePizarro3: Yes, I agree, but the problem is that we should try to standarize a bit the strings that serve as identifiers / classifiers in the metainfo (as @JFRudzinski pointed out, there is a lot of inconsistent naming). Here, our goal should be "let's try to define the minimal number of them which are useful". 

Now, I'd argue that `type` is useful because it is easier to look if a band gap has some string saying "direct" or "indirect", rather than looking at the momentum. But there might be other situations in which someone wants to redefine it a bit for convenience. Hence the comment on "add more examples".

Similarly, for me, `label` would be very useful for identifying `method_label`, i.e.,  if a property has been derived from one or another methodology. Again, you can argue "but you can look into the references and you will find the same information", and yes, but a simple string does not hurt. Instead of defining `method_label`, I think I could be using `label`, as well as other people could repurpose it in their specific properties. And because this is pretty much case-dependent, I see little to no overlapping on conflicting label naming.

> Let me play devil's advocate and ask you: why stop with type and label? There are also kind, category, classification, etc.

I answered a bit above, but for me `kind` and `type` serve the same purpose. Similarly, `category` would help us further classify, but I will keep it out for now (we might not even want to define it, but perhaps there is some interesting connection with the FAIRmat taxonomy). I am not sure what you meant with `classification`, but I am guessing that it could serve the same purpose than `category`?

_____

Now, if in the future we find use-cases where none of this holds, then I would suggest to revisit it and come with a better idea of how to handle easy string identifiers. **But, we need more examples**.

src/nomad_simulations/physical_property.py
JosePizarro3: When I was using `Quantity`, I could not manipulate lists to concatenate them so that `full_shape` is a Dimension as well and can be passed to the `value.shape`. I can ask Theodore about this.

JosePizarro3: So, for your third question: yes

src/nomad_simulations/physical_property.py
JosePizarro3: I can put examples in the description, thanks for pointing the vagueness of this.

For you to know, this can ref `AtomsState`, `OrbitalsState`, in order to differentiate `values_per_atom` or `values_per_orbital`. I think we can leave it as abstract as ref to an `Entity` so that it works in, effectively, any case. The ref to `ModelSystem` holds a special situation because of the time evolution (we need a lists of Outputs with refs to the elements in the list of ModelSystems).

src/nomad_simulations/physical_property.py
JosePizarro3: We will, for when we have components.

This is just a way of defining `is_derived` properties, e.g., an ElectronicBandGap which is derived from an ElectronicDOS.

src/nomad_simulations/physical_property.py
JosePizarro3: For outputs, you will need to use `SCFOutputs`. A section could look like:
```sh
.
└── data(Simulation)/
    └── outputs(SCFOutputs)/
        ├── 0/
        │   ├── scf_steps(Outputs)/
        │   │   ├── 0/
        │   │   │   └── total_energy(TotalEnergy)/
        │   │   │       └── value
        │   │   ├── 1/
        │   │   │   └── total_energy(TotalEnergy)/
        │   │   │       └── value
        │   │   └── 2/
        │   │       └── total_energy(TotalEnergy)/
        │   │           └── value
        │   └── total_energy(TotalEnergy)/
        │       ├── value
        │       ├── self_consistency_ref
        │       └── is_scf_converged
        ├── 1/
        │   └── ...
        └── 2/
            └── ...
```

src/nomad_simulations/physical_property.py
JosePizarro3: Yes, these are internal properties used to calculate the `full_shape` of value, so we don't want them searchable.

src/nomad_simulations/physical_property.py
JosePizarro3: Cool, thanks a lot, I could never remember _concatenation_, I was always using the concept of _adding_ two lists 😄  I will use it in the description, as I think it is easier to understand.

src/nomad_simulations/physical_property.py
JosePizarro3: I prefer `variables`. I will answer to your 2nd question in your comment at the end.

src/nomad_simulations/physical_property.py
JosePizarro3: You also commented:
> Btw, why can't these be other properties in their own right?

I think it is hard for me to understand specific cases where this might happen. I am guessing you are thinking about physical properties that are **functionals**, but for those, I think anyways we should see some example before taking a decision.

In any other case, I'll argue that the concept of `Energy(Variables)` is slightly different from `Energy(PhysicalProperty)`: the first is just the discretization of your space for measuring a property, the later is the property you are measuring directly. I am pretty sure Area A will disagree with this, but I will argue that in any case, when someone is doing a measurement of, e.g., an spectra, you **vary** a certain parameter while **measuring** what is happening to the other.

___

Do you have in any case some specific example where this does not happen like I am arguing?

src/nomad_simulations/physical_property.py
JosePizarro3: > has many Area C specific fields

I very much disagree with this. Would you mind elaborating? Maybe the only one being specific is the `self_consistency_ref` and `is_scf_converged` quantities. Anything else is pretty general.

src/nomad_simulations/outputs.py
JosePizarro3: > - We don't currently have a concept for what you are doing here with rank: section definitions only consist of fields for quantities and subsections, whereas rank here is some kind of a default value for all instances of this section. This may cause all kinds of problems in the current code and in the YAML schemas.

Not sure if I understand this part. `rank` **must** be defined by default for any physical property defined in the schema. This is just the rank of the tensor that we are refering to, so I don't know why any YAML schema should touch on this. Or do you mean that, defining by default like in here is not possible in a YAML?

> - Are the functions needed here? Remember also that if someone uses PhysicalProperty as a base section in YAML schemas, they cannot define functions.

Yes! This is so important, we also want to normalize the physical properties to maintain interoperability. This is a drawback of using YAML instead of Python schemas imo, not of what we defined here.

> - value is also kind of an interesting case: you need to re-refine it for each subclass (because the units or data type change) and should get an error if you don't define it, as it is part of the interface that all PhysicalProperties should follow.

Yes, this would be a bit better if `value` could take Any generic type, shape, unit at the `PhysicalProperty` level as well. But yes, once someone defines a new inherited PhysicalProperty, they need to overwrite `value` with the specific type and units (and shape via `rank`).


_____

> We should discuss this in the meeting next Wednesday. You could present this code as an example of what defining a new PhysicalProperty looks like as this is ultimately what users would need to use a lot in custom schemas.

Sure, there are a bunch of issues in this repo I want to tackle, to fully show the capabilities of `PhysicalProperty`, and how to use it in a complex case. I will try to have it for next week.

JosePizarro3:
Hi there,

So in order to advance a bit on specific examples, let me merge this. I will move @lauri-codes and @ndaelman-hu comments into a Discussion, so we can keep talking about design ideas there.

However, we need some base structure that we can work on with some specific examples to check if the ideas here make sense or not.

Thanks for the reviews and comments 🙂

PR #44, opened 2024-04-09
Fix setup to respect other plugins

ndaelman-hu:
Clarify use of `include` and `exclude` in the setup instructions.
Should we maybe just refer to the general docs for the setup?

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,16 +71,17 @@ description: |
   This is a collection of NOMAD schemas for simulation data.
 ```
 
-and define the ```nomad.yaml``` configuration file of your NOMAD instance in the root folder with the following content:
+and register the plugin in the ```nomad.yaml``` configuration file of your NOMAD root folder:
 
 ```yaml
 plugins:
-  include:
-    - schemas/nomad_simulations
   options:
     schemas/nomad_simulations:
       python_package: nomad_simulations
 ```
+
+If you want to administer your plugins, you can use `include` or `exclude` flags.
+ATTENTION: when using `include`, any plugin not specified will not be loaded.
 
 You also need to add the package folder to the `PYTHONPATH` of the Python environment of your local NOMAD installation. This can be done by specifying the relative path to this repository. Either run the following command every time you start a new terminal for running the appworker, or add it to your virtual environment in `<path-to-local-nomad-installation>/.pyenv/bin/activate` file:
 


README.md
```diff
@@ -71,17 +71,18 @@ description: |
   This is a collection of NOMAD schemas for simulation data.
 ```
 
-and define the ```nomad.yaml``` configuration file of your NOMAD instance in the root folder with the following content:
+and register the plugin in the ```nomad.yaml``` configuration file of your NOMAD root folder:
 
 ```yaml
 plugins:
-  include:
```
JosePizarro3, line 78: Please, we can leave this here. It is just about probably adding the comment that you can find in the nomad.yaml if you feel so.

Furthermore, it seems that manually installing nomad-lab[parsing] fixes the built-in plugins.

JosePizarro3 (commented):
I can work on improving this README part on the local installation setup. It is just not the moment now.

Please, leave this as a PR for now, we can work it out next week or in a couple of weeks.

README.md
ndaelman-hu: Well, this is the only correction, so no need to further work on it: I haven't seen any other errors, and have verified that this here works...
You can ask Lauri to take a look.

ndaelman-hu: Just noticed this comment now:

> Please, we can leave this here. It is just about probably adding the comment that you can find in the nomad.yaml if you feel so

No, this setup disables ALL other plugins, which should never be the default! There are only a few scenarios where this is wishful. You only need `options` to add plugins to the list, then you can run all of them.
I do think that this has not been properly explained in the general docs. I'll see to amend that.

ndaelman-hu: Just to double-check, @JFRudzinski or @Bernadette-Mohr, following the old setup, can you still run the regular parsers?
Not the tests, but the actual parsers.

JosePizarro3:
@ndaelman-hu 

I am curious about which exact problems you are having now. We should simply change the text pointing to the diffs between having a local installation and a NOMAD oasis, and I agree this is in the general docu. 

Here we just fix the text.

Did you also check the general documentation page? More specifically, this one: https://nomad-lab.eu/prod/v1/staging/docs/howto/oasis/plugins_install.html#add-a-plugin-to-nomad

ndaelman-hu:
> @ndaelman-hu
> 
> I am curious about which exact problems you are having now. We should simply change the text pointing to the diffs between having a local installation and a NOMAD oasis, and I agree this is in the general docu.
> 
> Here we just fix the text.
> 
> Did you also check the general documentation page? More specifically, this one: https://nomad-lab.eu/prod/v1/staging/docs/howto/oasis/plugins_install.html#add-a-plugin-to-nomad

The issue is that the README instructions give undesired results, i.e. **I can no longer utilize the other plugins, such as parsers**, after applying these steps.
The use of `include` is superfluous here, period. All the rest is fine. Rather than fully removing `include`, I now added an explanation on its proper use, as is given in the link that you posted. Still, an explanation should not be included as a comment in a code snippet.

JosePizarro3:
Given that the setup has changed, and now it works without conflict, I am closing this PR.

PR #46, opened 2024-04-09
45 add basic structure for energy classes

### LINKED ISSUE #45: Add basic structure for energy classes

JFRudzinski:
@ndaelman-hu @JosePizarro3 @Bernadette-Mohr Here is a first proposal for a basic structure for various energies. Please provide feedback at your convenience (you can exclusively look at `energies.py`)


JFRudzinski:
Thinking now about extending this to forces: Should the forces structure not be an exact mirror of energies? If so, how can we encourage or ensure that they are always updated/extended simultaneously?

JosePizarro3:
> Thinking now about extending this to forces: Should the forces structure not be an exact mirror of energies? If so, how can we encourage or ensure that they are always updated/extended simultaneously?

Are they tho? I am guessing that forces are matrices (or vectors?), while energies are scalars, no? 

What do you mean "updated/extended simultaneously"? We can do a quick chatting in a Discord room to understand this point too.

src/nomad_simulations/outputs.py
```diff
@@ -17,16 +17,40 @@
...
+        unit='joule',
+        description="""
+        The value of the electronic band gap.
+        """,
+    )
+
+    # TODO add more functionalities here
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
```
ndaelman-hu: `value` cannot be negative. Pls add a check.

src/nomad_simulations/outputs.py
```diff
@@ -17,16 +17,40 @@
...
+
+    value = Quantity(
+        type=np.float64,
+        unit='joule',
+        description="""
+        The value of the electronic band gap.
+        """,
+    )
+
+    # TODO add more functionalities here
```
ndaelman-hu: Maybe we can specify the momentum vector when `type == indirect`

src/nomad_simulations/outputs.py
```diff
@@ -78,53 +75,21 @@ class Outputs(ArchiveSection):
...
+    # List of properties
+    # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+    electronic_band_gap = SubSection(sub_section=ElectronicBandGap.m_def, repeats=True)
 
     def normalize(self, archive, logger) -> None:
         super().normalize(archive, logger)
 
         # Resolve if the output property `is_derived` or not.
-        self.is_derived = self.resolve_is_derived(self.outputs_ref)
+        # self.is_derived = self.resolve_is_derived(self.outputs_ref)
```
ndaelman-hu: Can this be removed?

src/nomad_simulations/outputs.py
```diff
@@ -78,53 +75,21 @@ class Outputs(ArchiveSection):
...
 
 
 class SCFOutputs(Outputs):
     """
-    This section contains the self-consistent (SCF) steps performed to converge an output property,
-    as well as the information if the output property `is_converged` or not, depending on the
-    settings in the `SelfConsistency` base class defined in `numerical_settings.py`.
+    This section contains the self-consistent (SCF) steps performed to converge an output property.
 
     For simplicity, we contain the SCF steps of a simulation as part of the minimal workflow defined in NOMAD,
```
ndaelman-hu: I would remove "For simplicity"

src/nomad_simulations/outputs.py
```diff
@@ -140,23 +105,52 @@ class SCFOutputs(Outputs):
...
+        for property_name in self.m_def.all_sub_sections.keys():  # ! Check this
+            # We get the property of `SCFOutputs` that we want to check
+            prop = getattr(self, property_name)
+
+            # If it is None or does not contain a `self_consistency_ref` quantity we continue
+            if prop is None:
+                continue
+            scf_parameters = prop.self_consistency_ref
+            if scf_parameters is None:
+                continue
```
ndaelman-hu: Why are these checks needed, if only `self.scf_steps` is used to verify convergence?

src/nomad_simulations/outputs.py
```diff
@@ -140,23 +105,52 @@ class SCFOutputs(Outputs):
...
+                logger.info(
+                    f'The SCF process for the property `{property_name}` did not converge.'
+                )
+                prop.is_scf_converged = False
 
     def normalize(self, archive, logger) -> None:
         super().normalize(archive, logger)
+
+        # Resolve the `is_scf_converged` flag for all SCF obtained properties
+        self.resolve_is_derived(logger)
```
ndaelman-hu: I would only run this after checking that `is_derived` hasn't been set: typically, the parser can extract convergence.

src/nomad_simulations/outputs.py
```diff
@@ -140,23 +105,52 @@ class SCFOutputs(Outputs):
...
+                    break
+                if scf_prop.value.unit != scf_parameters.threshold_unit:
+                    logger.error(
+                        f'The units of the `scf_step.{property_name}.value` does not coincide with the units of the `self_consistency_ref.threshold_unit`.'
+                    )
+                    return
+                scf_props.append(scf_prop.value.magnitude)  # we use only the magnitude
+
+            # We check if the property `is_scf_converged` checking its difference with respect to `self_consistency_ref.threshold_change`
+            threshold = scf_parameters.threshold_change
```
ndaelman-hu: no need to use this assignment

src/nomad_simulations/outputs.py
```diff
@@ -140,23 +105,52 @@ class SCFOutputs(Outputs):
...
+            scf_diff = abs(scf_props[0] - scf_props[1])
```
ndaelman-hu: Technically all fine. Maybe we want to construct a generic method later on? So, add a TODO?

ndaelman-hu (commented):
Reviewed `outputs.py`, can do the rest if needed.
No real conceptual comments, just specifics.

JosePizarro3:
> Reviewed `outputs.py`, can do the rest if needed. No real conceptual comments, just specifics.

You are reviewing the wrong changes. The ones in #43 are the ones you want to take a look.

JFRudzinski:
> > Thinking now about extending this to forces: Should the forces structure not be an exact mirror of energies? If so, how can we encourage or ensure that they are always updated/extended simultaneously?
> 
> Are they tho? I am guessing that forces are matrices (or vectors?), while energies are scalars, no?
> 
> What do you mean "updated/extended simultaneously"? We can do a quick chatting in a Discord room to understand this point too.

Sorry for being unclear. I just meant that besides the total force, there are various force contributions defined currently, e.g., "free" and "t0". These directly correspond to some energy contributions (although the schema is not nearly as built out as for energies. In principle, there could be a force contribution/definition for every energy, since they are directly related.

My question was, why not enforce there to always be both terms added, for each type of contribution? And if this is desired how can we try to achieve this in practice?

For the classical contributions I will build this mirroring regardless, since energy and force contributions coming from particular interactions are commonly stored.

ndaelman-hu:
> > Reviewed `outputs.py`, can do the rest if needed. No real conceptual comments, just specifics.
> 
> You are reviewing the wrong changes. The ones in #43 are the ones you want to take a look.

@JFRudzinski if you branch off from a non-merged branch, pls specify this in the PR or set the target branch to the proper base...

JFRudzinski:
> > > Reviewed `outputs.py`, can do the rest if needed. No real conceptual comments, just specifics.
> > 
> > 
> > You are reviewing the wrong changes. The ones in #43 are the ones you want to take a look.
> 
> @JFRudzinski if you branch off from a non-merged branch, pls specify this in the PR or set the target branch to the proper base...

Ah! Got it, Sorry about that!

ndaelman-hu:
> My question was, why not enforce there to always be both terms added, for each type of contribution? And if this is desired how can we try to achieve this in practice?

I don't think that we should enforce this across the board for 2 reasons:

1. to derive a (general) force, you need to know the energy manifold in terms of its dependent variables. It's not a given that this information will always be present.
2. I'm not sure whether it's always desirable for the (general) "forces" to be known. E.g. in the case of total DFT energy, I have little interest in the chemical potential. This is only relevant from a convergence perspective.

With that said, I think it's a good idea to provide an optional functionality. Would recommend something numpy-based as implementation.

JFRudzinski:
> ide an optional functionality. Would recommend something numpy-based as implementation.

I don't really understand what you are saying tbh. I am not talking about enforcing the storage of such quantities. I was talking about schema development. I think "enforcing" is a bit too strong of a word, but I had in mind the case that someone wants to store some energy contribution and implements a class for it, but doesn't create the corresponding force contribution because they personally do not store this in there simulations. But then another user comes along and wants to store this force contribution and now has to extend the schema themselves, whereas if the original person had been encouraged to add a corresponding force contribution it would have made everyone's life a lot easier.

Of course there might be some energy definitions that it doesn't make sense to define a corresponding force. In that case, obviously it should not be done, but I think in most cases it is possible at least in principle.

JosePizarro3:
To be honest, I need to see or talk about an example to fully understand the situation. But I'd wildly suggest to use `physical_property_ref` (without setting it to `is_derived == True`) for these cases.

But again, I need to see the example in a board or in a piece of paper.

ndaelman-hu:
> > ide an optional functionality. Would recommend something numpy-based as implementation.
> 
> I don't really understand what you are saying tbh. I am not talking about enforcing the storage of such quantities. I was talking about schema development. I think "enforcing" is a bit too strong of a word, but I had in mind the case that someone wants to store some energy contribution and implements a class for it, but doesn't create the corresponding force contribution because they personally do not store this in there simulations. But then another user comes along and wants to store this force contribution and now has to extend the schema themselves, whereas if the original person had been encouraged to add a corresponding force contribution it would have made everyone's life a lot easier.
> 
> Of course there might be some energy definitions that it doesn't make sense to define a corresponding force. In that case, obviously it should not be done, but I think in most cases it is possible at least in principle.

Ow, it sounded like you meant the actual storage, hence my reply.

As for schema extension, while the scenario you mention is indeed suboptimal, how would you enforce your suggestion?
We can _ask them to pretty-please follow this guideline_, either in the README and/or class description. For the latter, they should copy that instruction over. Or, _we couple both properties together_ in their own `ArchiveSection`. Not sure how the actual implementation would look like.

JosePizarro3:
@JFRudzinski bear in mind I merged something which created the sub-folder, `properties/energies.py`, with `FermiLevel` and `ChemicalPotential` defined in there. When you work again on this, you can use that then.

JFRudzinski:
> @JFRudzinski bear in mind I merged something which created the sub-folder, `properties/energies.py`, with `FermiLevel` and `ChemicalPotential` defined in there. When you work again on this, you can use that then.

Yes, got it, thank you :+1:

JosePizarro3:
@JFRudzinski to avoid confusion I am closing this in favor of #91 

You can add (in the description) a "Closes #numberofissue" so Github closes automatically the issue(s) related with #91

PR #53, opened 2024-04-11
Add BandGap output

### LINKED ISSUE #31: Add `BandGap` output
@ndaelman-hu

We have to add the `BandGap` base class with the same functionalities as in the normalizers. Important links to the old schema:

JosePizarro3:
Closes #31


JosePizarro3:
@JFRudzinski what do you think about creating a sub-folder with the specific physical properties modules? Like I did here for the ElectronicBandGap.

JFRudzinski:
> @JFRudzinski what do you think about creating a sub-folder with the specific physical properties modules? Like I did here for the ElectronicBandGap.

Not sure if there are any best practices in this direction, but personally, yeah, I think that is quite nice. I like separating out related properties a bit so none of the files become too too large to manage, and then in that case it definitely helps organizationally to have the subfolder :raised_hands:

JosePizarro3:
@ndaelman-hu being a band gap, would you mind reviewing this quickly? I think it is very easy, I just had to fix a bit the logic of the __setattr__ in PhysicalProperty. 

I'd like to merge this today, if there are not too many changes.

src/nomad_simulations/physical_property.py
```diff
@@ -56,6 +56,14 @@ class PhysicalProperty(ArchiveSection):
         """,
     )
 
+    iri = Quantity(
+        type=str,
```
ndaelman-hu: If we have any specialized type for URLs, use that instead. Else `str` is fine.

src/nomad_simulations/properties/band_gap.py
```diff
@@ -0,0 +1,147 @@
...
+
+from ..physical_property import PhysicalProperty
+
+
+class ElectronicBandGap(PhysicalProperty):
+    """
+    Energy difference between the highest occupied electronic state and the lowest unoccupied electronic state.
+    """
+
+    iri = 'https://fairmat-nfdi.github.io/fairmat-taxonomy/#/ElectronicBandGap'
```
ndaelman-hu: Given that this taxonomy is a work in progress, it's fine to link to the GitHub.
Once a version is established, it should be published with a permanent link (you can ask Shandor) to enforce stability.

src/nomad_simulations/properties/band_gap.py
```diff
@@ -0,0 +1,147 @@
...
+
+    # ? can `type` change character depending on the `variables`?
```
ndaelman-hu, line 36: Not sure if this is a comment or a question (forgot the semantics behind `# ?`)
If it's a question: the answer's yes, that's what I was pointing out before too.

src/nomad_simulations/properties/band_gap.py
```diff
@@ -0,0 +1,147 @@
...
+        description="""
+        If the `ElectronicBandGap` is `'indirect'`, the reciprocal momentum transfer for which the band gap is defined
+        in units of the `reciprocal_lattice_vectors`. The initial and final momentum 3D vectors are given in the first
+        and second element. Example, the momentum transfer in bulk Si2 happens between the Γ and the (approximately)
+        X points in the Brillouin zone; thus:
+            `momentum_transfer = [[0, 0, 0], [0.5, 0.5, 0]]`.
+
+        Note: this quantity only refers to scalar `value`, not to arrays of `value`.
+        """,
+    )
```
ndaelman-hu: Interesting: I was thinking of the difference vector between final and initial, but this might be better, not sure.
I guess this quantity is mostly relevant in transport processes, which now become direction dependent. Does the initial k-point vector matter then?

If we go with the difference vector, then `type=direct` is no longer a special case, and we default to the zero-vector there.

src/nomad_simulations/properties/band_gap.py
```diff
@@ -0,0 +1,147 @@
...
+
+    spin_channel = Quantity(
+        type=np.int32,
+        description="""
+        Spin channel of the corresponding `ElectronicBandGap`. It can take values of 0 or 1.
+        """,
+    )
```
ndaelman-hu: I was thinking, since the whole electronic structure is heavily dependent on the underlying k-space, basis set sampling, how about we group them together around that in something `ElectronicStructure`?

Granted, this would be 1 more layer of nesting, but then you'd know that the band structure, DOS, band gap, Fermi surface, etc. all come from the same representation.
Else, we will encounter cases with multiple samplings where we first have to loop all the properties and scan to find all the derived properties... My suggestion, meanwhile, covers FHI-aims and CP2k (one can split sampling by spin-channel).

If we end up going with my suggestion, then you can move `spin_channel` up to that section.

ndaelman-hu: - What about the spin-restricted case? Should this stay `None` or be `0`?
- What if we relax the spin, but no polarization occurs? Should we then have the same band gap value twice as a different property, or maybe store both `[0, 1]`?

The latter question really boils down to _which information do we want to convey by splitting out properties?_
If you don't go with my suggestion of `ElectronicStructure`, you could also just forego mentioning the spin-channel, and use an array for `value`.

src/nomad_simulations/properties/band_gap.py
```diff
@@ -0,0 +1,147 @@
...
+
+    value = Quantity(
+        type=np.float64,
+        unit='joule',
+        description="""
+        The value of the `ElectronicBandGap`.
+        """,
```
ndaelman-hu: """
The band gap size, i.e. a non-negative value.
"""

src/nomad_simulations/properties/band_gap.py
```diff
@@ -0,0 +1,147 @@
...
+        self.name = self.m_def.name
+        self.rank = []  # ? Is this here or in the attrs instantiation better?
+
+    def check_negative_values(self, logger: BoundLogger) -> pint.Quantity:
+        """
+        Checks if the electronic band gaps is negative and sets them to 0 if they are.
+
+        Args:
+            logger (BoundLogger): The logger to log messages.
+        """
```
ndaelman-hu: Should it automatically be set to zero?
I'm not so sure that we can recover from this error: I would print out an error message and leave `value` blank.

src/nomad_simulations/properties/band_gap.py
```diff
@@ -0,0 +1,147 @@
...
+        value = self.value.magnitude
+        if not isinstance(self.value.magnitude, np.ndarray):  # for scalars
+            value = np.array([value])
```
ndaelman-hu: I'm confused: the `Quantity` specification for `np.array` seems restricted to scalars (atm at least).

src/nomad_simulations/properties/band_gap.py
```diff
@@ -0,0 +1,147 @@
...
+            (Optional[str]): The resolved `type` of the electronic band gap.
+        """
+        if (
+            self.momentum_transfer is None or len(self.momentum_transfer) < 2
+        ) and self.type == 'indirect':
+            logger.warning(
+                "The `momentum_transfer` is not properly defined for an `type='indirect'` electronic band gap."
+            )
+            return None
+        if self.momentum_transfer is not None and len(self.momentum_transfer) > 0:
```
ndaelman-hu: the 2nd conditional is redundant given the 1 below

src/nomad_simulations/properties/band_gap.py
```diff
@@ -0,0 +1,147 @@
...
+        unit='joule',
+        description="""
+        The value of the `ElectronicBandGap`.
+        """,
+    )
+
+    def __init__(self, m_def: Section = None, m_context: Context = None, **kwargs):
+        super().__init__(m_def, m_context, **kwargs)
+        self.name = self.m_def.name
+        self.rank = []  # ? Is this here or in the attrs instantiation better?
```
ndaelman-hu: What do you mean with `attrs instantiation`?

src/nomad_simulations/properties/band_gap.py
```diff
@@ -0,0 +1,147 @@
...
+                return 'direct'
+            else:
+                return 'indirect'
+        return self.type
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
+
+        # Checks if the `value` is negative and sets it to 0 if it is.
+        self.value = self.check_negative_values(logger)
```
ndaelman-hu, line 145: Also add a check whether `self.value == None`. If so, throw an error.

src/nomad_simulations/properties/band_gap.py
ndaelman-hu, line 1: I remember that you wanted to have `OpticalBandGap` as a separate property, but can you remind me of the argumentation again?
If the normalization is the same (or very similar), then we could just distinguish using a `Quantity` similar to `label`.

src/nomad_simulations/properties/band_gap.py
ndaelman-hu: Btw, this kind of behavior is something that could fall under the tasks of `Quantity`.
Another alternative (but not my preference) is to define an interface `MinimalQuantities` for it.

tests/test_band_gap.py
```diff
@@ -0,0 +1,114 @@
...
+        )
+        assert electronic_band_gap.name == 'ElectronicBandGap'
+        assert electronic_band_gap.rank == []
+
+    @pytest.mark.parametrize(
+        'value, result',
+        [
+            (1.0, 1.0),
+            (-1.0, 0.0),
+        ],
```
ndaelman-hu: Maybe also add the case `(0., 0.)`

tests/test_band_gap.py
```diff
@@ -0,0 +1,114 @@
...
+        t_dependent_band_gap = ElectronicBandGap(
+            variables=[Temperature(grid_points=[0, 10, 20, 30] * ureg.kelvin)],
+            type='direct',
+        )
+        t_dependent_band_gap.value = [1.0, -2.0, 3.0, 4.0] * ureg.joule
+        t_dependent_band_gap.normalize(None, logger)
+        assert t_dependent_band_gap.type == 'direct'
+        assert (
+            np.isclose(t_dependent_band_gap.value.magnitude, [1.0, 0.0, 3.0, 4.0])
+        ).all()
```
ndaelman-hu: Maybe with the `pytest.mark.parametrize` decorator?
If you plan on using the vanilla and T-dependent band gaps more often, or if the setup is more involved, you could use a pytest fixture.

src/nomad_simulations/physical_property.py
```diff
@@ -185,24 +193,27 @@ def full_shape(self) -> list:
...
+                return Quantity(
                     type=quant.type,
                     unit=quant.unit,  # ? this can be moved to __setattr__
                     description=quant.description,
                 )
-                break
+
+    def __init__(self, m_def: Section = None, m_context: Context = None, **kwargs):
+        super().__init__(m_def, m_context, **kwargs)
+        # self._new_value = self._new_value
```
ndaelman-hu: You can remove this code comment.

src/nomad_simulations/physical_property.py
```diff
@@ -185,24 +193,27 @@ def full_shape(self) -> list:
         """
         return self.variables_shape + self.rank
 
-    def __init__(self, m_def: Section = None, m_context: Context = None, **kwargs):
-        super().__init__(m_def, m_context, **kwargs)
-
+    @property
+    def _new_value(self) -> Quantity:
         # initialize a `_new_value` quantity copying the main attrs from the `_value` quantity (`type`, `unit`,
```
ndaelman-hu: I guess you mean `value`, not the private `_value`?

src/nomad_simulations/physical_property.py
```diff
@@ -214,9 +225,9 @@ def __setattr__(self, name: str, val: Any) -> None:
                     f'The shape of the stored `value` {value_shape} does not match the full shape {self.full_shape} '
                     f'extracted from the variables `n_grid_points` and the `shape` defined in `PhysicalProperty`.'
                 )
-            self._new_value.shape = self.full_shape
-            self._new_value = val.magnitude * val.u
-            return super().__setattr__(name, self._new_value)
+            _new_value.shape = self.full_shape
+            _new_value = val.magnitude * val.u
+            return super().__setattr__(name, _new_value)
         return super().__setattr__(name, val)
```
ndaelman-hu, line 251: Given that you're using the `property` decorator, you can just define the setter and getter for `value` there.
No need for all these intermediate values. Check [FreeCodeCamp](https://www.freecodecamp.org/news/python-property-decorator/) out for an example.

At least, this would be an option in typical Python. No guarantee for our NOMAD dialect. See my other remarks on `value`.

src/nomad_simulations/properties/band_gap.py
ndaelman-hu: Okay, from your test below, I'm got a better understanding for your distinction between scalars and arrays in value.
Only after checking `outputs.py` did I see the hidden magic that powers this.

While in principle fine, I feel like here I would like some kind of indication, given that the `Quantity.shape` in and of itself is misleading. Probably you should use a `Variable` object directly, or extend `Quantity`.
Also consult with @TLCFEM here, since Markus is having him implement better type checks for the `Quantity`.
This might be something that breaks when his changes are applied.

ndaelman-hu (commented):
Hey, some nice implementations, but I would be careful with some pitfalls.
Regarding the schema, there are some points that I'd like clarified or where I offer an alternative.

src/nomad_simulations/physical_property.py
JosePizarro3: True, I changed it to URL.

src/nomad_simulations/properties/band_gap.py
JosePizarro3: Indeed, thanks for spotting this out. We should include the proper IRI, and not the web link. I changed it (even if the taxonomy is not published we just define it here like it should from the very beginning)

src/nomad_simulations/properties/band_gap.py
JosePizarro3: Can you give me an example?

src/nomad_simulations/properties/band_gap.py
JosePizarro3: I think we keep better the 2 vectors, as this is much simpler to extract from the bands. For other cases, users are free to set the first vector to [0, 0, 0]

src/nomad_simulations/properties/band_gap.py
JosePizarro3: I would prefer not. It is totally fine to have categories, but really we don't gain anything from having `spin_channel` as the unique quantity under some over-nested tree. Furthermore, there is no normalization done in different spin channels.

Let's simplify and keep it flat.

JosePizarro3: In this structure, both channels live separately, as elements of a list (`band_gap[0]` and `band_gap[1]`). The character of the state is determined by a complete different property (actually, by the **sign** of another physical property, the `magnetization`).

I prefer to keep 2 elements for `band_gap`. My argument also goes in the direction of keeping `variables` defined for each channel independently, to avoid potential issues. Your suggestion also implies having to restructure the shape of a `value` in a more complex way.

src/nomad_simulations/properties/band_gap.py
JosePizarro3: I added more description

src/nomad_simulations/properties/band_gap.py
JosePizarro3: I decided to keep it to 0. This is our default value anyways.

src/nomad_simulations/properties/band_gap.py
JosePizarro3: This is already discussed. We talked about meeting next week.

src/nomad_simulations/properties/band_gap.py
JosePizarro3: No, it is taken care of having a defined length for later operations

src/nomad_simulations/properties/band_gap.py
JosePizarro3: Similar to `iri`, setting it up at definition of the attributes of the class.

src/nomad_simulations/properties/band_gap.py
JosePizarro3: Because of some extra in-gap levels.

tests/test_band_gap.py
JosePizarro3: I think is easier to understand like this. It is quite powerful to call something "scalar_band_gap" and "t_dependent_band_gap", and I don't plan to change this setup in the future.

src/nomad_simulations/physical_property.py
JosePizarro3: Not sure if I understand: I need to anyways check some shapes to make sure that the setattr is consistent with the expected one extracted from concatenating and getting `full_shape`. 

I prefer to keep it here.

src/nomad_simulations/properties/band_gap.py
ndaelman-hu: Well, I would let the ultimate use case be the deciding factor. Anyhow, it's fine for now, we can revisit this when tackling transport properties. Just add a TODO comment.

src/nomad_simulations/properties/band_gap.py
ndaelman-hu: > we don't gain anything from having spin_channel as the unique quantity under some over-nested tree

Uhm, that's not quite what I meant: there's no over-nested tree, nor is spin-channel the only unique quantity.
Rather, I'm proposing something along the lines of the following (contrived) example below.
Note that the actual properties contained also differ. If they're the same, we can aggregate their spin channels.

```python
output = [
    ElectronicStructure(k_mesh=k_grid_1, spin_channel=[0], dos=DOS(...), band_gapElectronicBandGap(...)),
    ElectronicStructure(k_mesh=k_grid_1, spin_channel=[1], dos=DOS(...), band_gapElectronicBandGap(...)),
    ElectronicStructure(k_mesh=k_path_1, spin_channel=[0, 1], band_structure=BandStructure(...), dos=DOS(...), band_gapElectronicBandGap(...)),
]
```

You see? This setup emphasizes band structures, DOS, band gaps that come from the same setup.
If we flatten everything out, I first have to scan all `output` to find the matching `BandStructure` and `DOS`.
I'm really thinking of the user here, since I don't want them to have to write `for`-loops, just to extract this kind of information. This also predefines the packages for visualization, e.g. band structure and vertical DOS, side-by-side vs standalone, horizontal DOS.

ndaelman-hu: Well, my last suggestion is an alternative to the one above. I like that one better, tbh, and it's also fully compatible with your suggestion here.

Still, **you didn't really address my questions / concerns**:

- where are you communicating _spin-restriction_? Does a user always have to analyze the methodologies first before tackling the output?

1. While this might generally be the proper way of working, I'm not convinced everyone will comply. I think a slight redundancy is fine, if it eases the data handling.
2. Our queries (especially `ArchiveQuery`) enables _targeted data extraction_. If spin-restriction isn't communicated in some way or shape in the output, users will also have to download (part of) the method.

- how do you communicate that 2 spin-channels have the same band gap / DOS, i.e. the material exhibits _no spontaneous spin-polarization_? Again, I think we should _avoid that users have to build out their own analysis for something so basic_.

_That last phrase is really the determining factor for me._ Overall, I agree with your approach and want to stay clear from nesting. Still, one has to be a pragmatist, not an ideolog.

src/nomad_simulations/properties/band_gap.py
ndaelman-hu: > This is our default value anyways.

Ow, that's an **absolute no-go**. Under the motto "_better to say nothing than saying something false_", band gaps don't have a default.
A `value` of `None` indicates that we don't know (*). This has been discussed and settled before, and it's also what I implemented. Now, given `normalization`, the concept of "_not knowing_" is extended to include "_knowing that something is worng_".

Let me put it differently: a parser extracts `...-  0.5` as a `-.5 eV` band gap, misreading the dash as a minus sign.
The real value should thus be `.5 eV`, NOT `0 eV`. If we report `None` (with an error or warning), we've sufficiently communicated that we tried but failed to extract this value. Not all cases will be as simple as this, and I don't think there's a catch-all solution.

(*): I agree that storing `ElectronicBandGap` with `value = None` is kind of silly. You can also have it delete itself in that case. Just note the slightly technical trick: https://stackoverflow.com/questions/293431/python-object-deleting-itself

src/nomad_simulations/properties/band_gap.py
ndaelman-hu: So you're asking whether a band gap could ever become a tensor?
Naively, I'd say no, unless one starts considering different directions along the k-space, but that sounds very contrived.

src/nomad_simulations/properties/band_gap.py
ndaelman-hu: Ah, yes. How would you write that?

tests/test_band_gap.py
ndaelman-hu: > It is quite powerful to call something "scalar_band_gap" and "t_dependent_band_gap"

Yeah, that's completely fine. It's also quite powerful to indicate the core test logic (especially when it's as simple as here):

```python
@pytest.mark.parametrize(...)
def test_normalize(test_band_gap, results):
  test_band_gap.normalize(None, logger)
  assert test_band_gap.type == results['type']
  assert (
    np.isclose(t_dependent_band_gap.value.magnitude, results['values'])
  ).all()
```

You can define the setups above as objects (fixtures aren't needed then) and just add them to the decorator. Note how each component focuses exactly on its own task.

src/nomad_simulations/physical_property.py
ndaelman-hu: It's just cleaner, that's all. Take a look at the example I linked under the header `@property: Syntax and Logic` and decide yourself. If you want further help, we can tackle it in-person.

src/nomad_simulations/properties/band_gap.py
ndaelman-hu: Okay, just provide an update afterwards then.

src/nomad_simulations/properties/band_gap.py
ndaelman-hu: Ow, sorry, I misread before: I didn't take in `variables` properly, maybe call it `Variable` for clarity.
Coming back to the actual contents: not sure whether you mean the actual stored type, or the range of allowed types. In case of the latter, I think we should aim to cover all options regardless of the variables.

src/nomad_simulations/properties/band_gap.py
ndaelman-hu: You mean for `np.isclose(momentum_difference, np.zeros(3))`? Okay, but the logic is still clunkily formulated.
You can just roll up both `if`-statements into one:

```python
mtr = self.momentum_transfer if self.momentum_transfer is not None else []
if len(mtr) < 2:
    logger.warning(
       'The `momentum_transfer` should have at least two elements so that the difference can be calculated and the type of electronic band gap can be resolved.'
    )
    return None
```

Then the rest of the code no longer has to live under the 1st `if`-statement.
Moreover, I think you can further tighten the conditional to `len(mtr) != 3`.

src/nomad_simulations/properties/band_gap.py
```diff
@@ -0,0 +1,152 @@
...
+        """,
+    )
+
+    value = Quantity(
+        type=np.float64,
+        unit='joule',  # ! this units need to be fixed when we have dynamical units
+        description="""
+        The value of the electronic band gap. This value has to be positive, otherwise it will
+        be set to a 0 by the `normalize()` function.
+        """,
```
ndaelman-hu: I'd use a synonym, like "size" here to clarify.
Also, while it's good to express a default-value, I vehemently disagree with that specific default here (see my reply).

src/nomad_simulations/properties/band_gap.py
```diff
@@ -0,0 +1,152 @@
...
+    spin_channel = Quantity(
+        type=np.int32,
+        description="""
+        Spin channel of the corresponding electronic band gap. It can take values of 0 or 1.
+        """,
+    )
+
+    value = Quantity(
+        type=np.float64,
+        unit='joule',  # ! this units need to be fixed when we have dynamical units
```
ndaelman-hu: Do you mean just changing the scale to e.g. eV, Hartree, etc.?
Or also modifying the dimensionality (energy)? I don't really see a scenario for the latter.

ndaelman-hu:
Okay, I replied to unresolved (and some resolved) threads. Let's just work through those points, the rest is fine.

src/nomad_simulations/properties/band_gap.py
JosePizarro3: I was putting myself in the following situation:

Band gap varies with X. At some critical Xc, it changes `type` from `'direct'` to `'indirect'`. However, I don't know if such a variable X might exist.

I was thinking that anyways we should `type` and other id strings as general global quantities. I.e., when these become list they lost all their usefulness.

src/nomad_simulations/properties/band_gap.py
JosePizarro3: Transport or other calculations might have different definitions of momentum. This is the most general one I think.

src/nomad_simulations/properties/band_gap.py
JosePizarro3: In my mind, if you don't parse correctly the `value`, a property should not even be stored. I think this should be treated differently anyways, i.e., `PhysicalProperty.normalize()` must check that value is not None. I will change it.

src/nomad_simulations/properties/band_gap.py
JosePizarro3: No idea, we can tackle this later.

src/nomad_simulations/properties/band_gap.py
JosePizarro3: The latter, the dimensionality. Ok, I delete the comment.

src/nomad_simulations/properties/band_gap.py
JosePizarro3: I am not sure what is that you are writing.

A real case example would look like:
```python
simulation = Simulation(
    model_system=...,
    model_method=...,
    outputs=Outputs(
        electronic_band_gap=[
            ElectronicBandGap(spin_channel=0, physical_property_ref=electronic_dos[0], value=...),
            ElectronicBandGap(spin_channel=1, physical_property_ref=electronic_dos[1], value=...),
            ElectronicBandGap(physical_property_ref=electronic_band_structure[0], value=...),
        ],
        electronic_dos=[
            ElectronicDOS(spin_channel=0, value=...),
            ElectronicDOS(spin_channel=1, value=...),
            ElectronicDOS(physical_property_ref=electronic_band_structure[0], value=...),
        ],
        electronic_band_structure=[
            ElectronicBandStructure(value=...),
        ]
    )
)
```

But I doubt this example is realistic. In any case, you will not get rid of writing for loops in your example (if I understand correctly, you want 3 sections for electronic structure under outputs). Furthermore, why to do this classification for the electronic properties and not for other categories? What about when a property is shared amongst 2 or more categories?

I am trying to avoid that.

JosePizarro3: > where are you communicating spin-restriction? Does a user always have to analyze the methodologies first before tackling the output?

I'd argue that if it is a special case for only treating one spin_channel and not the other, indeed, there should be a flag indicating this, and spin_channel could be set to 0. 
But that does not mean the user needs to "analyze the methodologies first before tackling the output". I mean, for what? This is only an indicator of a calculation done in certain spin channels, hence `spin_channel` should already tell you about it.

> how do you communicate that 2 spin-channels have the same band gap / DOS, i.e. the material exhibits no spontaneous spin-polarization?

What? There is truly only 2 situations:
1. Simulation is not spin-polarized: then there is only one element for band gap and spin_channel is not assigned.
2. Simulation is spin polarized: there are 2 elements band gap and spin_channel is assign.

Now, if band gap is equal in both spin channels, it means that the state is either AFM or PM. Nobody is going to avoid having to go to another property (`magnetization`) and check the nature of the state.

___ 

In any case, I am not sure what you meant. I am not 100% sure your example above is realistic either.
Btw, this has nothing to do with me not understanding you, but rather on how you communicate this (so there is no need of "putting this in any other way"). Just a couple of suggestions (don't feel I am trying to be harsh, it is just that it gets very complex and lengthy any kind of review, and I'd like to keep it sharp to the point):
- Please, stay away from abusing the use of **bold** and _italics_. It truly makes your messages unreadable.  Furthermore, I (and probably others as well) understand what you mean in each phrase, as my brain is capable of emphasizing without the need of bolding or doing whole phrases in italics, so you should not worry about that but rather:
- Convey in defining a clear and concise idea in one or two phrases before[1] doing the review. If your messages are this long, and they keep being long reply after reply, it means that something is not going on properly. My suggestion would be that you re-summarize the message before sending it, not once, but a couple of times (this is also a problem I have btw, maybe I am overdoing it to the point my messages are cryptic, but I am sure there should be an intermediate point).


[1] See, I didn't need to bold **before** to put the emphasis, as I know you are more than capable of understanding where the emphasis lies in that phrase.

src/nomad_simulations/properties/band_gap.py
JosePizarro3: Ok, I changed a bit the logic to make it more understandable.

JosePizarro3:
@ndaelman-hu 

Ok, I added a new method in `Outputs` for returning the two spin channels of a property if present (otherwise it returns an empty list).

My reasoning for this is that the `data.outputs` can be a list connected with some model_system evolution, and hence we will be mixing concepts. Then, we would end up having an issue anyways of users needing to handle such situations manually. In any case, let's see what happens along the line, we can implement something better in a near future, no problem.

I also changed the negative `value` to be None, and I had to improve on the `__setattr__` for `PhysicalProperty`.

src/nomad_simulations/properties/band_gap.py
ndaelman-hu: > Band gap varies with X. At some critical Xc, it changes type from 'direct' to 'indirect'. However, I don't know if such a variable X might exist.

Rather than the variable, wouldn't your example more so depend on the conditions under which the band gap was measured / computed?

Anyhow, I agree that `type` as a global quantity is fine.

src/nomad_simulations/properties/band_gap.py
ndaelman-hu: Okay, but in the example that you lay out, how does the data scientist know

1. the spin channel of `ElectronicBandGap(physical_property_ref=electronic_band_structure[0], value=...)`?
Should she read its reference? Should it store the `spin_channel` of the property its refering to?
2. if the material (not the calculation) is spin-polarized or not? Is it now up to the data scientist to write a check herself?

src/nomad_simulations/properties/band_gap.py
ndaelman-hu: > In my mind, if you don't parse correctly the value, a property should not even be stored.

Yes, I'm basically making the same argument above. However, mistakes occur, and a parser might have a bug where it does generate nonsensical data. Preemptively capturing these bugs is where the normalizer comes in handy.

src/nomad_simulations/properties/band_gap.py
```diff
@@ -0,0 +1,154 @@
...
+            return 'indirect'
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
+
+        # Checks if the `value` is negative and sets it to None if it is.
+        self.value = self.check_negative_values(logger)
+        if self.value is None:
+            logger.error('The `value` of the electronic band gap is not stored.')
+            return
```
ndaelman-hu, line 149: Add a TODO note so we look into self-deletion somewhere down the line. I think that behavior is more appropriate.

src/nomad_simulations/outputs.py
```diff
@@ -94,12 +64,29 @@ class Outputs(ArchiveSection):
...
+            (list): The list of spin-polarized properties.
+        """
+        spin_polarized_properties = []
+        properties = getattr(self, property_name)
+        for prop in properties:
+            if prop.spin_channel is None:
+                continue
+            spin_polarized_properties.append(prop)
+        return spin_polarized_properties
+
```
ndaelman-hu, line 86: This is fine, but note that there's a distinction between a calculation that allowed for spin-relaxation and one where the properties differ between spin-channel.
I think that the former is already covered well enough in `model_system.py`.
I would use the latter definition here, though: if the DOS up / down are the same, I don't consider it polarized (magnetic).

src/nomad_simulations/properties/band_gap.py
ndaelman-hu: Okay, I saw that you more or less addressed this in your comment below + commits.

ndaelman-hu: > Now, if band gap is equal in both spin channels, it means that the state is either AFM or PM. Nobody is going to avoid having to go to another property (magnetization) and check the nature of the state.

I'm fine if you want to convey that information in another property, but then I would add it to this commit: a PM state is easily checked for.

The reason I'm mentioning this is that users will want to easily search and understand whether a property was spin-polarized because it was artificially constrained, or as an actual outcome of the setup.

JosePizarro3: A PM state is indistinguishable from AFM only by looking at BandGap. Maybe we can leave it for when we have a proper example and the `magnetization` defined?

src/nomad_simulations/outputs.py
JosePizarro3: Still, both spin-channels are reported. This distinction into the specific magnetic state should come from another property. If the proper occupation per spin channel is not defined, how can you ensure one or another type of magnetic state?

JosePizarro3:
Ok, I am merging this, but @ndaelman-hu: would you mind if we move this discussion of the magnetic state to #22? I think there we can keep talking + we can eventually invite Hongbin and the postdoc responsible for data to join the discussion 🙂

ndaelman-hu (approved):

Issue #52, opened 2024-04-11
Add `ElectronicEigenvalues` and `ElectronicBandStructure` properties


JosePizarro3:
Adding a base class for the `ElectronicEigenvalues(PhysicalProperty)`, `ElectronicBandStructure(PhysicalProperty)` and `FermiSurface(PhysicalProperty)`. These should contain the electronic energies in terms of the k points for different conditions.

@ndaelman-hu question here: do we define a new `Variables` class for the 3d k points, or simply add a `k_mesh_ref`? I would go for the second option, as we do not want to keep things doubly stored, but maybe is a good oportunity to think about the interplay between ModelMethod and Variables?

JosePizarro3:
I forgot: this is connected with #33, as `ElectronicBandStructure` should inherit from this class. At least this is what I had in mind, it is open for discussion.

ndaelman-hu:
> @ndaelman-hu question here: do we define a new `Variables` class for the 3d k points, or simply add a `k_mesh_ref`? I would go for the second option, as we do not want to keep things doubly stored, but maybe is a good oportunity to think about the interplay between ModelMethod and Variables?

I'd also rather go for the 2nd suggestion, yes. Not only to circumvent doubly storing, but also doubly writing the same information.

JosePizarro3:
After our meeting with Alvin: should we define the value as a HDF5Reference for the ElectronicEigenvalues? This might save tons of storage (but I am not thinking we have to store eigenvalues ALL the time)

ndaelman-hu:
> After our meeting with Alvin: should we define the value as a HDF5Reference for the ElectronicEigenvalues? This might save tons of storage (but I am not thinking we have to store eigenvalues ALL the time)

This touches again on _which actor decides the storage mode_: the schema, normalizer, or parser? In case of the former, wouldn't we always be forced to store it as such?

Personally, I think that we should have some automated system that decides at which data size to store in HDF5.
NOMAD should then handle reading/writing to message pack or HDF5 automatically. But for now this is a pipe dream...

So, if we have to make a generic choice via the schema, I'm kind of unsure. I haven't seen any case yet where eigenvalues warrant the heavy storage. I think that electron densities and wavefunctions are more appropriate use case.

We could in principle decide to **write almost all longer arrays to HDF5**, though, retaining mostly small metadata in message pack. My main worries there are:

1. the _HDF5 implementation is quite new_, so we should test it a bit first.
[...]

PR #54, opened 2024-04-12
Add ElectronicDOS and XASSpectra properties

### LINKED ISSUE #32: Add `ElectronicDOS` and `XASSpectra` properties
@ndaelman-hu

We have to add the `ElectronicDOS` and `XASSpectra` base class with the same functionalities as in the normalizers. Important links to the old schema:

[Spectra normalizer](https://github.com/nomad-coe/nomad-normalizer-plugin-spectra)

JosePizarro3:
@ndaelman-hu 

@JFRudzinski @Bernadette-Mohr only for you: I added `properties/energies.py` and a couple of properties, `FermiLevel` and `ChemicalPotential`. You can disregard the rest of the message.

I just finished this initial version. Mainly, you have to check `properties/spectral_profile.py` and `properties/energies.py`. I added some tests, but I have the issue that I would need some other sections in order to properly test the functionalities in `ElectronicDensityOfStates`, namely, I need `ElectronicEigenvalues` containing the info of the `highest_occupied_energy` and `lowest_occupied_energy`.

Besides these, I changed `OrbitalsState` and `AtomsState` to inherit from `Entity`. Do you mind giving your thoughts in here? I wanted to give an `entity_ref` that can be used to reference if a property is atomic or orbitally resolved, e.g., the projected DOS (but there might be others).

### DIFF
--- a/src/nomad_simulations/atoms_state.py
+++ b/src/nomad_simulations/atoms_state.py
@@ -25,12 +25,13 @@ from nomad.units import ureg
 
 from nomad.metainfo import Quantity, SubSection, MEnum, Section, Context
 from nomad.datamodel.data import ArchiveSection
+from nomad.datamodel.metainfo.basesections import Entity
 from nomad.datamodel.metainfo.annotations import ELNAnnotation
 
 from .utils import RussellSaundersState
 
 
-class OrbitalsState(ArchiveSection):
+class OrbitalsState(Entity):
     """
     A base section used to define the orbital state of an atom.
     """
@@ -537,7 +538,7 @@ class HubbardInteractions(ArchiveSection):
                 )
 
 
-class AtomsState(ArchiveSection):
+class AtomsState(Entity):
     """
     A base section to define each atom state information.
     """
--- a/src/nomad_simulations/outputs.py
+++ b/src/nomad_simulations/outputs.py
@@ -42,7 +42,7 @@ def resolve_output(section, section_instance):
     return quantity
 
 
-class Outputs(PhysicalProperty):
+class Outputs(ArchiveSection):
     """
     Output properties of a simulation. This base class can be used for inheritance in any of the output properties
     defined in this schema.
@@ -116,7 +116,7 @@ class Outputs(PhysicalProperty):
         return None
 
     def normalize(self, archive, logger) -> None:
-        # super().normalize(archive, logger)
+        super().normalize(archive, logger)
 
         # Set ref to the last ModelSystem if this is not set in the output
         if self.model_system_ref is None:
--- a/src/nomad_simulations/outputs.py
+++ b/src/nomad_simulations/outputs.py
@@ -57,14 +57,6 @@ class Outputs(ArchiveSection):
         a_eln=ELNAnnotation(component='ReferenceEditQuantity'),
     )
 
-    custom_physical_property = SubSection(
-        sub_section=PhysicalProperty.m_def,
-        repeats=True,
-        description="""
-        A custom physical property used to store properties not yet covered by the NOMAD schema.
-        """,
-    )
-
     # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
     # List of properties
     # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
--- a/src/nomad_simulations/outputs.py
+++ b/src/nomad_simulations/outputs.py
@@ -31,6 +31,18 @@ from .properties import (
 )
 
 
+def resolve_output(section, section_instance):
+    quantity = None
+    if section.m_parent is None:
+        return quantity
+    for output in section.m_parent.outputs:
+        if isinstance(output, section_instance):
+            quantity = output.value
+            if quantity is not None:
+                break
+    return quantity
+
+
 class Outputs(ArchiveSection):
     """
     Output properties of a simulation. This base class can be used for inheritance in any of the output properties
--- a/src/nomad_simulations/outputs.py
+++ b/src/nomad_simulations/outputs.py
@@ -28,6 +28,10 @@ from .physical_property import PhysicalProperty
 from .numerical_settings import SelfConsistency
 from .properties import (
     ElectronicBandGap,
+    FermiLevel,
+    ChemicalPotential,
+    ElectronicDensityOfStates,
+    XASSpectra,
 )
 
 
@@ -66,6 +70,16 @@ class Outputs(ArchiveSection):
     # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 
     electronic_band_gap = SubSection(sub_section=ElectronicBandGap.m_def, repeats=True)
+
+    fermi_level = SubSection(sub_section=FermiLevel.m_def, repeats=True)
+
+    chemical_potential = SubSection(sub_section=ChemicalPotential.m_def, repeats=True)
+
+    electronic_dos = SubSection(
+        sub_section=ElectronicDensityOfStates.m_def, repeats=True
+    )
+
+    xas_spectra = SubSection(sub_section=XASSpectra.m_def, repeats=True)
 
     # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
     # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
--- a/src/nomad_simulations/outputs.py
+++ b/src/nomad_simulations/outputs.py
@@ -69,11 +69,11 @@ class Outputs(ArchiveSection):
     # List of properties
     # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 
-    electronic_band_gap = SubSection(sub_section=ElectronicBandGap.m_def, repeats=True)
-
     fermi_level = SubSection(sub_section=FermiLevel.m_def, repeats=True)
 
     chemical_potential = SubSection(sub_section=ChemicalPotential.m_def, repeats=True)
+
+    electronic_band_gap = SubSection(sub_section=ElectronicBandGap.m_def, repeats=True)
 
     electronic_dos = SubSection(
         sub_section=ElectronicDensityOfStates.m_def, repeats=True
--- a/src/nomad_simulations/outputs.py
+++ b/src/nomad_simulations/outputs.py
@@ -22,7 +22,6 @@ from typing import Optional
 from nomad.datamodel.data import ArchiveSection
 from nomad.metainfo import Quantity, SubSection
 from nomad.datamodel.metainfo.annotations import ELNAnnotation
-
 from .model_system import ModelSystem
 from .physical_property import PhysicalProperty
 from .numerical_settings import SelfConsistency
@@ -43,7 +42,7 @@ def resolve_output(section, section_instance):
     return quantity
 
 
-class Outputs(ArchiveSection):
+class Outputs(PhysicalProperty):
     """
     Output properties of a simulation. This base class can be used for inheritance in any of the output properties
     defined in this schema.
@@ -117,7 +116,7 @@ class Outputs(ArchiveSection):
         return None
 
     def normalize(self, archive, logger) -> None:
-        super().normalize(archive, logger)
+        # super().normalize(archive, logger)
 
         # Set ref to the last ModelSystem if this is not set in the output
         if self.model_system_ref is None:
--- a/src/nomad_simulations/outputs.py
+++ b/src/nomad_simulations/outputs.py
@@ -22,24 +22,13 @@ from typing import Optional
 from nomad.datamodel.data import ArchiveSection
 from nomad.metainfo import Quantity, SubSection
 from nomad.datamodel.metainfo.annotations import ELNAnnotation
+
 from .model_system import ModelSystem
 from .physical_property import PhysicalProperty
 from .numerical_settings import SelfConsistency
 from .properties import (
     ElectronicBandGap,
 )
-
-
-def resolve_output(section, section_instance):
-    quantity = None
-    if section.m_parent is None:
-        return quantity
-    for output in section.m_parent.outputs:
-        if isinstance(output, section_instance):
-            quantity = output.value
-            if quantity is not None:
-                break
-    return quantity
 
 
 class Outputs(ArchiveSection):
--- a/src/nomad_simulations/properties/__init__.py
+++ b/src/nomad_simulations/properties/__init__.py
@@ -17,3 +17,4 @@
 # limitations under the License.
 
 from .band_gap import ElectronicBandGap
+from .spectral_profile import SpectralProfile
--- a/src/nomad_simulations/properties/__init__.py
+++ b/src/nomad_simulations/properties/__init__.py
@@ -18,4 +18,9 @@
 
 from .energies import FermiLevel, ChemicalPotential
 from .band_gap import ElectronicBandGap
-from .spectral_profile import SpectralProfile, ElectronicDensityOfStates, XASSpectra
+from .spectral_profile import (
+    SpectralProfile,
+    DOSProfile,
+    ElectronicDensityOfStates,
+    XASSpectra,
+)
--- a/src/nomad_simulations/properties/__init__.py
+++ b/src/nomad_simulations/properties/__init__.py
@@ -16,5 +16,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from .energies import FermiLevel, ChemicalPotential
 from .band_gap import ElectronicBandGap
-from .spectral_profile import SpectralProfile
+from .spectral_profile import SpectralProfile, ElectronicDensityOfStates, XASSpectra
--- a/src/nomad_simulations/properties/band_gap.py
+++ b/src/nomad_simulations/properties/band_gap.py
@@ -96,10 +96,7 @@ class ElectronicBandGap(PhysicalProperty):
 
         # Set the value to 0 when it is negative
         if (value < 0).any():
-            logger.warning('The electronic band gap cannot be defined negative.')
-            # return None
-            # value[value < 0] = None
-            # return value * self.value.u
+            logger.error('The electronic band gap cannot be defined negative.')
             return None
 
         if not isinstance(self.value.magnitude, np.ndarray):  # for scalars
--- a/src/nomad_simulations/properties/band_gap.py
+++ b/src/nomad_simulations/properties/band_gap.py
@@ -67,7 +67,7 @@ class ElectronicBandGap(PhysicalProperty):
 
     value = Quantity(
         type=np.float64,
-        unit='joule',
+        unit='joule',  # ! this units need to be fixed when we have dynamical units
         description="""
         The value of the electronic band gap. This value has to be positive, otherwise it will
         prop an error and be set to None by the `normalize()` function.
--- a/src/nomad_simulations/properties/band_gap.py
+++ b/src/nomad_simulations/properties/band_gap.py
@@ -31,6 +31,8 @@ class ElectronicBandGap(PhysicalProperty):
     Energy difference between the highest occupied electronic state and the lowest unoccupied electronic state.
     """
 
+    # ! implement `iri` and `rank` as part of `m_def = Section()`
+
     iri = 'http://fairmat-nfdi.eu/taxonomy/ElectronicBandGap'
 
     type = Quantity(
--- a/src/nomad_simulations/properties/band_gap.py
+++ b/src/nomad_simulations/properties/band_gap.py
@@ -81,62 +81,6 @@ class ElectronicBandGap(PhysicalProperty):
         self.name = self.m_def.name
         self.rank = []
 
-    def check_negative_values(self, logger: BoundLogger) -> Optional[pint.Quantity]:
-        """
-        Checks if the electronic band gaps is negative and sets them to None if they are.
-
-        Args:
-            logger (BoundLogger): The logger to log messages.
-        """
-        value = self.value.magnitude
-        if not isinstance(self.value.magnitude, np.ndarray):  # for scalars
-            value = np.array(
-                [value]
-            )  # ! check this when talking with Lauri and Theodore
-
-        # Set the value to 0 when it is negative
-        if (value < 0).any():
-            logger.error('The electronic band gap cannot be defined negative.')
-            return None
-
-        if not isinstance(self.value.magnitude, np.ndarray):  # for scalars
-            value = value[0]
-        return value * self.value.u
-
-    def resolve_type(self, logger: BoundLogger) -> Optional[str]:
-        """
-        Resolves the `type` of the electronic band gap based on the stored `momentum_transfer` values.
-
-        Args:
-            logger (BoundLogger): The logger to log messages.
-
-        Returns:
-            (Optional[str]): The resolved `type` of the electronic band gap.
-        """
-        mtr = self.momentum_transfer if self.momentum_transfer is not None else []
-
-        # Check if the `momentum_transfer` is [], and return the type and a warning in the log for `indirect` band gaps
-        if len(mtr) == 0:
-            if self.type == 'indirect':
-                logger.warning(
-                    'The `momentum_transfer` is not stored for an `indirect` band gap.'
-                )
-            return self.type
-
-        # Check if the `momentum_transfer` has at least two elements, and return None if it does not
-        if len(mtr) == 1:
-            logger.warning(
-                'The `momentum_transfer` should have at least two elements so that the difference can be calculated and the type of electronic band gap can be resolved.'
-            )
-            return None
-
-        # Resolve `type` from the difference between the initial and final momentum transfer
-        momentum_difference = np.diff(mtr, axis=0)
-        if (np.isclose(momentum_difference, np.zeros(3))).all():
-            return 'direct'
-        else:
-            return 'indirect'
-
     def _check_negative_values(self, logger: BoundLogger) -> Optional[pint.Quantity]:
         """
         Checks if the electronic band gaps is negative and sets them to None if they are.
--- a/src/nomad_simulations/properties/band_gap.py
+++ b/src/nomad_simulations/properties/band_gap.py
@@ -67,7 +67,7 @@ class ElectronicBandGap(PhysicalProperty):
 
     value = Quantity(
         type=np.float64,
-        unit='joule',  # ! this units need to be fixed when we have dynamical units
+        unit='joule',
         description="""
         The value of the electronic band gap. This value has to be positive, otherwise it will
         prop an error and be set to None by the `normalize()` function.
--- a/src/nomad_simulations/properties/band_gap.py
+++ b/src/nomad_simulations/properties/band_gap.py
@@ -21,7 +21,6 @@ from structlog.stdlib import BoundLogger
 import pint
 from typing import Optional
 
-from nomad.units import ureg
 from nomad.metainfo import Quantity, MEnum, Section, Context
 
 from ..physical_property import PhysicalProperty
@@ -68,7 +67,7 @@ class ElectronicBandGap(PhysicalProperty):
 
     value = Quantity(
         type=np.float64,
-        unit='joule',
+        unit='joule',  # ! this units need to be fixed when we have dynamical units
         description="""
         The value of the electronic band gap. This value has to be positive, otherwise it will
         prop an error and be set to None by the `normalize()` function.
@@ -91,7 +90,9 @@ class ElectronicBandGap(PhysicalProperty):
         """
         value = self.value.magnitude
         if not isinstance(self.value.magnitude, np.ndarray):  # for scalars
-            value = np.array([value])
+            value = np.array(
+                [value]
+            )  # ! check this when talking with Lauri and Theodore
 
         # Set the value to 0 when it is negative
         if (value < 0).any():
--- a/src/nomad_simulations/properties/band_gap.py
+++ b/src/nomad_simulations/properties/band_gap.py
@@ -84,17 +84,25 @@ class ElectronicBandGap(PhysicalProperty):
 
     def check_negative_values(self, logger: BoundLogger) -> pint.Quantity:
         """
-        Checks if the electronic band gap is negative and sets it to 0 if it is.
+        Checks if the electronic band gaps is negative and sets them to 0 if they are.
 
         Args:
             logger (BoundLogger): The logger to log messages.
         """
-        if self.value < 0.0:
-            logger.error(
-                'The electronic band gap cannot be defined negative. We set it up to 0.'
+        value = self.value.magnitude
+        if not isinstance(self.value.magnitude, np.ndarray):  # for scalars
+            value = np.array([value])
+
+        # Set the value to 0 when it is negative
+        if (value < 0).any():
+            logger.warning(
+                'The electronic band gap cannot be defined negative. We set them up to 0.'
             )
-            return 0.0 * ureg.eV
-        return self.value
+        value[value < 0] = 0
+
+        if not isinstance(self.value.magnitude, np.ndarray):  # for scalars
+            value = value[0]
+        return value * self.value.u
 
     def resolve_type(self, logger: BoundLogger) -> Optional[str]:
         """
@@ -106,14 +114,21 @@ class ElectronicBandGap(PhysicalProperty):
         Returns:
             (Optional[str]): The resolved `type` of the electronic band gap.
         """
-        if self.momentum_transfer is None and self.type == 'indirect':
+        if (
+            self.momentum_transfer is None or len(self.momentum_transfer) < 2
+        ) and self.type == 'indirect':
             logger.warning(
-                "The `momentum_transfer` is not defined for an `type='indirect'` electronic band gap."
+                "The `momentum_transfer` is not properly defined for an `type='indirect'` electronic band gap."
             )
             return None
-        if self.momentum_transfer is not None:
+        if self.momentum_transfer is not None and len(self.momentum_transfer) > 0:
+            if len(self.momentum_transfer) == 1:
+                logger.warning(
+                    'The `momentum_transfer` should have at least two elements so that the difference can be calculated and the type of electronic band gap can be resolved.'
+                )
+                return None
             momentum_difference = np.diff(self.momentum_transfer, axis=0)
-            if (momentum_difference == np.zeros(3)).all():
+            if (np.isclose(momentum_difference, np.zeros(3))).all():
                 return 'direct'
             else:
                 return 'indirect'
--- a/src/nomad_simulations/properties/band_gap.py
+++ b/src/nomad_simulations/properties/band_gap.py
@@ -67,7 +67,7 @@ class ElectronicBandGap(PhysicalProperty):
 
     value = Quantity(
         type=np.float64,
-        unit='joule',  # ! this units need to be fixed when we have dynamical units
+        unit='joule',
         description="""
         The value of the electronic band gap. This value has to be positive, otherwise it will
         prop an error and be set to None by the `normalize()` function.
@@ -81,9 +81,9 @@ class ElectronicBandGap(PhysicalProperty):
         self.name = self.m_def.name
         self.rank = []  # ? Is this here or in the attrs instantiation better?
 
-    def check_negative_values(self, logger: BoundLogger) -> pint.Quantity:
+    def check_negative_values(self, logger: BoundLogger) -> Optional[pint.Quantity]:
         """
-        Checks if the electronic band gaps is negative and sets them to 0 if they are.
+        Checks if the electronic band gaps is negative and sets them to None if they are.
 
         Args:
             logger (BoundLogger): The logger to log messages.
@@ -96,10 +96,11 @@ class ElectronicBandGap(PhysicalProperty):
 
         # Set the value to 0 when it is negative
         if (value < 0).any():
-            logger.warning(
-                'The electronic band gap cannot be defined negative. We set them up to 0.'
-            )
-        value[value < 0] = 0
+            logger.warning('The electronic band gap cannot be defined negative.')
+            # return None
+            # value[value < 0] = None
+            # return value * self.value.u
+            return None
 
         if not isinstance(self.value.magnitude, np.ndarray):  # for scalars
             value = value[0]
--- a/src/nomad_simulations/properties/band_gap.py
+++ b/src/nomad_simulations/properties/band_gap.py
@@ -79,7 +79,7 @@ class ElectronicBandGap(PhysicalProperty):
     ) -> None:
         super().__init__(m_def, m_context, **kwargs)
         self.name = self.m_def.name
-        self.rank = []  # ? Is this here or in the attrs instantiation better?
+        self.rank = []
 
     def check_negative_values(self, logger: BoundLogger) -> Optional[pint.Quantity]:
         """
--- a/src/nomad_simulations/properties/band_gap.py
+++ b/src/nomad_simulations/properties/band_gap.py
@@ -113,25 +113,29 @@ class ElectronicBandGap(PhysicalProperty):
         Returns:
             (Optional[str]): The resolved `type` of the electronic band gap.
         """
-        if (
-            self.momentum_transfer is None or len(self.momentum_transfer) < 2
-        ) and self.type == 'indirect':
+        mtr = self.momentum_transfer if self.momentum_transfer is not None else []
+
+        # Check if the `momentum_transfer` is [], and return the type and a warning in the log for `indirect` band gaps
+        if len(mtr) == 0:
+            if self.type == 'indirect':
+                logger.warning(
+                    'The `momentum_transfer` is not stored for an `indirect` band gap.'
+                )
+            return self.type
+
+        # Check if the `momentum_transfer` has at least two elements, and return None if it does not
+        if len(mtr) == 1:
             logger.warning(
-                "The `momentum_transfer` is not properly defined for an `type='indirect'` electronic band gap."
+                'The `momentum_transfer` should have at least two elements so that the difference can be calculated and the type of electronic band gap can be resolved.'
             )
             return None
-        if self.momentum_transfer is not None and len(self.momentum_transfer) > 0:
-            if len(self.momentum_transfer) == 1:
-                logger.warning(
-                    'The `momentum_transfer` should have at least two elements so that the difference can be calculated and the type of electronic band gap can be resolved.'
-                )
-                return None
-            momentum_difference = np.diff(self.momentum_transfer, axis=0)
-            if (np.isclose(momentum_difference, np.zeros(3))).all():
-                return 'direct'
-            else:
-                return 'indirect'
-        return self.type
+
+        # Resolve `type` from the difference between the initial and final momentum transfer
+        momentum_difference = np.diff(mtr, axis=0)
+        if (np.isclose(momentum_difference, np.zeros(3))).all():
+            return 'direct'
+        else:
+            return 'indirect'
 
     def _check_negative_values(self, logger: BoundLogger) -> Optional[pint.Quantity]:
         """
--- a/src/nomad_simulations/properties/band_gap.py
+++ b/src/nomad_simulations/properties/band_gap.py
@@ -21,6 +21,7 @@ from structlog.stdlib import BoundLogger
 import pint
 from typing import Optional
 
+from nomad.units import ureg
 from nomad.metainfo import Quantity, MEnum, Section, Context
 
 from ..physical_property import PhysicalProperty
@@ -79,7 +80,44 @@ class ElectronicBandGap(PhysicalProperty):
     ) -> None:
         super().__init__(m_def, m_context, **kwargs)
         self.name = self.m_def.name
-        self.rank = []
+        self.rank = []  # ? Is this here or in the attrs instantiation better?
+
+    def check_negative_values(self, logger: BoundLogger) -> pint.Quantity:
+        """
+        Checks if the electronic band gap is negative and sets it to 0 if it is.
+
+        Args:
+            logger (BoundLogger): The logger to log messages.
+        """
+        if self.value < 0.0:
+            logger.error(
+                'The electronic band gap cannot be defined negative. We set it up to 0.'
+            )
+            return 0.0 * ureg.eV
+        return self.value
+
+    def resolve_type(self, logger: BoundLogger) -> Optional[str]:
+        """
+        Resolves the `type` of the electronic band gap based on the stored `momentum_transfer` values.
+
+        Args:
+            logger (BoundLogger): The logger to log messages.
+
+        Returns:
+            (Optional[str]): The resolved `type` of the electronic band gap.
+        """
+        if self.momentum_transfer is None and self.type == 'indirect':
+            logger.warning(
+                "The `momentum_transfer` is not defined for an `type='indirect'` electronic band gap."
+            )
+            return None
+        if self.momentum_transfer is not None:
+            momentum_difference = np.diff(self.momentum_transfer, axis=0)
+            if (momentum_difference == np.zeros(3)).all():
+                return 'direct'
+            else:
+                return 'indirect'
+        return self.type
 
     def _check_negative_values(self, logger: BoundLogger) -> Optional[pint.Quantity]:
         """
--- a/src/nomad_simulations/properties/energies.py
+++ b/src/nomad_simulations/properties/energies.py
@@ -28,11 +28,11 @@ class FermiLevel(PhysicalProperty):
     Energy required to add or extract a charge from a material at zero temperature. It can be also defined as the chemical potential at zero temperature.
     """
 
-    iri = 'https://fairmat-nfdi.github.io/fairmat-taxonomy/FermiLevel'
+    iri = 'http://fairmat-nfdi.eu/taxonomy/FermiLevel'
 
     value = Quantity(
         type=np.float64,
-        unit='joule',  # ! this units need to be fixed when we have dynamical units
+        unit='joule',
         description="""
         Value of the Fermi level.
         """,
@@ -51,11 +51,11 @@ class ChemicalPotential(PhysicalProperty):
     Free energy cost of adding or extracting a particle from a thermodynamic system.
     """
 
-    iri = 'https://fairmat-nfdi.github.io/fairmat-taxonomy/ChemicalPotential'
+    iri = 'http://fairmat-nfdi.eu/taxonomy/ChemicalPotential'
 
     value = Quantity(
         type=np.float64,
-        unit='joule',  # ! this units need to be fixed when we have dynamical units
+        unit='joule',
         description="""
         Value of the chemical potential.
         """,
--- a/src/nomad_simulations/properties/energies.py
+++ b/src/nomad_simulations/properties/energies.py
@@ -34,13 +34,13 @@ class FermiLevel(PhysicalProperty):
         type=np.float64,
         unit='joule',
         description="""
-        Value of the Fermi level.
+        The value of the Fermi level.
         """,
     )
 
     def __init__(self, m_def: Section = None, m_context: Context = None, **kwargs):
         super().__init__(m_def, m_context, **kwargs)
-        self.rank = []  # ? Is this here or in the attrs instantiation better?
+        self.rank = []
 
     def normalize(self, archive, logger) -> None:
         super().normalize(archive, logger)
@@ -57,13 +57,13 @@ class ChemicalPotential(PhysicalProperty):
         type=np.float64,
         unit='joule',
         description="""
-        Value of the chemical potential.
+        The value of the chemical potential.
         """,
     )
 
     def __init__(self, m_def: Section = None, m_context: Context = None, **kwargs):
         super().__init__(m_def, m_context, **kwargs)
-        self.rank = []  # ? Is this here or in the attrs instantiation better?
+        self.rank = []
 
     def normalize(self, archive, logger) -> None:
         super().normalize(archive, logger)
--- a/src/nomad_simulations/properties/energies.py
+++ b/src/nomad_simulations/properties/energies.py
@@ -1,0 +1,69 @@
+#
+# Copyright The NOMAD Authors.
+#
+# This file is part of NOMAD. See https://nomad-lab.eu for further info.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import numpy as np
+
+from nomad.metainfo import Quantity, Section, Context
+
+from ..physical_property import PhysicalProperty
+
+
+class FermiLevel(PhysicalProperty):
+    """
+    Energy required to add or extract a charge from a material at zero temperature. It can be also defined as the chemical potential at zero temperature.
+    """
+
+    iri = 'https://fairmat-nfdi.github.io/fairmat-taxonomy/FermiLevel'
+
+    value = Quantity(
+        type=np.float64,
+        unit='joule',  # ! this units need to be fixed when we have dynamical units
+        description="""
+        Value of the Fermi level.
+        """,
+    )
+
+    def __init__(self, m_def: Section = None, m_context: Context = None, **kwargs):
+        super().__init__(m_def, m_context, **kwargs)
+        self.rank = []  # ? Is this here or in the attrs instantiation better?
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
+
+
+class ChemicalPotential(PhysicalProperty):
+    """
+    Free energy cost of adding or extracting a particle from a thermodynamic system.
+    """
+
+    iri = 'https://fairmat-nfdi.github.io/fairmat-taxonomy/ChemicalPotential'
+
+    value = Quantity(
+        type=np.float64,
+        unit='joule',  # ! this units need to be fixed when we have dynamical units
+        description="""
+        Value of the chemical potential.
+        """,
+    )
+
+    def __init__(self, m_def: Section = None, m_context: Context = None, **kwargs):
+        super().__init__(m_def, m_context, **kwargs)
+        self.rank = []  # ? Is this here or in the attrs instantiation better?
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
--- a/src/nomad_simulations/properties/energies.py
+++ b/src/nomad_simulations/properties/energies.py
@@ -43,6 +43,7 @@ class FermiLevel(PhysicalProperty):
     ) -> None:
         super().__init__(m_def, m_context, **kwargs)
         self.rank = []
+        self.name = self.m_def.name
 
     def normalize(self, archive, logger) -> None:
         super().normalize(archive, logger)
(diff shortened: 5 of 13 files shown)


JFRudzinski:
> regard the rest of the messa

Thanks @JosePizarro3, I did see this. I will eventually use this as a basis and combine with the other energies that I had written down

src/nomad_simulations/atoms_state.py
```diff
@@ -25,12 +25,13 @@
...
 from nomad.metainfo import Quantity, SubSection, MEnum, Section, Context
 from nomad.datamodel.data import ArchiveSection
+from nomad.datamodel.metainfo.basesections import Entity
 from nomad.datamodel.metainfo.annotations import ELNAnnotation
 
 from .utils import RussellSaundersState
 
 
-class OrbitalsState(ArchiveSection):
+class OrbitalsState(Entity):
```
ndaelman-hu, line 34: This is fine, though a proper implementation would use _dependency injection_ for ontology references.
Since we don't have such a framework in NOMAD, this is okay. Trying to implement any, is outside the scope of this PR, anyways.

src/nomad_simulations/properties/energies.py
```diff
@@ -0,0 +1,75 @@
...
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
+
+
+class ChemicalPotential(PhysicalProperty):
+    """
+    Free energy cost of adding or extracting a particle from a thermodynamic system.
+    """
+
+    iri = 'http://fairmat-nfdi.eu/taxonomy/ChemicalPotential'
```
ndaelman-hu: Note that you also need a specification for which particle the chemical potential value is given.
In chemical systems, you may have various molecules that apply (reference to AtomsState, Orbitalstate?).

src/nomad_simulations/properties/energies.py
```diff
@@ -0,0 +1,75 @@
...
+
+from ..physical_property import PhysicalProperty
+
+
+class FermiLevel(PhysicalProperty):
+    """
+    Energy required to add or extract a charge from a material at zero temperature. It can be also defined as the chemical potential at zero temperature.
+    """
+
+    iri = 'http://fairmat-nfdi.eu/taxonomy/FermiLevel'
```
ndaelman-hu, line 33: 1. If links don't exist yet, keep them as comments for now.
2. `http`? I think you mean `https`

src/nomad_simulations/properties/energies.py
```diff
@@ -0,0 +1,75 @@
...
+
+from nomad.metainfo import Quantity, Section, Context
+
+from ..physical_property import PhysicalProperty
+
+
+class FermiLevel(PhysicalProperty):
+    """
+    Energy required to add or extract a charge from a material at zero temperature. It can be also defined as the chemical potential at zero temperature.
+    """
```
ndaelman-hu, line 29: > Energy required to add or extract a charge from a material at zero temperature.

Firstly, use "electron" instead of "charge", pls.
Furthermore, I'd define the Fermi level is the _highest occupied level (wrt to some energy reference) at vanishing temperature_.

In the case of independent-fermions (read Fermi-Dirac distribution), the Fermi-level and chemical potential both coincide. This also the inflection point of the distribution, so something like _point of maximal energy gain_.

> It can be also defined as the chemical potential at zero temperature.

I think your comparison also properly includes correlated systems. Though, I would rephrase it in terms of "the zero-temperature limit".

src/nomad_simulations/properties/__init__.py
ndaelman-hu, line 1: At some point, it may just be better to include all sections at once.

src/nomad_simulations/outputs.py
```diff
@@ -65,8 +69,18 @@ class Outputs(ArchiveSection):
...
+    chemical_potential = SubSection(sub_section=ChemicalPotential.m_def, repeats=True)
+
     electronic_band_gap = SubSection(sub_section=ElectronicBandGap.m_def, repeats=True)
 
+    electronic_dos = SubSection(
+        sub_section=ElectronicDensityOfStates.m_def, repeats=True
+    )
+
+    xas_spectra = SubSection(sub_section=XASSpectra.m_def, repeats=True)
+
```
ndaelman-hu, line 75: So, just to make sure I'm following you:

- you are setting up a 2-tier system of properties: those that come from the schema and those that are manually added, but both follow the `PhysicalProperty` interface (note that Area A does use distinct interfaces in similar cases)?
- only, the former have to be added as a subsection each time, while the others are all listed in `custom_physical_property`?
- so is the 2-tier more like a 2nd-class citizenship, where `custom_physical_property` can only be used to further specify already defined properties?

ndaelman-hu: If I understood you well, this doesn't square with the pydocs of `Outputs`, since inheritance isn't being used to extend it.

Unless, by "extension" you mean _adding properties_? If so, consider my previous comment here. Also, I think that _schema inheritance_ is becoming confusing with _class inheritance_. For the latter, one would prefer dependency injection over inheritance, but that doesn't make sense for schemas.

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,601 @@
...
+        return True
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
+
+        if self.is_valid_spectral_profile() is False:
+            logger.error(
+                'Invalid negative intensities found: could not validate spectral profile.'
+            )
+            return
```
ndaelman-hu, line 88: I like this approach: leaving the data intact but throwing a warning. We probably should have users check the warnings more, or make them more visible.

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,601 @@
...
+    def __init__(
+        self, m_def: Section = None, m_context: Context = None, **kwargs
+    ) -> None:
+        super().__init__(m_def, m_context, **kwargs)
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
+
+        # TODO add normalization for `projected_dos` to extract `name` (if `m_parent` is `ElectronicDOS`
+        # TODO (cont'd) and `entity_ref` is set)
```
ndaelman-hu: I don't get the point of this class. Projected DOS should still have the same units.

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,601 @@
...
+    fermi_level = Quantity(
+        type=np.float64,
+        unit='joule',
+        description="""
+        The Fermi level is the highest occupied energy level at zero temperature. For insulators and semiconductors,
+        some codes have some difficulty in determining the Fermi level, and it is often set to the middle of the band gap,
+        at the top of the valence band, or even in the bottom of the conduction band. This quantity is extracted
+        from `FermiLevel.value` property.
+        """,
+    )
```
ndaelman-hu: What is the distinction in responsiblities between this and `FermiLevel`?

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,601 @@
...
+    """
+
+    iri = 'http://fairmat-nfdi.eu/taxonomy/ElectronicDensityOfStates'
+
+    spin_channel = Quantity(
+        type=np.int32,
+        description="""
+        Spin channel of the corresponding electronic DOS. It can take values of 0 or 1.
+        """,
+    )
```
ndaelman-hu, line 156: 1. `type=MEnum(0,1)`?
2. better still, define this quantity globally and just copy it in
3. best option, to include non-collinear, we generate a specialized section.

src/nomad_simulations/properties/spectral_profile.py
ndaelman-hu: > For insulators and semiconductors, some codes have some difficulty in determining the Fermi level, and it is often set to the middle of the band gap, [...]

This is actually the proper definition for anything with a band gap. The top of the valence band is wrong, i.e. too simplistic for these cases. A similar extension exists to account for externally applied voltages.

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,601 @@
...
+
+    energies_origin = Quantity(
+        type=np.float64,
+        unit='joule',
+        description="""
+        Energy level denoting the origin along the energy axis, used for comparison and visualization. It is
+        defined as the `ElectronicEigenvalues.highest_occupied` and does not necessarily coincide with `fermi_level`.
+        """,
+    )
```
ndaelman-hu: While I like the sentiment, this is not proper. Most (DFT) codes are extremely vague wrt their energy reference, so we have no other choice but to align wrt the Fermi level or highest-occupied orbital. A consistent choice would also make it easier to interpret and compare our automated plots.

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,601 @@
...
+
+    normalization_factor = Quantity(
+        type=np.float64,
+        description="""
+        Normalization factor for electronic DOS to get a cell-independent intensive DOS. The cell-independent
+        intensive DOS is as the integral from the lowest (most negative) energy to the `fermi_level` for a neutrally
+        charged system (i.e., the sum of `AtomsState.charge` is zero).
+        """,
+    )
```
ndaelman-hu: I say this as the person who was tasked implementing this:
the only real use here, is for DOS comparison metrics. Maybe it should be triggered in the analysis pipeline (though the functionality may remain here). You can double-check with Luca.

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,601 @@
...
+    #     """,
+    # )
+
+    projected_dos = SubSection(
+        sub_section=DOSProfile.m_def,
+        repeats=True,
+        description="""
+        Projected DOS. It can be atom- (different elements in the unit cell) or orbital-projected. These can be calculated in a cascade as:
+            - If the total DOS is not present, we can sum all atom-projected DOS to obtain it.
+            - If the atom-projected DOS is not present, we can sum all orbital-projected DOS to obtain it.
```
ndaelman-hu: At least acknowledge the potential lack in coverage depending on the projector functions used.

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,601 @@
...
+
+        In `projected_dos`, `name` and `entity_ref` must be set in order to normalization to work:
```
ndaelman-hu: typo: "in order for normalization to work"

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,601 @@
...
+            logger (BoundLogger): The logger to log messages.
+
+        Returns:
+            (Optional[pint.Quantity]): The resolved Fermi level.
+        """
+        fermi_level_value = self.fermi_level
+        if fermi_level_value is None:
+            fermi_level = get_sibling_section(
+                section=self, sibling_section_name='fermi_level', logger=logger
+            )  # we consider `index_sibling` to be 0
```
ndaelman-hu: What if multiple Fermi levels are extracted? I literally have such a case.

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,601 @@
...
+
+    def resolve_energies_origin(
+        self,
+        energies: pint.Quantity,
+        fermi_level: Optional[pint.Quantity],
+        logger: BoundLogger,
+    ) -> Optional[pint.Quantity]:
+        """
+        Resolve the origin of reference for the energies from the sibling `ElectronicEigenvalues` section and its
+        `highest_occupied` level, or if this does not exist, from the `fermi_level` value as extracted from `resolve_fermi_level()`.
```
ndaelman-hu: Given the Fermi level and a broad enough sampling of the DOS, one can always obtain the `highest_occupied_state`. The eigenvalues are only needed in case the Fermi level is missing.

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,601 @@
...
+        Extract the electronic band gap from the `highest_occupied_energy` and `lowest_occupied_energy` stored
+        in `m_cache` from `resolve_energies_origin()`. If the difference of `highest_occupied_energy` and
+        `lowest_occupied_energy` is negative, the band gap `value` is set to 0.0.
+
+        Returns:
+            (Optional[ElectronicBandGap]): The extracted electronic band gap section to be stored in `Outputs`.
+        """
+        band_gap = None
+        homo = self.m_cache.get('highest_occupied_energy')
+        lumo = self.m_cache.get('lowest_occupied_energy')
```
ndaelman-hu, line 393: Is there an issue with reporting these properties in the schema?

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,601 @@
...
+        fermi_level_value = self.fermi_level
+        if fermi_level_value is None:
+            fermi_level = get_sibling_section(
+                section=self, sibling_section_name='fermi_level', logger=logger
+            )  # we consider `index_sibling` to be 0
+            if fermi_level is not None:
+                fermi_level_value = fermi_level.value
+        return fermi_level_value
+
+    def resolve_energies_origin(
```
ndaelman-hu, line 207: I'm not implying that you badly migrated this code, but the algorithm wasn't all that great to start with and probably merits a double check.

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,601 @@
...
+    )
+
+    def __init__(
+        self, m_def: Section = None, m_context: Context = None, **kwargs
+    ) -> None:
+        super().__init__(m_def, m_context, **kwargs)
+        # Set the name of the section
+        self.name = self.m_def.name
+
+    # ? Maybe this method (and the one for ElectronicDensityOfStates.generate_from_projected_dos) should be moved to `PhysicalProperty`?
```
ndaelman-hu: I would define the interfaces `Contributor` with `def post_contribution` and `Composed` with `def accumulate_contribtions`, so the former calls the latter. Probably, the implementation can also be handled quite abstractly.

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,601 @@
...
+    def generate_from_contributions(self, logger: BoundLogger) -> None:
+        """
+        Generate the `value` of the XAS spectra by concatenating the XANES and EXAFS contributions. It also concatenates
+        the `Energy` grid points of the XANES and EXAFS parts.
```
ndaelman-hu, line 564: Not an expert, but is this generically true? Doesn't the concatenation depend on the exact energy ranges?

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,601 @@
...
+        super().__init__(m_def, m_context, **kwargs)
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
+
+        # TODO add normalization for `projected_dos` to extract `name` (if `m_parent` is `ElectronicDOS`
+        # TODO (cont'd) and `entity_ref` is set)
+
+
+class ElectronicDensityOfStates(DOSProfile):
```
ndaelman-hu, line 143: Given the complexity of the normalizers in this class, I think it is best to develop an electronic structure post-analysis suite, similar to MatID for symmetries. There already exist `PyProcar` and `ASH`, but they're fairly limited and code specific.

@JFRudzinski where would this fit into our planning? Note that **a proper package would be worthy of a publication**!

ndaelman-hu, line 143: Some codes use smearing to soften the DOS. Maybe include a quantity for it.

ndaelman-hu (commented):
Here's a 1st set of feedback. Revising the tests comes later.

src/nomad_simulations/properties/energies.py
JosePizarro3: `entity_ref` could be used for that.

src/nomad_simulations/properties/energies.py
JosePizarro3: 1. I will move `iri` and `rank` to `m_def = Section()` most probably, depending on Area D (Lauri) implementation
2. I checked, and in the taxonomy they appear as http. I will ask Sascha about it.

src/nomad_simulations/properties/energies.py
JosePizarro3: All these are descriptions extracted from the taxonomy by hand. I'd prefer to leave it as it is, and you are more than welcome to give feedback there. I will ask Pepe if it is ok to open this to more people.

src/nomad_simulations/properties/__init__.py
JosePizarro3: What do you mean? I want to be able to import properties by doing `from nomad_simulations.properties import ...`

src/nomad_simulations/outputs.py
JosePizarro3: I am not 100% sure of the `custom_physical_property` idea tbh. Might make more sense to simply delete and only allow for extending the schema.

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: I need this class indeed to have `value` set up with the proper `'1/joule'` units in `projected_dos`. Otherwise, I need to point to `SpectralProfile` and there `value` does not have units (hence `projected_dos.value` does not either).

It is a bit complicated but it is the only way of keeping these contributions with the proper units in their values.

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: > What is the distinction in responsiblities between this and FermiLevel?

None, actually, this quantity is extracted from `FermiLevel`. Maybe I should keep it simple and delete this and only use the extracted Fermi level as a `property` of this class.

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: MEnum does not work with integers. And yes, I agree we should have a specialized section for the spin information. Please, go to #22 and write it there too.

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: The consistent choice is precisely: use `highest_occupied`. I can add more information for us to distinguish if something was properly extracted or not, but for that I need to work out the `ElectronicEigenvalues` schema.

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: Can you elaborate? How do you have several Fermi levels, are they in terms of some variable?

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: We can talk about it. Similarly, we need to decide about the dos_integrator.py.

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: I think the Fermi level is quite obscure sometimes. I would actually go with this logic:
1. Check if `ElectronicEigenvalues.highest_occupied_energy` exists. If so, define the ref as this quantity.
2. If not, check the `FermiLevel.value`.

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: I think they must live in `ElectronicEigenvalues`, otherwise, they are irrelevant (and hence using m_cache).

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: Yeah, I simply migrated, and indeed, it should be checked once more. Again, I need some cases with `ElectronicEigenvalues`, so I will add a TODO here.

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: I am not 100% sure... I think it mostly depends on the experimental side. I will ask someone who knows more about it.

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: > I think it is best to develop an electronic structure post-analysis suite, similar to MatID for symmetries.

Yeah it might be, but we need the whole picture first. IMO, we could publish this repo as it is, and later think of some post-processing analysis tools which could be generalized. As of now, it is true that I am trying to keep as much as possible from the former normalization, but improving the schema.

JosePizarro3: I think `smearing` is a concept general enough to be under `PhysicalProperty`, same as `errors`.

src/nomad_simulations/properties/energies.py
JosePizarro3: About the https: this seems to be an old implementation by the time these ontologies started, so Sascha has no control over it.

src/nomad_simulations/outputs.py
JosePizarro3: Ok, I decided to delete it. It makes more bad than good.

JosePizarro3:
@ndaelman-hu thanks for the comments

Would you mind checking again? I can also discuss the details on implementation is something is unclear.

ndaelman-hu:
> @ndaelman-hu thanks for the comments
> 
> Would you mind checking again? I can also discuss the details on implementation is something is unclear.

Okay, will take a look somewhere this evening.

src/nomad_simulations/outputs.py
ndaelman-hu: Yeah, I think that's the best decision here.
The idea might be worthwhile to revisit in the future (I can see use cases), but we should 1st have a well-thought protocol for the standard schema.

src/nomad_simulations/properties/band_gap.py
```diff
@@ -31,6 +31,8 @@ class ElectronicBandGap(PhysicalProperty):
     Energy difference between the highest occupied electronic state and the lowest unoccupied electronic state.
     """
 
+    # ! implement `iri` and `rank` as part of `m_def = Section()`
+
```
ndaelman-hu, line 35: Yes, this sounds more appropriate.

src/nomad_simulations/properties/energies.py
ndaelman-hu: I see... Then pls pass along my feedback.
Is this the interoperability task force that defined it? I remember them wanting to poll the PIs, and later on include the domain experts. Shouldn't the PI feedback phase have ended by now?

You may resolve after reading this and adding a TODO comment.

src/nomad_simulations/properties/energies.py
ndaelman-hu: Ouch... If they have registered domain name they could get `https`, which is more modern and secure.
Anyhow, I only care for the link itself here.

tests/conftest.py
```diff
@@ -16,14 +16,25 @@
...
 from nomad.units import ureg
 
+from . import logger
+
+from nomad_simulations.model_system import ModelSystem, AtomicCell
+from nomad_simulations.atoms_state import AtomsState, OrbitalsState
 from nomad_simulations.outputs import Outputs, SCFOutputs
 from nomad_simulations.numerical_settings import SelfConsistency
-from nomad_simulations.properties import ElectronicBandGap
+from nomad_simulations.variables import Energy2 as Energy
```
ndaelman-hu, line 33: Why do we have an `Energy2`?

src/nomad_simulations/properties/energies.py
JosePizarro3: > Shouldn't the PI feedback phase have ended by now?

Ha ha, it barely started.

tests/test_energies.py
ndaelman-hu, line 1: Both tests are a bit silly, honestly. This is just proper initializer behavior.
I may be wrong, but for me, the point of unit testing is more so to (exhaustively) cover all input scenarios.
It's nice that you made it, but idt that we want to apply this to all cases going forward.

_Suggested actions_
I think they can safely be removed, especially in light of the suggested `iri` and `rank` changes.
Regarding `name`, I think this behavior is more general than just `FermiLevel` and `ChemcialPotential`, right? You should test only the common interface, if you really want to retain it.

tests/conftest.py
```diff
@@ -36,7 +47,42 @@ def pytest_internalerror(excinfo):
         raise excinfo.value
 
 
+def get_model_system(
+    type: str = 'original',
+    positions: List[List[float]] = [[0, 0, 0], [0.5, 0.5, 0.5]],
+    chemical_symbols: List[str] = ['Ga', 'As'],
+    orbital_symbols: List[List[str]] = [['s'], ['px', 'py']],
```
ndaelman-hu: PEP states that defaults should not have spaces around the assignment (`=`) operator.
Ruff should've caught this, though. Pls check that it's working fine.

ndaelman-hu: Tbh, I even think that the defaults are superfluous. Each test will have to specify their own systems of interest.

Actually, the idea behind this function is so nice, it shouldn't just be part of testing, but **become a utility for parser development**.

tests/conftest.py
```diff
@@ -36,7 +47,42 @@ def pytest_internalerror(excinfo):
...
+    Create a `ModelSystem` object with the given parameters.
+    """
+    model_system = ModelSystem()
+    atomic_cell = AtomicCell(type=type, positions=positions * ureg.meter)
+    model_system.cell.append(atomic_cell)
+
+    # Add atoms_state to the model_system
+    atoms_state = []
+    for index, atom in enumerate(chemical_symbols):
+        orbitals = orbital_symbols[index]
```
ndaelman-hu: You have 2 options here:
1. require `chemical_symbols` and `orbital_symbols` be of the same length, then you remove `index` (don't introduce superfluous variables)
```python
for orbitals, atom in zip(orbital_symbols, chemical_symbols)
```
2. accommodate for empty (`None`) positions in `orbital_symbols`, in which case you need extra checks

tests/conftest.py
ndaelman-hu: I also am in the habit of using plural for lists or tuples of data. Given that this is a list of lists, I'd use a double plural, i.e. *orbital**s**_symbols*.

Actually, `symbols` is redundant: you even drop it in your `for`-loop later on.
Maybe use `atoms_orbitals`, which contains `atom_orbitals` that in turn contain `atom_orbital`?

(Btw, this is just a suggestion. I'm not going to argue about variable names, they just have to be understandable enough. In that sense, this here suffices.)

tests/conftest.py
```diff
@@ -36,7 +47,42 @@ def pytest_internalerror(excinfo):
...
+        orbitals_state = []
+        for orbital in orbitals:
+            orbitals_state.append(
+                OrbitalsState(
+                    l_quantum_symbol=orbital[0], ml_quantum_symbol=orbital[1:]
```
ndaelman-hu, line 94: This indicates that we want a setter that automatically splits (and converts) states to l and m_l.

tests/conftest.py
```diff
@@ -36,7 +47,42 @@ def pytest_internalerror(excinfo):
...
+                )
+            )
+        atom_state = AtomsState(chemical_symbol=atom, orbitals_state=orbitals_state)
+        # and obtain the atomic number for each AtomsState
+        atom_state.resolve_chemical_symbol_and_number(logger)
```
ndaelman-hu: If we have these functions return `self`, you could just chain both together.

tests/conftest.py
```diff
@@ -36,7 +47,42 @@ def pytest_internalerror(excinfo):
...
+        atoms_state.append(atom_state)
+    atomic_cell.atoms_state = atoms_state
+    return model_system
+
+
 def get_scf_electronic_band_gap_template(threshold_change: float = 1e-3) -> SCFOutputs:
```
ndaelman-hu: Just a minor comment: I'd use `get`/`set` rather for retrieving/storing data from a state.
Here, you're more so _generating_, _creating_, or _setting up_ from scratch.

tests/conftest.py
```diff
@@ -57,6 +103,54 @@ def get_scf_electronic_band_gap_template(threshold_change: float = 1e-3) -> SCFO
...
+        entity_ref=model_system.cell[0].atoms_state[1].orbitals_state[1],
+    )
+    orbital_s_Ga_pdos.value = [0.2, 0.5, 0, 0, 0, 0.0, 0.0] * ureg('1/joule')
+    orbital_px_As_pdos.value = [1.0, 0.2, 0, 0, 0, 0.3, 0.0] * ureg('1/joule')
+    orbital_py_As_pdos.value = [0.3, 0.5, 0, 0, 0, 0.5, 1.3] * ureg('1/joule')
+    electronic_dos.projected_dos = [
+        orbital_s_Ga_pdos,
+        orbital_px_As_pdos,
+        orbital_py_As_pdos,
+    ]
```
ndaelman-hu: This is a shame... I think one could set this up based on a model system input, together with a partial dos.

tests/conftest.py
```diff
@@ -57,6 +103,54 @@ def get_scf_electronic_band_gap_template(threshold_change: float = 1e-3) -> SCFO
...
+
+
 @pytest.fixture(scope='session')
 def scf_electronic_band_gap() -> SCFOutputs:
     return get_scf_electronic_band_gap_template()
+
+
+@pytest.fixture(scope='session')
+def electronic_dos() -> ElectronicDensityOfStates:
+    return get_electronic_dos()
```
ndaelman-hu: If these are functionalities for producing tests (they allow for systems to be defined on-the-fly), they should be turned into fixtures in the respective test files.

tests/conftest.py
ndaelman-hu, line 1: My main feedback on `get_model_system`, and how _it should be a utility or setter of the respective class_, also applies to other functions.

tests/conftest.py
```diff
@@ -36,7 +47,42 @@ def pytest_internalerror(excinfo):
...
+    type: str = 'original',
+    positions: List[List[float]] = [[0, 0, 0], [0.5, 0.5, 0.5]],
+    chemical_symbols: List[str] = ['Ga', 'As'],
+    orbital_symbols: List[List[str]] = [['s'], ['px', 'py']],
+) -> ModelSystem:
+    """
+    Create a `ModelSystem` object with the given parameters.
+    """
+    model_system = ModelSystem()
+    atomic_cell = AtomicCell(type=type, positions=positions * ureg.meter)
```
ndaelman-hu, line 84: If you want to avoid conversions and stay with the default, you can leave out `ureg.meter`. It even comes across as less silly, i.e. a meter spacing between atoms :p

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,256 @@
...
+    """
+    Test the `SpectralProfile` class defined in `properties/spectral_profile.py`.
+    """
+
+    def test_is_valid_spectral_profile(self):
+        """
+        Test the `is_valid_spectral_profile` method.
+        """
+        spectral_profile = SpectralProfile(
+            variables=[Energy(grid_points=[-3, -2, -1, 0, 1, 2, 3] * ureg.joule)]
```
ndaelman-hu: If you want to avoid conversions and stay with the default, you can leave out `ureg.joule`.

ndaelman-hu:
```python
grid_points=list(range(-3, 4))
```

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,256 @@
...
+        )
+        spectral_profile.value = [1.5, 1.2, 0, 0, 0, 0.8, 1.3]
+        assert spectral_profile.is_valid_spectral_profile()
+        spectral_profile.value = [3, 2, 0, 0, 0, -4, 1]
+        assert not spectral_profile.is_valid_spectral_profile()
+        # default value inherited in other SpectralProfile classes
+        assert spectral_profile.rank == []
```
ndaelman-hu: can be removed

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,256 @@
...
+        """
+        Test the `is_valid_spectral_profile` method.
+        """
+        spectral_profile = SpectralProfile(
+            variables=[Energy(grid_points=[-3, -2, -1, 0, 1, 2, 3] * ureg.joule)]
+        )
+        spectral_profile.value = [1.5, 1.2, 0, 0, 0, 0.8, 1.3]
+        assert spectral_profile.is_valid_spectral_profile()
+        spectral_profile.value = [3, 2, 0, 0, 0, -4, 1]
+        assert not spectral_profile.is_valid_spectral_profile()
```
ndaelman-hu: One is invalid because of the negative argument, right?
You can emphasize this better by using only 1 or 2 list elements.

src/nomad_simulations/properties/spectral_profile.py
ndaelman-hu: I see. Still, the DOS does not require any special treatment. You should rather rethink how you handle spectra in general: in the end, all spectra lie on an energy or frequency axis (with a simple inter-conversion), right? Their amplitudes are completely defined by that axis and what they integrate to. Idk if there's a generic dimensionality for the latter, maybe _quasi-particles_?

Most scientists don't bother labelling the amplitude axis, since they only care for relative heights inside a single sample. Likewise, most DOS is also reported in a.u. This approach just prevents _inter-sample comparison_, which might be wishful.

_Implementation suggestion_
If `Quantity` cannot accommodate your needs, go with a dynamic check, as you've done before. Or you forego the check, if you can't formulate one.

src/nomad_simulations/properties/spectral_profile.py
ndaelman-hu: Okay, the standalone module is more an idea for down the line (@JFRudzinski would also like your feedback) and a way to get an easy paper. I'm not suggesting this for now! Still, it feels a bit out of place compared to the simpler, and more to-the-point checks that make up most of the other normalizers.

What is relevant at the moment, though, is _stress testing these algos_. We have numerous cases in our repo, where the band structure or DOS extraction fail, and it's painful to see. This has always been a very tricky issue.
Since this PR focuses on migrating and extending the schema, **I implore you to open a new issue where we tackle this aspect**. In the meantime, I'll think up more test cases.

src/nomad_simulations/properties/spectral_profile.py
ndaelman-hu: > MEnum does not work with integers.

I'll open up an issue.

src/nomad_simulations/properties/spectral_profile.py
ndaelman-hu: Then I leave it up to them to resolve, or you add TODO.

ndaelman-hu (approved):
Almost there. Mostly small stuff, and a bit of reshuffling of where code belongs.

I still need to take a look at larger tests in `test_spectral_profile`, but I'll do that after a break.

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,256 @@
...
+        """
+        Test the default quantities assigned when creating an instance of the `ElectronicDensityOfStates` class.
+        """
+        electronic_dos = ElectronicDensityOfStates()
+        assert (
+            electronic_dos.iri
+            == 'http://fairmat-nfdi.eu/taxonomy/ElectronicDensityOfStates'
+        )
+        assert electronic_dos.name == 'ElectronicDensityOfStates'
+        assert electronic_dos.rank == []
```
ndaelman-hu, line 73: Similar to my previous comments

tests/test_spectral_profile.py
ndaelman-hu, line 1: I'm not sure why you stopped using `pytest.mark.parametrize`.
I think that plenty of these tests lend themselves to it, and it makes everything much clearer to read.

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def test_check_energy_variables(self):
+        """
+        Test the `_check_energy_variables` method.
+        """
+        electronic_dos = ElectronicDensityOfStates()
+        electronic_dos.variables = [
+            Temperature(grid_points=[-3, -2, -1, 0, 1, 2, 3] * ureg.kelvin)
```
ndaelman-hu: just store the grid points in a single variable and condense the definition as shown before

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,580 @@
...
+        """
+        Check if the required `Energy` variable is present in the `variables`. If so, it returns
+        the grid points of the `Energy` variable.
+
+        Args:
+            logger (BoundLogger): The logger to log messages.
+
+        Returns:
+            (Optional[pint.Quantity]): The grid points of the `Energy` variable.
+        """
```
ndaelman-hu: On naming: `check` functions sound like it should return `bool`.
If you prefer to retain this type signature (depends on the use), call it a `get`ter.

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,580 @@
...
+        for var in self.variables:
+            if isinstance(var, Energy):
+                return var.grid_points
```
ndaelman-hu: Why return only the `grid_points`? I'd return the object as a whole...

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,256 @@
...
+        )
+        energies = electronic_dos._check_energy_variables(logger)
+        assert (energies.magnitude == np.array([-3, -2, -1, 0, 1, 2, 3])).all()
+
+    def test_resolve_energies_origin(self):
+        """
+        Test the `resolve_energies_origin` method.
+        """
+        # ! add test when `ElectronicEigenvalues` is implemented
+        assert True
```
ndaelman-hu: `pass`?

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,256 @@
...
+        for atom in atoms_state:
+            atom.resolve_chemical_symbol_and_number(logger)
+        atomic_cell.atoms_state = atoms_state
+        # Non spin-polarized
+        normalization_factor = electronic_dos.resolve_normalization_factor(logger)
+        assert np.isclose(normalization_factor, 0.015625)
+        # Spin-polarized
+        electronic_dos.spin_channel = 0
+        normalization_factor = electronic_dos.resolve_normalization_factor(logger)
+        assert np.isclose(normalization_factor, 0.0078125)
```
ndaelman-hu: would be nice to highlight that one is the half of the other

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,256 @@
...
+        # assert (
+        #     orbital_projected[1].entity_ref
+        #     == outputs.model_system_ref.cell[0].atoms_state[1].orbitals_state[0]
+        # )  # orbital `px` in `As` atom
+        # assert (
+        #     orbital_projected[1].entity_ref
+        #     == outputs.model_system_ref.cell[0].atoms_state[1].orbitals_state[1]
+        # )  # orbital `py` in `As` atom
+
+        # Note: `val` is reported from `self.value`, not from the extraction
```
ndaelman-hu: Well... this has to be solved first.

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,256 @@
...
+        assert atom_projected_names == ['atom Ga', 'atom As']
+        # ! These tests are not passing, despite these are the same sections
+        # assert (
+        #     atom_projected[0].entity_ref
+        #     == outputs.model_system_ref.cell[0].atoms_state[0]
+        # )  # `Ga` atom
+        # assert (
+        #     atom_projected[1].entity_ref
+        #     == outputs.model_system_ref.cell[0].atoms_state[1]
+        # )  # `As` atom
```
ndaelman-hu: Same here

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,256 @@
...
+        """
+        electronic_dos = ElectronicDensityOfStates()
+        assert (
+            electronic_dos.iri
+            == 'http://fairmat-nfdi.eu/taxonomy/ElectronicDensityOfStates'
+        )
+        assert electronic_dos.name == 'ElectronicDensityOfStates'
+        assert electronic_dos.rank == []
+
+    def test_check_energy_variables(self):
```
ndaelman-hu: Additional edge cases to test for:
- child classes from `Energy`
- non-`Energy` objects with energy units.

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,256 @@
...
+        assert len(electronic_dos.projected_dos) == 3  # only orbital projected DOS
+        orbital_projected = electronic_dos.extract_projected_dos('orbital', logger)
+        atom_projected = electronic_dos.extract_projected_dos('atom', logger)
+        assert len(orbital_projected) == 3 and len(atom_projected) == 0
+        orbital_projected_names = [orb_pdos.name for orb_pdos in orbital_projected]
+        assert orbital_projected_names == [
+            'orbital s Ga',
+            'orbital px As',
+            'orbital py As',
+        ]
```
ndaelman-hu, line 184: ... this `name` attribute is useless...

ndaelman-hu: I guess this is just a shorthand for checking multiple attributes. Still useless

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,256 @@
...
+        #     == outputs.model_system_ref.cell[0].atoms_state[1].orbitals_state[0]
+        # )  # orbital `px` in `As` atom
+        # assert (
+        #     orbital_projected[1].entity_ref
+        #     == outputs.model_system_ref.cell[0].atoms_state[1].orbitals_state[1]
+        # )  # orbital `py` in `As` atom
+
+        # Note: `val` is reported from `self.value`, not from the extraction
+        val = electronic_dos.generate_from_projected_dos(logger)
+        assert (val.magnitude == electronic_dos.value.magnitude).all()
```
ndaelman-hu: Move this out into a separate test.

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,256 @@
...
+    Test the `XASSpectra` class defined in `properties/spectral_profile.py`.
+    """
+
+    # ! Include this initial `test_default_quantities` method when testing your PhysicalProperty classes
+    def test_default_quantities(self):
+        """
+        Test the default quantities assigned when creating an instance of the `XASSpectra` class.
+        """
+        xas_spectra = XASSpectra()
+        assert xas_spectra.iri is None  # Add iri when available
```
ndaelman-hu, line 264: Or better still, remove this test.

tests/test_spectral_profile.py
ndaelman-hu: Btw, do you check the following for `PhysicalProperty`:
- what if there's only 1 `variables`?
- what if both `variables` have different lengths?
- what if `variables` mix up data types?

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,256 @@
...
+        assert len(xas_spectra.variables) == 1
+        assert len(xas_spectra.variables[0].grid_points) == 6
+        assert len(xas_spectra.variables[0].grid_points) == (
+            len(xanes_spectra.variables[0].grid_points)
+            + len(exafs_spectra.variables[0].grid_points)
+        )
+        assert (
+            xas_spectra.variables[0].grid_points.magnitude == [0, 1, 2, 3, 4, 5]
+        ).all()
+        assert (xas_spectra.value == [0.5, 0.1, 0.3, 0.2, 0.4, 0.6]).all()
```
ndaelman-hu: Also test for:
- only 1 of either being present
- multiple xas or xanes entries
- non-matching grid points
- irregular grid points

ndaelman-hu (changes requested):
So, I brought up some additional edge cases to test for.
The failing tests are what worry me most, though. Since it is all related to the pDOS, it could be tackled in a separate issue, but that has to be high on the priority list.

Regarding DOS in general, it really doesn't merit a fundamentally different treatment from other spectra. I think we can have a single class where `name` is used to specify the profile type.

tests/test_spectral_profile.py
JosePizarro3: This is not a bad behavior, it is actually expected (because I am passing the fixture). I could set the testing such that it is extracted from the projected values, let me check.

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: It is not like that: `projected_dos` is a list of elements which can contain atom-pdos or orbital-pdos, something like:
```python
projected_dos = [
   atom Ga:DOSProfile(name='atom Ga', value=...), 
   orbital px Ga:DOSProfile(name='orbital px Ga', value=...), 
   orbital py Ga:DOSProfile(name='orbital py Ga', value=...),
]
```

Hence the special requirement of looping and getting one name or another matched (see `extract_projected_dos(type, logger)`)

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: I deleted it and keep it separately in its own section.

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: @ndaelman-hu do you have any example where this happens?

tests/conftest.py
JosePizarro3: There is a bug when defining things with the same names. The `DOSNormalizer` is getting the `run.calculation.dos_electronic` section, but for some reason when normalizing, it is getting this `data.outputs.dos_electronic.energy` as the energies... I will open an issue for Theodore once this is merged.

tests/test_energies.py
JosePizarro3: I would keep this default quantities, probably changing it once Area D implement this m_def for them. So they stay for now, but indeed, the tests are just checks to see whether the `iri`, `rank` and `name` are assigned properly.

tests/conftest.py
JosePizarro3: Well here I am exclusively testing the schema, so I think this conftest lives good here.

JosePizarro3: Regarding PEP: https://peps.python.org/pep-0008/#other-recommendations

So it is good like this when assigning the type.

tests/conftest.py
JosePizarro3: Yes, I have the other branch with unit testing, I can add it there and change this.

tests/conftest.py
JosePizarro3: Indeed, thanks. Done!

tests/conftest.py
JosePizarro3: Yeah, changed it for `generate_`.

tests/conftest.py
JosePizarro3: Well, `entity_ref` works like that. The name was redundant, but I should add a method to extract the name (like I add as a TODO). Let me do it then.

tests/conftest.py
JosePizarro3: Ok, problem is some of these fixtures might be used in other test files, creating the need of generating the fixture in each of them.

tests/conftest.py
JosePizarro3: It just helps understanding the units of these quantities, so I prefer to leave it (same for the others).

tests/test_spectral_profile.py
JosePizarro3: Some of them yes, but sometimes I just want to test a very specific use case of a function.

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: I am only using the points, but yes, it might be easier to return the object.

JosePizarro3: Or change the name to `_get_energy_points`.

JosePizarro3:
> So, I brought up some additional edge cases to test for. The failing tests are what worry me most, though. Since it is all related to the pDOS, it could be tackled in a separate issue, but that has to be high on the priority list.
> 
> Regarding DOS in general, it really doesn't merit a fundamentally different treatment from other spectra. I think we can have a single class where `name` is used to specify the profile type.

Ok, tricky error in the conftest for fixing the testing for pDOS: I need to create a full `Simulation` object, and not simply let `ModelSystem` hang around. I will push something today so you can check the new `generate_simulation_electronic_dos` conftests (which returns the full `Simulation`, and not the `ElectronicDensityOfStates` section).

tests/test_spectral_profile.py
JosePizarro3: Ok fixed, please check the new testing with fixtures. I separated this testing (one for `extract_projected_dos` and another for `generate_from_pdos`).

I am resolving all other threads related with this.

src/nomad_simulations/properties/spectral_profile.py
ndaelman-hu: But this is a different consideration than the one you mentioned before regarding the units:
> I need this class indeed to have value set up with the proper '1/joule' units in projected_dos

All I'm saying is that you could set generic units (e.g. `qp_mass/joule`) for spectra, no need for `DOSProfile`. You can determine the amplitude units from the energy axis (converting frequency axes) and the no. QPs.

I guess all spectra should be able to contain contributions of other spectra.
These can then either be used to construct those, as with the pDOS, or concatenated, as with XAS and XANES.
These specific functionalities (interfaces) could either both be specified at `Spectra` or (maybe better) in the child classes representing the 3 techniques.

JosePizarro3: Ok, I am resolving, but we can revisit the units in a future.

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: Resolving after our internal discussion.

tests/test_spectral_profile.py
JosePizarro3: > what if there's only 1 variables?
  what if both variables have different lengths?
  what if variables mix up data types?

Done in `test_static_properties` in `test_physical_properties`. The last case does not make sense, as `PhysicalProperty` only care about lengths of `variables.grid_points`, and not their type. Nothing prevents you from having different types.

tests/test_spectral_profile.py
JosePizarro3: Ok, I added better functionalities in the `XASSpectra.generate_from_contributions` thanks to the test.

JosePizarro3:
Ok, finished. @ndaelman-hu do you mind checking before going on holidays? Thanks!

src/nomad_simulations/properties/spectral_profile.py
ndaelman-hu: Yeah, I saw it later on. In the old schema, `energies_origin` (or a similar naming lie `energy_reference`) referred more so to the vacuum alignment. Hence my confusion.

src/nomad_simulations/properties/spectral_profile.py
ndaelman-hu: I guess this comes down to the units of the amplitude. Did you end up deciding anything there?
If not, just add a comment for future references and we leave it for now.

src/nomad_simulations/properties/spectral_profile.py
ndaelman-hu: > I think the Fermi level is quite obscure sometimes.

I agree. This is fine.

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: For now, we keep it like this, but the units have to defined differently in DOS and spectras, so anyways. I will add a note to check again during the workshop.

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,303 @@
...
+
+    def test_get_energy_points(self):
+        """
+        Test the `_get_energy_points` method.
+        """
+        electronic_dos = ElectronicDensityOfStates()
+        electronic_dos.variables = [
+            Temperature(grid_points=list(range(-3, 4)) * ureg.kelvin)
+        ]
+        assert electronic_dos._get_energy_points(logger) is None
```
ndaelman-hu, line 83: `pylint` seems to prefer that protected functions are not called outside the class.
Good to know, but if this is an exception to the typical use, it's fine. Else, remove the underscore.

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,303 @@
...
+        electronic_dos.variables.append(
+            Energy(grid_points=list(range(-3, 4)) * ureg.joule)
+        )
+        energies = electronic_dos._get_energy_points(logger)
+        assert (energies.magnitude == np.array([-3, -2, -1, 0, 1, 2, 3])).all()
```
ndaelman-hu, line 88: mimnor: you can also contract the list here

tests/test_spectral_profile.py
ndaelman-hu: So, as discussed, this is for representation in the GUI.
Here, it used as a shorthand, but fine...

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,303 @@
...
+        simulation.model_system.append(model_system)
+        outputs.model_system_ref = simulation.model_system[0]
+        assert electronic_dos.resolve_normalization_factor(logger) is None
+
+        # No `model_system_ref.cell.atoms_state`
+        atomic_cell = AtomicCell(
+            type='original', positions=[[0, 0, 0], [0.5, 0.5, 0.5]] * ureg.meter
+        )
+        model_system.cell.append(atomic_cell)
+        assert electronic_dos.resolve_normalization_factor(logger) is None
```
ndaelman-hu, line 123: This more so a general remark about the NOMAD schema than an actionable piece of feedback to you:

a lot of sophisticated state checking is going here, to verify every step along the pipeline. This, of course, bloats the rest of the test logic.
In my branch on computing distances and angles, I'm trying out alternatives to _state_ to reduce the need such kind of checks.
Just keep this in mind when reviewing there. Nothing has to change here now.

src/nomad_simulations/properties/spectral_profile.py
```diff
@@ -0,0 +1,604 @@
...
+            eigenvalues.lowest_occupied if eigenvalues is not None else None
+        )
+        # and set defaults for `highest_occupied_energy` and `lowest_occupied_energy` in `m_cache`
+        if highest_occupied_energy is not None:
+            self.m_cache['highest_occupied_energy'] = highest_occupied_energy
+        if lowest_occupied_energy is not None:
+            self.m_cache['lowest_occupied_energy'] = lowest_occupied_energy
+
+        # Set thresholds for the energies and values
+        energy_threshold = config.normalize.band_structure_energy_tolerance
```
ndaelman-hu, line 254: My IDE is telling me that _`config` has no member `normalize`_. Are you using a different branch then `develop` for your base `nomad` installation?

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,303 @@
...
+        atom_projected = electronic_dos.extract_projected_dos('atom', logger)
+        assert len(orbital_projected) == 3 and len(atom_projected) == 0
+
+    @pytest.mark.parametrize(
+        'value, result',
+        [
+            (None, [1.5, 1.2, 0, 0, 0, 0.8, 1.3]),
+            ([1.5, 1.2, 0, 0, 0, 0.8, 1.3], [1.5, 1.2, 0, 0, 0, 0.8, 1.3]),
+            ([30.5, 1.2, 0, 0, 0, 0.8, 1.3], [30.5, 1.2, 0, 0, 0, 0.8, 1.3]),
+        ],
```
ndaelman-hu: What is the aim of the middle test? The first / last one checks for a DOS built up from pDOS / fully determined by itself.

tests/test_spectral_profile.py
```diff
@@ -0,0 +1,303 @@
...
+    ):
+        """
+        Test the `generate_from_contributions` method.
+        """
+        # ! extend this test
+        xas_spectra = XASSpectra()
+        if xanes_energies is not None:
+            xanes_spectra = SpectralProfile()
+            xanes_spectra.variables = [Energy(grid_points=xanes_energies * ureg.joule)]
+            xanes_spectra.value = [0.5, 0.1, 0.3]
```
ndaelman-hu, line 292: Woudln't this also be more appropriate as an input parameter?

tests/test_spectral_profile.py
ndaelman-hu, line 1: I like the improvements! The added comments are helpful.
Do note that to understand some of these tests, I had to search back up to the `conftest` or the method implementation to get the full context.
Some other time, we should think on how to document tests so they can be understood more standalone.

tests/conftest.py
```diff
@@ -57,6 +129,56 @@ def get_scf_electronic_band_gap_template(threshold_change: float = 1e-3) -> SCFO
...
+        variables=variables_energy,
+        entity_ref=model_system.cell[0].atoms_state[1].orbitals_state[0],
+    )
+    orbital_py_As_pdos = DOSProfile(
+        variables=variables_energy,
+        entity_ref=model_system.cell[0].atoms_state[1].orbitals_state[1],
+    )
+    orbital_s_Ga_pdos.value = [0.2, 0.5, 0, 0, 0, 0.0, 0.0] * ureg('1/joule')
+    orbital_px_As_pdos.value = [1.0, 0.2, 0, 0, 0, 0.3, 0.0] * ureg('1/joule')
+    orbital_py_As_pdos.value = [0.3, 0.5, 0, 0, 0, 0.5, 1.3] * ureg('1/joule')
```
ndaelman-hu, line 162: minor: Any reason why these aren't set upon object instantiation?

tests/conftest.py
```diff
@@ -36,7 +48,67 @@ def pytest_internalerror(excinfo):
...
 
-def get_scf_electronic_band_gap_template(threshold_change: float = 1e-3) -> SCFOutputs:
+def generate_simulation(
+    model_system: Optional[ModelSystem] = None,
+    model_method: Optional[ModelMethod] = None,
+    outputs: Optional[Outputs] = None,
+) -> Simulation:
+    """
+    Generate a `Simulation` section with the main sub-sections, `ModelSystem`, `ModelMethod`, and `Outputs`. If `ModelSystem`
+    and `Outputs` is not None, then it adds `ModelSystem` as a reference in `Outputs`.
```
ndaelman-hu: `None` instead of None

Tbh, you can also go for a more natural-language way of formulating, such as "[...] is set".

tests/conftest.py
```diff
@@ -36,7 +48,67 @@ def pytest_internalerror(excinfo):
...
+    simulation = Simulation()
+    if model_method is not None:
+        simulation.model_method.append(model_method)
+    if model_system is not None:
+        simulation.model_system.append(model_system)
+    if outputs is not None:
+        simulation.outputs.append(outputs)
+        if model_system is not None:
+            outputs.model_system_ref = model_system
+    return simulation
```
ndaelman-hu: Is the `None` check necessary? NOMAD should be able to handle that.

ndaelman-hu (approved):
All good, thank you very much!!

tests/test_spectral_profile.py
JosePizarro3: I want to leave it clear what the array is on the == check

tests/test_spectral_profile.py
JosePizarro3: I leave the privacy of the method intact, but indeed, I am testing the functionality in this test.

src/nomad_simulations/properties/spectral_profile.py
JosePizarro3: I am using a branch where I added this repo as a submodule. I saw the config and models/config + check locally and I can use the normalize no problem.

tests/test_spectral_profile.py
JosePizarro3: Indeed, it is pretty dumb. I deleted it.

tests/test_spectral_profile.py
JosePizarro3: Well, for this testing it is more interesting to keep the Energy axis properly defined. Whatever is in value does not matter too much.

tests/conftest.py
JosePizarro3: Just simplicity. I think this are not quantities that will change for our tests.

tests/conftest.py
JosePizarro3: Indeed, I've deleted it.

JosePizarro3:
Thanks @ndaelman-hu ! 

I am merging this and continue working on #52

PR #58, opened 2024-04-28
Extend model system

ndaelman-hu:
This PR adds inter-atomic distances and angles, with optional extension to dihedral angles, to the NOMAD schema.

---

# Schema
These distributions are stored in `AtomicCell.geometry_distributions: list[GeometryDistribution]` as a repeating subsection. Information on the (cutoffs, used to set up the) bins, is stored under `AtomicCell.geometry_analysis_cutoffs`.

`GeometryDistribution` objects are effectively histograms by element pair / triple. This is a choice to facilitate searches and visualizations over the distribution themselves, which are normalized to reproduce the same frequency for primitive cells as supercells. Additional advantages include a limited storage consumption. It is, however, not suitable for extracting the exact atomic pairs / triples matching a given distance / angle.

Note: triples / quadruples are defined around a specific geometric primitive (point / line piece). Their notation accounts for this, by placing them up front in the elements list.

# Additional functionalities
To keep state within `AtomicCell` simple and exert control over the stages in calculation, pure Python helper classes with the following functions are added, in line with the _Single Responsibility Principle_:

- `DistributionFactory` scans the combinatorial space of elements and generates their `Distribution`s.
- `Distribution` leverages `ase` for computing the distances / angles for the elemental combo provided. It can also produce a `DistributionHistogram` of itself.
- `DistributionHistogram` contain the histogram version of their combo distribution.

While only `DistributionHistogram` will eventually be written out to `GeometryDistribution`, the rest of the data artifacts are retained under `AtomicCell` during run-time. As such, once computed, they can be used in other computation, or data analysis (conversion to `pandas.DataFrame` in a separate PR).

Inclusion of the actual computation pipeline, meanwhile, is toggled via `AtomicCell.analyze_geometry: bool`.
The execution is handled by the `normalizer`.

# Miscellaneous

To facilitate a cleaner check for missing NOMAD attributes, i.e. quantities, subsections, I added the `check_attributes` decorator. It is used in `AtomicCell.set_ase_atoms`, as an example. If accepted, I recommend that it be reallocated to `utils.py`.

To support a more functional pipeline notation, next to the current imperative style, I make all computational functions and `normalize` return `self`. I checked with Area D that the latter does not violate the `normalize` interface specifications.

---
P.s.: I'm considering adding this PR description as documentation. Let me know your thoughts.


JosePizarro3:
> To support a more functional pipeline notation, next to the current imperative style, I make all computational functions and normalize return self. I checked with Area D that the latter does not violate the normalize interface specifications.

I doubt it does not break anything, especially the `MetainfoNormalizer`. And if so, I wonder whether you are willing to change ALL `normalize()` functions to now return self. In any case, I think this is not a good practice here: the only reason I'd see to return self is to chain methods one after another, i.e., if you want to do something like:
```python
class Example:

    def method_a(self):
         ...
         return self

    def method_b(self):
         ...
         return self

    def method_c(self):
         ...
         return self


example = Example()
example_2 = example.method_c().method_a().method_c().method_c()
```

Did you check that the methods (besides normalize, in which I won't do it) you implemented this return, this kind of logic is what we want?

src/nomad_simulations/model_system.py
```diff
@@ -317,59 +562,79 @@ class AtomicCell(Cell):
...
-    def __init__(self, m_def: Section = None, m_context: Context = None, **kwargs):
-        super().__init__(m_def, m_context, **kwargs)
-        # Set the name of the section
-        self.name = self.m_def.name
+    geometry_analysis_cutoffs = Quantity(
+        type=np.float64,
+        shape=['*'],
+        description="""
+        Cutoff radius within which atomic geometries are analyzed.
+        Defaults to 3x the covalent radius denoted by the [`ase` package](https://wiki.fysik.dtu.dk/ase/ase/neighborlist.html).
```
JFRudzinski: What happens if the atom_labels are `X` and no cutoff is explicitly specified?

src/nomad_simulations/model_system.py
```diff
@@ -317,59 +562,79 @@ class AtomicCell(Cell):
...
+            self.simple_distributions = DistributionFactory(
+                self.ase_atoms,
+                self.neighbor_list,
+            ).produce_distributions()
+
+            self.histogram_distributions, self.distributions = [], []
+            for distribution in self.simple_distributions:
+                if distribution.type == 'distances':
+                    bins = np.arange(
+                        np.arange(0, max(self.geometry_analysis_cutoffs), 0.001)
```
JFRudzinski: The bin spacing (i.e., 0.001 in this case) should also be giving as a parameter, similar to the cutoff. Imagine you have a system where the VDW radii are much larger than the atomic scale, then your cutoff should be much larger, and you will want to also expand your spacing. 

I would probably specify a default number of bins, rather than a bin spacing, and then generate the spacing based on the number of bins. But then allow for someone to specify either the number of bins or the bin spacing within atomic cell.

src/nomad_simulations/model_system.py
```diff
@@ -317,59 +562,79 @@ class AtomicCell(Cell):
...
+                    )
+                elif distribution.type == 'angles':
+                    bins = np.arange(0, 180, 0.01)
```
JFRudzinski: same as above, although it's not as big of a deal in this case. Still you may want to adjust the smoothness of your distribution, I would still apply the same parameters as for distances for generality of usage.

src/nomad_simulations/model_system.py
```diff
@@ -283,12 +318,216 @@ def normalize(self, archive, logger) -> None:
...
+    """Convert a combo tuple to a hypen-separated string.
+    For angles, the center element is now indeed placed at the center."""
+    if 1 < len(combo) < 4:
+        if len(combo) == 3:
+            combo = (combo[1], combo[0], combo[2])  # flip order
+        return '-'.join(combo)
+    raise ValueError('Only pairs and triplets are supported.')
+
+
+class GeometryDistribution(ArchiveSection):
```
JFRudzinski, line 293: I think we will want to change the structure here a little:

Instead of a repeating subsection of the same class type, I think this might need to be changed to a container section which may hold any geometrical representation. Then, individual types of representations can be specified as subsections. These subsections should then inherit from PhysicalProperty for consistency purposes (as we talked about, there is a gray area between storing certain representations under ModelSystem or ModelOutputs). This will also help general consistency of naming and such, and will allow you to deal with bin units individually for the distinct types of distributions.

src/nomad_simulations/model_system.py
```diff
@@ -283,12 +318,216 @@ def normalize(self, archive, logger) -> None:
...
+    )
+
+    n_bins = Quantity(
+        type=np.int32,
+        description="""
+        Number of bins used to create the histogram.
+        """,
+    )
+
+    bins = Quantity(
```
JFRudzinski, line 348: units?!

src/nomad_simulations/model_system.py
```diff
@@ -283,12 +318,216 @@ def normalize(self, archive, logger) -> None:
...
+    raise ValueError('Only pairs and triplets are supported.')
+
+
+class GeometryDistribution(ArchiveSection):
+    """
+    A base section used to specify the geometry distributions of the system.
+    These include frequency counts of inter-atomic distances and angles.
+    """
+
+    name = Quantity(
```
JFRudzinski, line 299: I think this could be more clear what you expect to be stored here. It sounds like you want to store a str with all element names involved in the calculation. Why not a list of strings? And why not specify (in the description) atom_labels if you expect that these are linked to the atom_labels stored under Atoms()? 

This might change if you end up inheriting from PhysicalProperty, which will also give you more options for descriptors, but in any case I think the descriptions could improve.

src/nomad_simulations/model_system.py
```diff
@@ -317,59 +562,79 @@ class AtomicCell(Cell):
         """,
     )
 
-    def __init__(self, m_def: Section = None, m_context: Context = None, **kwargs):
-        super().__init__(m_def, m_context, **kwargs)
-        # Set the name of the section
-        self.name = self.m_def.name
+    geometry_analysis_cutoffs = Quantity(
```
JFRudzinski: why plural?

src/nomad_simulations/model_system.py
JFRudzinski: This quantity needs units as well. And how are you going to link to the individual distributions? Why doesn't this go inside GeometryDistribution?

src/nomad_simulations/model_system.py
```diff
@@ -231,25 +254,36 @@ class Cell(GeometricSpace):
         """,
     )
 
+    # topology
+    bond_list = Quantity(
+        type=np.int32,
+        description="""
+        List of index pairs corresponding to (covalent) bonds, e.g., as defined by a force field.
+        """,
+    )  # ? @JosephRudzinski: move position?
```
JFRudzinski: I think actually this should go under AtomicCell directly, it's not really a byproduct of an analysis of the Geometry, rather it comes from the pre-specified model parameters (at least in my case).

We could end up making a section sort of like atoms_state but for "topological" quantities, but for now with just the single quantity I think it can stand alone.

src/nomad_simulations/model_system.py
```diff
@@ -231,25 +254,36 @@ class Cell(GeometricSpace):
...
+
+    # geometry and analysis
     positions = Quantity(
         type=np.float64,
         shape=['n_cell_points', 3],
         unit='meter',
         description="""
-        Positions of all the atoms in Cartesian coordinates.
+        Positions of all the atoms in absolute, Cartesian coordinates.
```
JFRudzinski, line 244: What is this for exactly? I think perhaps the name should reflect the difference from the atomic_cell positions.

src/nomad_simulations/model_system.py
```diff
@@ -283,12 +318,216 @@ def normalize(self, archive, logger) -> None:
...
 
 
+def convert_combo_to_name(combo: tuple[str]) -> str:
+    """Convert a combo tuple to a hypen-separated string.
+    For angles, the center element is now indeed placed at the center."""
+    if 1 < len(combo) < 4:
+        if len(combo) == 3:
+            combo = (combo[1], combo[0], combo[2])  # flip order
+        return '-'.join(combo)
+    raise ValueError('Only pairs and triplets are supported.')
```
Bernadette-Mohr: Do you expect the central particle of an angle definition to be at the first position in the string? Is that handled somewhere upstream?
I'm afraid we can't expect that from the various simulation engines/methods.

src/nomad_simulations/model_system.py
ndaelman-hu: Because it may be defined for multiple elements.

src/nomad_simulations/model_system.py
ndaelman-hu: > What happens if the atom_labels are X and no cutoff is explicitly specified?

Good question. We get our element `X` from `ase`, so I expect it to be covered.
Anyhow, just checked: `X` has a default cutoff of `0.2` angstrom.

ndaelman-hu: > This quantity needs units as well.

Added.

src/nomad_simulations/model_system.py
ndaelman-hu: > The bin spacing (i.e., 0.001 in this case) should also be giving as a parameter, similar to the cutoff.

> I would probably specify a default number of bins, rather than a bin spacing

I see your point, however, these values will become searchable in the whole of NOMAD. That's part of the reason for generating a histogram (the other is storage capacity).

Anyhow, I'd stick with a single, default distribution. I chose 0.001 angstrom, as it's the limit of accuracy for any method. If you want, I can change the value, though.

src/nomad_simulations/model_system.py
ndaelman-hu: > I would still apply the same parameters as for distances for generality of usage.

I chose the default based on the order of accuracy limit. Since angle and distance histograms shouldn't be compared, comparing bin size is also non-sensical.

src/nomad_simulations/model_system.py
ndaelman-hu: The angle distribution format is set up in `DistributionFactory`, where it is indeed defined as such (see `get_elemental_triples_centered`). I could see to define a `dict` mapping or class to handle the definition.

It's not meant to be passed from a parser, at least in the first instance. Do you see a scenario where this could apply?

src/nomad_simulations/model_system.py
ndaelman-hu: > Instead of a repeating subsection of the same class type, I think this might need to be changed to a container section which may hold any geometrical representation.

Okay... What would be the defining feature of the container section then? That it's a histogram? I'm fine looking into defining this with you, but this sounds as it should be a follow-up issue. **I can open up an issue.**

> These subsections should then inherit from PhysicalProperty for consistency purposes

I'm not really sure about the approach behind `PhysicalProperty`, but it sounds contradicting to the multiple subsections statement. I would go either with a _(object) compositional approach_ or a _type system_, but not mix them.

src/nomad_simulations/model_system.py
ndaelman-hu: Yes, I'm, running into an issue here:
- the bin units are different for distances and angles
- setting the dynamically is not possible in NOMAD Python atm

I could, however, produce a `GeometryDistanceDistribution` and a `GeometryAngleDistribution` with the units specified.
Would this also lie in line with your suggestion for a common container class?

src/nomad_simulations/model_system.py
ndaelman-hu: > It sounds like you want to store a str with all element names involved in the calculation.

Indeed

> Why not a list of strings?

There is a quantity called `elements` that exaclt fulfills this condition. I'll rename it to `atom_labels`.

Some quantities, like `name` are rendered in a privileged way in the GUI. It just gives as nice depiction.

I think this is part of a more generic discussion on which is preferable: search (+ dynamic rendering) vs static rendering.

src/nomad_simulations/model_system.py
ndaelman-hu: Ow, apparantly , I am setting the units dynamically in `DistributionHistogram`. Let me check how well this works.

ndaelman-hu: Okay, doesn't work, am falling back to my previous suggestion.

src/nomad_simulations/model_system.py
ndaelman-hu: > What is this for exactly?

With a (periodic) cell defined, you could either express the position vectors in terms of the lattice vector basis set, or the basis set in which the lattice vectors were defined (typically spanning a Cartesian space).
We end up going with the latter, since it simplifies large-scale manipulation. I think it's also the most common convention in ab initio.

src/nomad_simulations/model_system.py
Bernadette-Mohr: No, that was really just what I wanted to know. Is the order of the atoms with respect to their position in the angle handled somewhere else so that you can be sure the central atom is always listed at the first position here? That does seem to be the case, so it's all good.

ndaelman-hu: Okay! I understand that the semantics of the format could be emphasized with a mapping or class, but it looks kinda like overkill to me. Let's just keep it in mind.

I did further specialize `GeometryDistribution`, where the semantics are clearly specified.

src/nomad_simulations/model_system.py
JFRudzinski: ??

src/nomad_simulations/model_system.py
JFRudzinski: The question was what is the difference between the positions stored here and the positions stored in atomic_cell? And can/should the attribute name here denote that difference (i.e., not just the description)?

src/nomad_simulations/model_system.py
JFRudzinski: That's fine, then please open an issue for later. But yes, the general structure is a histogram.

I don't see the contradiction. I am saying that GeometricDistribution() should (perhaps) inherit from PhysicalProperty(), and then the individual classes inherit from GeometricDistribution() as below. Am I missing something? (Anyway, we should also deal with this later)

src/nomad_simulations/model_system.py
JFRudzinski: I was not saying that angles and distances should literally have the same bin spacing magnitude, and I did not mean that they should not have default values. My point here was that The GeometryDistribution() should contain an attribute called `bin_spacing`, such that default values can be set individually within the classes DihedralGeometryDistribution(), etc. That way they can also be changed from the default if necessary.

JFRudzinski (approved):
It's good for now, just take a look at my comments.

JosePizarro3:
Let me know if you need a review. I opened a question about returning self, and whether if this is what we want.

src/nomad_simulations/model_system.py
ndaelman-hu: > My point here was that The GeometryDistribution() should contain an attribute called bin_spacing

Yes, I understood this from your previous comment too. Still, I'd advise against it for the following 2 reasons:

1. We want to make these histograms searchable. At NOMAD's scale, this means constructing an index. Schema uniformity plays a pretty big role (will defer to Lauri or Markus here) in achieving that performance. If storage was more of a concern, I'd instead go with a sparse representation. Actually, now that distances and angles are dealt with separately, we could even remove the `bins` altogether.
2. Schema complexity: I understand the desire to placate to each use case, but this comes with trade-offs (such as point 1). This setup is generic enough that we can accomodate all covalent distributions and a bit beyond. These classes could be used as a jumping-off point for similar distribution analyses, but at different scales. That's fine, but I'd store that under a different name (and different issue).

ndaelman-hu: Regarding the sparse vs fixed binning, I'm actually not sure which would be better. I could see to deploy a ES instance and try it out myself.

src/nomad_simulations/model_system.py
ndaelman-hu: Ow, yes, I see. Tbh, I think some confusion entered the setup. `Cell` is both the parent to `AtomicCell`, but seems to also contain a grid, i.e. `n_cell_points`. The `positions` and the rest refer to them, though my correction may be wrong.

Let me take a closer look. Chema did push some changes here. I'll get back to you, after merging those.

ndaelman-hu: @JosePizarro3 could you pls clarify the meaning of `Cell.n_cell_points`, especially wrt to `AtomicCell`?
I actually added a note for you there when I asked for the review.

src/nomad_simulations/model_system.py
ndaelman-hu: > ??

Sorry, missed this one.

> I think actually this should go under AtomicCell directly

Note that this is actually under `Cell`, not `GeometryDistribution`.
Anyhow, I'm using the `neighbor_list` for defining the analysis. These are generated independently. I understood from @Bernadette-Mohr that this is fine.

So, I'll just move this then.

src/nomad_simulations/model_system.py
JosePizarro3: There is a typo in the description. This should be fixed.

I tried to break the geometric quantities of a cell and the specific atoms sitting in those positions. `Cell` contains thus the `lattice_vectors` and the 0-dimensional points in real space, i.e., `positions` (and also velocities and other info about the Cell), while `AtomsState` has the info of the atom sitting in that position.

So essentially, these are the same both in `Cell` and `AtomicCell`. The only addition of the atomic cell is the atoms state in each of these positions. Whether these `positions` are occupied by entities which are atoms, this depends on the problem. It could be a more complex thing which is centered around them, and for modeling purposes is described like this.

src/nomad_simulations/model_system.py
ndaelman-hu: So it sounds like a _type_ approach, sure. How on point is `PhysicalProperty`, though? I'll open an issue after this one.

ndaelman-hu:
> > To support a more functional pipeline notation, next to the current imperative style, I make all computational functions and normalize return self. I checked with Area D that the latter does not violate the normalize interface specifications.

> I doubt it does not break anything, especially the `MetainfoNormalizer`.

I held off on responding, until testing the normalization. No issues there. It's also unclear why any should occur.

> the only reason I'd see to return self is to chain methods one after another

Exactly, it's a way of supporting functional pipelines. I mentioned adding this 2x before. It's not really a big deal, but we can go over the points you raised.

> In any case, I think this is not a good practice here [...]
> Did you check that the methods (besides normalize, in which I won't do it) you implemented this return, this kind of logic is what we want?

There's no special logic going on here, just functional composition. I.e. the behavior of these functions does not change ('), and therefore it does not invalidate any old code.

Regarding "good practices" (''), it's just supporting a more compact, often clearer style of writing code ('''). Unless we want to constrain devs to a very procedural style, there's no issue. Actually, these kinds of pipelines are very common in modules like `pandas` or `react`, and form the backbone of _functional programming_. They are indeed becoming popular again across the board.

(') 1 disclaimer: it does not change the behavior of functions that return `None`. A lot of attributes are now set by the normalizer, i.e. it calls an object function that returns a result. In some cases, like those decorated by `@property` this is fine.
However, I've had instances, where parsers need access to a computed property, and there this could be beneficial, see (''').

('') Good practices more so have to do with producing safe code (including CI/CD pipelines) and clear abstractions (supplemented with good docs). While style consistency can be conducive here, the real objective is _readability_. If a different style produces better readability in some subcases, then it's appropriate there. Python itself supports a mix of procedural and declaritive alternatives.

(''') The nice aspect of writing code this way, is that you explain all the manipulations on top of an object in 1 line. You can now produce complex, but well-defined states without having to trigger the section's entire normalization (which might fail if some data is missing). 

> And if so, I wonder whether you are willing to change ALL `normalize()` functions to now return self.

We can try it out here first. It would be good to apply it consistently, yes.
For our own normalizers, this is a small issue. Pretty trivial.

JosePizarro3:
I think you can simply leave it out. I really doubt this does not break the `MetainfoNormalizer` functionality, and I won't lose time on even opening a discussion about this topic.

 I also have my doubts on readability, as returning self and chaining methods are rather unreadable. The only benefit is just writing one-line code. But I guess computer scientists would have a different point of view, perhaps, on the fact that `class1.method_a(...).method_b(...).method_c(...)` is more readable... I mean, just look at the `normalize()` functions, they are pretty readable thanks to the fact that the functions are not chained one after another...

src/nomad_simulations/model_system.py
ndaelman-hu: > So essentially, these are the same both in Cell and AtomicCell.

Great! That's what I had to know. Then I'll remove `positions` and `velocities` from `AtomicCell`.

ndaelman-hu: > There is a typo in the description.

Where?

JosePizarro3: > > There is a typo in the description.
> 
> Where?

Change `atoms` for `points`. And you can rename the quantity above from `n_cell_points` to `n_points`.

JosePizarro3:
Btw, it seems to me that this feature would live better in the nomad-analysis plugin, but we can leave it here for now.

src/nomad_simulations/model_system.py
ndaelman-hu: The bin spacing is set under the `geometry_distributions` subsection, so this is already covered.
`normalize` does not accept any `kwargs`, so it's not appropriate to modify the bin spacing there either.
This is what `DistributionFactory` is for.

Actually, on 2nd thought, I'll also move the cutoff under `GeometryDistribution`.

src/nomad_simulations/model_system.py
ndaelman-hu: I tried out the performance a bit this week: results are unclear. Both seem viable. I'll defer this to a follow-up issue.

PR #59, opened 2024-04-29
Moving `iri` and `rank` to m_def

JosePizarro3:
I was wrongly mixing concepts for quantities and definitions. After talking with Area D, I decided to move `iri` and `rank` to `m_def` at the beginning of each property. This means that the `nomad/metainfo/metainfo.py` had to be modified to include these as part of the `class Section(Definition)` attributes.

### DIFF
--- a/src/nomad_simulations/physical_property.py
+++ b/src/nomad_simulations/physical_property.py
@@ -54,18 +54,12 @@ class PhysicalProperty(ArchiveSection):
     # TODO add `errors`
     # TODO add `smearing`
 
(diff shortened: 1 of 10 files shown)


JosePizarro3:
I need to check the situations in which `self.m_def.rank` is defined finite: can I do:
```python
m_def = Section(iri='...', rank['n_orbitals'])
n_orbitals = Quantity(type=np.int32)
```
??

JosePizarro3:
Closing this in favor of whatever Area D implements in `nomad-lab`.

PR #61, opened 2024-04-30
Fix `HoppingMatrix` with new schema

### LINKED ISSUE #60: Fix `HoppingMatrix` with new schema
Currently, `HoppingMatrix` is a shared section between `ModelMethod` and `Outputs` and lives in a module `common.py`. However, it makes more sense to move its definition to `properties/hopping_matrix.py` and use the functionalities implemented for `PhysicalProperty`.

JosePizarro3:
I added `HoppingMatrix` and `CrystalFieldSplitting` properties, as well as `WignerSeitz` as a variable. 

@JFRudzinski @Bernadette-Mohr @ndaelman-hu please, note that rank, in order for setting the dynamic shape of `value` requires to have some quantities already define **during instantiation** of the class. In this example of `HoppingMatrix`, `rank=[self.n_orbitals, self.n_orbitals]`, hence we need to instantiate during parsing/normalizing like `hopping_matrix = HoppingMatrix(n_orbitals=4)`.

I still need to:
- [x] Add testing here
- [x] Add normalization in HoppingMatrix to extract other physical properties (like DOS, bands, Fermi surfaces) --> TODO in another branch with the help of @mohammadnakhaee 

Closes #60


JosePizarro3:
I will merge this (just a simple definition with dummy testing) and work on #52 . Then I will revisit the fact that from `HoppingMatrix` we can generate electronic structure properties (DOS, ElectronicEigenvalues, ElectronicBandGap, etc).

PR #63, opened 2024-05-13
 Add ElectronicEigenvalues and ElectronicBandStructure properties

### LINKED ISSUE #52: Add `ElectronicEigenvalues` and `ElectronicBandStructure` properties

JosePizarro3:
This pr contains the definition for the former `ElectronicBandStructure` with the new schema (`PhysicalProperties` and `Variables`), as well as the new property `ElectronicEigenvalues`. During development, I also wanted to define `FermiSurface`, but this makes more sense in a separate pr (#78).

Some important changes with respect to the old schema and normalization:
- Now, there is a `value_contributions` whose objective is to group all the former stuff that was in `BandEnergies` as `value_<name-of-the-contribution>`
- Now there is not a concept of segments. I rather changed for storing `KLinePath(Variables).points` as a flat list of points
- I completely restructured the old `add_band_gap` normalization. Now, we have `extract_band_gap` and, mainly, `resolve_homo_lumo_eigenvalues` to handle this.
- The old `get_special_points` function is re-defined under `numerical_settings.py:KMesh`. This was a very buggy function, specially when the path is not defined for points related by symmetry (e.g., `'X'` is both `[0.5, 0, 0]` and `[-0.5, 0, 0]`. Now we aim people to define the strings related with the high symmetry points OR manually define the path.

One question is whether `reciprocal_cell` will be still here in this schema or we can resolve it in another way. This quantity is only interesting for the front end.


JosePizarro3:
Hey @ladinesa this is the pr I told you about.

The focus should be the changes in the new module, `src/nomad_simulations/properties/band_structure.py`. I still have some todos, but I moved them to separated issues (#77 #78 #79).

Thanks!

src/nomad_simulations/properties/band_structure.py
```diff
@@ -0,0 +1,370 @@
...
+        Value of the electronic eigenvalues.
+        """,
+    )
+
+    def __init__(
+        self, m_def: Section = None, m_context: Context = None, **kwargs
+    ) -> None:
+        super().__init__(m_def, m_context, **kwargs)
+        # ! `n_bands` need to be set up during initialization of the class
+        self.rank = [self.n_bands]
```
ladinesa: or maybe get it from kwargs

src/nomad_simulations/properties/band_structure.py
```diff
@@ -0,0 +1,370 @@
...
+        """,
+    )
+
+    def __init__(
+        self, m_def: Section = None, m_context: Context = None, **kwargs
+    ) -> None:
+        super().__init__(m_def, m_context, **kwargs)
+        self.name = self.m_def.name
+
+    def _check_occupation_shape(self, logger: BoundLogger) -> bool:
```
ladinesa: maybe rename this to is_valid_occupation

src/nomad_simulations/properties/band_structure.py
```diff
@@ -0,0 +1,370 @@
...
```
ladinesa: not sure why you are passing logger? i would define it globally LOGGER = utils.get_logger(__name__)

src/nomad_simulations/properties/band_structure.py
```diff
@@ -0,0 +1,370 @@
...
+            logger (BoundLogger): The logger to log messages.
+
+        Returns:
+            (Tuple[Optional[pint.Quantity], Optional[pint.Quantity]]): The `highest_occupied` and
+            `lowest_unoccupied` eigenvalues.
+        """
+        # Sorting `value` and `occupation`
+        sorted_value, sorted_occupation = self.order_eigenvalues(logger)
+        if sorted_value is None or sorted_occupation is None:
+            if self.highest_occupied is not None and self.lowest_unoccupied is not None:
```
ladinesa, line 182: are you not supposed to return right away self.highest(lowest)(un)occupied if they are assigned? or maybe check if the values are consistent.

src/nomad_simulations/properties/band_structure.py
```diff
@@ -0,0 +1,370 @@
...
+        sorted_value_unit = sorted_value.u
+        sorted_value = sorted_value.magnitude
+
+        # Binary search ot find the transition point between `occupation = 2` and `occupation = 0`
+        tolerance = 1e-3  # TODO add tolerance from config fields
+        homo = self.highest_occupied
+        lumo = self.lowest_unoccupied
+        low_occ = 0
+        high_unocc = sorted_value.shape[-1] - 1
+        while low_occ <= high_unocc:
```
ladinesa: can't you use np.argwhere(sorted_occupation > tol) here? also, how about spin-pol case?

src/nomad_simulations/properties/band_structure.py
```diff
@@ -0,0 +1,370 @@
...
+        Extract the Fermi surface for metal systems and using the `FermiLevel.value`.
+
+        Args:
+            logger (BoundLogger): The logger to log messages.
+
+        Returns:
+            (Optional[FermiSurface]): The extracted Fermi surface section to be stored in `Outputs`.
+        """
+        # Check if the system has a finite band gap
+        homo, lumo = self.resolve_homo_lumo_eigenvalues(logger)
```
ladinesa: why not access the homo and lumo from msection instead of runnung the function again?

ladinesa:
Looks all fine to me except for some minor improvements. Do I check also the tests?

JosePizarro3:
> Looks all fine to me except for some minor improvements. Do I check also the tests?

Thanks! I think the tests are just there to help understand the code, so up to you. Coverage is 99% (whatever that means 😅), so I guess I did a good job.

src/nomad_simulations/properties/band_structure.py
JosePizarro3: Yeah, sorry, at this point I do not care too much about this; at the end, this is a definition, so it should go in m_def. 

I am waiting for Markus to implement something on this direction, so my point is: don't pay too much attention to this dirty assignment for now 😄

src/nomad_simulations/properties/band_structure.py
JosePizarro3: I used `_check` for this, but indeed, `is_valid` sounds better as a validation mechanism. I will change it in other modules as well.

src/nomad_simulations/properties/band_structure.py
ladinesa: I was just concerned about the same error you were having previously that you have to convert it to int when read from msection

src/nomad_simulations/properties/band_structure.py
JosePizarro3: Yes, this is confusing for me. As I am passing logger to `normalize()` would these conflict? I thought that I had to keep the same `logger` in all places to write the messages in the same bound logger.

src/nomad_simulations/properties/band_structure.py
JosePizarro3: Good point, but my argument is that normalization is more important here than parsing, i.e., if sorted_value and sorted_occupation could be extracted, then we used those (and not the parsed data) for extracting `highest_occupied` and `lowest_unoccupied`. Otherwise, we return then if they exist.

src/nomad_simulations/properties/band_structure.py
JosePizarro3: Hehe, I'll argue that binary search (which is logN) is more optimal than argwhere (which is linear). These arrays are going to be potentially large, but I can change it to argwhere if even tho you prefer readability.

src/nomad_simulations/properties/band_structure.py
JosePizarro3: I thought about it, but what if someone wants to `extract_band_gap` or `extract_fermi_surface` when homo and lumo are not set up and without normalizing first?

src/nomad_simulations/properties/band_structure.py
ladinesa: ah yeah right then ignore my comment.

src/nomad_simulations/properties/band_structure.py
JosePizarro3: Yes, I had to, thanks for catching it up. In the new version you can see I did `self.rank = [int(self.n_bands)]`, but I will change it 👍🏻

src/nomad_simulations/properties/band_structure.py
ladinesa: there is also this np.argmax that seems to return the first element matching criterion.

src/nomad_simulations/properties/band_structure.py
ladinesa: ah my bad, the function already has a check for the homo/lumo values in the archive.

src/nomad_simulations/properties/band_structure.py
JosePizarro3: Ok, I went for `validate_` instead.

src/nomad_simulations/properties/band_structure.py
JosePizarro3: It seems to be linear as well. ChatGPT helped me check this (for a sample of 10000 points)

time_complexity

src/nomad_simulations/properties/band_structure.py
JosePizarro3: no problem, indeed it checks this

src/nomad_simulations/properties/band_structure.py
ladinesa: interesting. thanks for testing. then go for the optimized approach.

JosePizarro3: I just found `numpy.searchsorted` which seems to perform binary search. Let me try it.

JosePizarro3: It seems to be almost the same. I tried some random arrays (hence the bit of randomness between results), averaged, and increased the size to 1e6.

time_complexity

Well, I can also use numpy.searchsorted if it is more readable, I don't mind actually.

ladinesa: I suggest you use numpy for easy maintenance.

src/nomad_simulations/numerical_settings.py
```diff
@@ -803,8 +803,8 @@ def resolve_reciprocal_lattice_vectors(
             # General checks to proceed with normalization
             if is_not_representative(model_system, logger):
                 continue
-            # TODO extend this for other dimensions (@ndaelman-hu)
-            if model_system.type != 'bulk':
```
ndaelman-hu: Don't remove this TODO, pls.

.vscode/settings.json
ndaelman-hu, line 1: Settings should be in `.gitignore`. Nobody wants their VSCode setup messed up because they opened a project.

src/nomad_simulations/numerical_settings.py
```diff
@@ -165,15 +165,15 @@ class KSpaceFunctionalities:
     A functionality class useful for defining methods shared between `KSpace`, `KMesh`, and `KLinePath`.
     """
 
-    def _check_reciprocal_lattice_vectors(
+    def validate_reciprocal_lattice_vectors(
```
ndaelman-hu, line 168: So, the name and visibility change are all fine.
I'd just ask to keep these for separate PRs (don't even have to be reviewed) as much as possible, as they distract from the important code changes at hand.

src/nomad_simulations/properties/band_structure.py
```diff
@@ -0,0 +1,377 @@
...
+from nomad_simulations.physical_property import PhysicalProperty
+from nomad_simulations.properties import ElectronicBandGap, FermiSurface
+from nomad_simulations.utils import get_sibling_section
+
+
+class BaseElectronicEigenvalues(PhysicalProperty):
+    """
+    A base section used to define basic quantities for the `ElectronicEigenvalues`, `FermiSurface`, and
+    `ElectronicBandStructure` properties. This section serves to define `FermiSurface` and without needing to specify
+    other quantities that appear in `ElectronicEigenvalues`
```
ndaelman-hu: > This section serves to define `FermiSurface` ~~and~~ without needing to

I'd even shorten it to: "This section defines common quantities between `FermiSurface` and `ElectronicBandStructure`.

src/nomad_simulations/properties/band_structure.py
```diff
@@ -0,0 +1,377 @@
...
+        type=np.float64,
+        shape=['*', 'n_bands'],
+        description="""
+        Occupation of the electronic eigenvalues. This is a number between 0 and 2, where 0 means
+        that the state is unoccupied and 2 means that the state is fully occupied. It is controlled
+        by the Fermi-Dirac distribution:
+
+            $ f(E) = 1 / (1 + exp((E - E_F) / kT)) $
+
+        The shape of this quantity is defined as `[KMesh.n_points, KMesh.dimensionality, n_bands]`, where `KMesh` is a `variable`.
```
ndaelman-hu: 1. Since you seem to be handling spin-states, shouldn't it be restricted to 1? I guess what you want, is that the max filling depends on whether spin_channel is set, right? Either elaborate in the description OR add a quantity `max_occupation`. The latter would simplify the code logic.
2. I understand what you want to say with the line on the Fermi-Dirac distribution, but it comes across as confusing, e.g. the band structure also depends on other factors, wha about strongly-correlated systems, etc.  Most of our band structure don't have any T-dependence, so I wouldn't bother mentioning this. For those that do, it'll be the calculation that dictates the data.

src/nomad_simulations/properties/band_structure.py
```diff
@@ -0,0 +1,377 @@
...
+    )
+
+    lowest_unoccupied = Quantity(
+        type=np.float64,
+        unit='joule',
+        description="""
+        Lowest unoccupied electronic eigenvalue. Together with `highest_occupied`, it defines the
+        electronic band gap.
+        """,
+    )
```
ndaelman-hu, line 109: As an aside, do you see any value in reporting the min/max of the eigenvalue sampling itself?
Most calculations don't cover the core states.

src/nomad_simulations/properties/band_structure.py
```diff
@@ -0,0 +1,377 @@
...
+
+    reciprocal_cell = Quantity(
+        type=KSpace.reciprocal_lattice_vectors,
+        description="""
+        Reference to the reciprocal lattice vectors stored under `KSpace`. This reference is useful when resolving the Brillouin zone
+        for the front-end visualization.
+        """,
+    )
```
ndaelman-hu: If you say that this is for visualization, this hould really only exist in `results`.
Personally, I think it can have a use here for other post-processing.

src/nomad_simulations/properties/band_structure.py
```diff
@@ -0,0 +1,377 @@
...
+        """
+        if self.occupation is None or len(self.occupation) == 0:
+            logger.warning('Cannot find `occupation` defined.')
+            return False
+        if self.value is not None and self.value.shape != self.occupation.shape:
+            logger.warning(
+                'The shape of `value` and `occupation` are different. They should have the same shape.'
+            )
+            return False
+        return True
```
ndaelman-hu: Just add `shape` to the definition of `value`. If you have to override it later on, that easy with `m_copy`.
Also, I'd use plural for quantities that store arrays.

src/nomad_simulations/properties/band_structure.py
```diff
@@ -0,0 +1,377 @@
...
+
+    def order_eigenvalues(
+        self, logger: BoundLogger
+    ) -> Tuple[Optional[pint.Quantity], Optional[np.ndarray]]:
```
ndaelman-hu: No need for `Optional` here, it'll only complicate the logic down the line. Just use empty arrays to denote "failure". It'll keep the logic simpler.

src/nomad_simulations/properties/band_structure.py
```diff
@@ -0,0 +1,377 @@
...
+            logger (BoundLogger): The logger to log messages.
+
+        Returns:
+            (Tuple[Optional[pint.Quantity], Optional[list]]): The flattened and sorted `value` and `occupation`.
+        """
+
+        # Check if `value` exists
+        if self.value is None or len(self.value) == 0:
+            logger.error('Could not find `value` defined.')
+            return None, None
```
ndaelman-hu: Like here: this teste isn't necessary if you maintain arrays as the output.
Moreover, shouldn't instantiation assure that only 1 condition has to be met?

src/nomad_simulations/properties/band_structure.py
```diff
@@ -0,0 +1,377 @@
...
+
+        # Check if `occupation` exists and have the same shape as `value`
```
ndaelman-hu: typo in the conjugation

src/nomad_simulations/numerical_settings.py
JosePizarro3: indeed, I just tackled the name change at once, but you are right

src/nomad_simulations/properties/band_structure.py
JosePizarro3: I've changed the description, as strictly speaking `FermiSurface` as a property contains different values (not the energies, but the actual k points $\vec{k}_F$)

src/nomad_simulations/properties/band_structure.py
JosePizarro3: Indeed, you are right, I totally overlooked the spin. I added `spin_channel` and I will adapt the description.

I will also adapt the Fermi-Dirac distribution. However, for a system of fermions, the Fermi-Dirac statistics is a necessary condition, independently of whether they are correlated or not; this is what is behind of certain filling conditions, e.g. the [Luttinger's theorem](https://en.wikipedia.org/wiki/Luttinger%27s_theorem), and it only depends on whether the system is a Fermi liquid or not, hence whether we can apply Fermi statistics or not. The only situations where these does not apply is near quantum criticality and non-Fermi liquid behaviors (I mean, the name says it everything 😄)$^{[1]}$.


$^{[1]}$ I was trying to find a good reference for this, but I couldn't. The most I can share with you is some of the initial chapters of Piers Coleman's book, [Introduction to Many Body Physics](https://www.cambridge.org/core/books/introduction-to-manybody-physics/B7598FC1FCEE0285F5EC767E835854C8). Let me know if you want the book.

src/nomad_simulations/properties/band_structure.py
JosePizarro3: I think it would be interesting to cover min/maxs of specific set of bands, e.g., around the Fermi level, higher energy bands separated by gaps, and equivalently, core bands separated by gaps.

I will add a question to see whether we can do this later along the road. Thanks for bringing it up.

src/nomad_simulations/properties/band_structure.py
JosePizarro3: Ok, I delete the phrase referring to the front-end visualization.

src/nomad_simulations/properties/band_structure.py
JosePizarro3: Problem comes with the units of the returned `sorted_value`.

src/nomad_simulations/properties/band_structure.py
JosePizarro3: Well, someone could call for this method without `value` set up. Now, for the `is None` and `len()`, it is true that the length check does not make sense, so I will remove it.

src/nomad_simulations/properties/band_structure.py
JosePizarro3: There are other things going on with `value` and its `__setattr__` defined in `PhysicalProperty`, so defining `shape` will not have any effect.

I just want to double check that `occupation` and `value` has the same shape, but it is true that this could maybe be a decorator, as it can become a quite popular check?

JosePizarro3:
Ok, this pr is ready. I added also a validator for comparing a quantity with respect to `value` in any PhysicalProperty, to ensure consistence in all the cases where we don't want to consider a quantity a physical property, but to be actually under an existing one.

Thanks for the review @ladinesa @ndaelman-hu ! 🥳

src/nomad_simulations/properties/band_structure.py
ndaelman-hu: I'll take a look. I thought that correlation would be a factor in deviating from the Fermi liquid model, but I'll read up some more.

src/nomad_simulations/properties/band_structure.py
ndaelman-hu: I think a lot of our checks could be recast into decorators. I was already looking into it, but I'd have to gather all use cases. This is cleaning up for later.

src/nomad_simulations/properties/band_structure.py
ndaelman-hu: But, empty lists can also have units...

JosePizarro3: Yeah, but knowing the unit implies that the extraction works. I think None and Optional is much more clean. I also doubt that an empty pint list is not screwing things on the archive, meaning, that the quantity is set to an empty value instead of being None, which is the behavior we want

ndaelman-hu: > I also doubt that an empty pint list is not screwing things on the archive, meaning, that the quantity is set to an empty value instead of being None, which is the behavior we want

That would be good to know, actually! Let's check it out. Tbh, the archive should have a way of handling empy arrays.

ndaelman-hu: > I think None and Optional is much more clean.

They are definitely more verbose, which indeed can be more communicative.
I'd call simpler logic _cleaner_, though. If a type (e.g. `list`, `int`) can handle "failure" (e.g. `[]`, `0`) by itself, I wouldn't complicate matters.
This is, of course, very situation-dependent, as sometimes "failure" may present itself differently, e.g. `1 / 0`.

Honestly, Python's way of handling `None` is pretty naive. In a language like Haskell I'd make less of an issue, as the logic is better separated. There, `Nothing` is a constructor of the **monad** `Maybe`, which handles "failure" propogation in a more automated and modular fashion.

Issue #64, opened 2024-05-13
Adapting the new entrypoints plugin mechanism and changing name of repo and package


JosePizarro3:
@ndaelman-hu @Bernadette-Mohr @JFRudzinski @ladinesa 

With your agreement, I'd like to change the name of the package to `nomad_simulations_schema`. I also need to fix imports to be absolute to the package path and not relative (i.e., change `from .numerical_settings import KMesh` to `from nomad_simualtions_schema.numerical_settings import KMesh`).

My idea was that:
- We call the general repo for parsers, `nomad_simulations_parsers`.
- Each individual parser is called `nomad_<codename>_parser`.
- The workflow plugin is called `nomad_simulations_workflow_schema`.

I think that would be enough to be clear. Still, the path in the NOMAD central repo should be `dependencies/parsers/simulation/...` or `dependencies/schema/simulation/...`.

ndaelman-hu:
> With your agreement, I'd like to change the name of the package to nomad_simulations_schema

What's the _time frame_ here? I don't care much for the naming, but I want a heads-up so all my open issues can first be merged.
Then I can deal with reconfiguring.

Is this now the default naming convention then, i.e. can you assure that there won't be any other renamings down the road?

JosePizarro3:
> What's the _time frame_ here? I don't care much for the naming, but I want a heads-up so all my open issues can first be merged. Then I can deal with reconfiguring.

It is a quick patch, so no problem on waiting, but the decision has to be made.

> Is this now the default naming convention then, i.e. can you assure that there won't be any other renamings down the road?

The naming convention is the one we will decide.

ladinesa:
okay for me.

ndaelman-hu:
> > What's the _time frame_ here? I don't care much for the naming, but I want a heads-up so all my open issues can first be merged. Then I can deal with reconfiguring.
> 
> It is a quick patch, so no problem on waiting, but the decision has to be made.

I'm finishing up 2 issues. Pls hold off till after merging those. This has no urgency.

> > Is this now the default naming convention then, i.e. can you assure that there won't be any other renamings down the road?
> 
> The naming convention is the one we will decide.

So, no, it's not guaranteed to last.

JosePizarro3:
> So, no, it's not guaranteed to last.

Why not? I think we can make this decision ourselves

JosePizarro3:
[...]

PR #66, opened 2024-05-13
 Move config tolerances to utils

### LINKED ISSUE #65: Move config tolerances to utils

JosePizarro3:
Hi @ladinesa 

I moved some of the config.normalize values to here, as the normalization will now be contained in the datamodel. For now, I only added the relevant config values found in default.yaml in NOMAD, but this has to be extended when needed in normalization.

What do you think?

Closes #65

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ requires-python = ">=3.9"
 dependencies = [
     "nomad-lab>=1.2.2dev578",
     "matid>=2.0.0.dev2",
+    'pydantic>=1.10.8,<2.0.0',
 ]
 
 [project.urls]
--- a/src/nomad_simulations/model_system.py
+++ b/src/nomad_simulations/model_system.py
@@ -33,7 +33,6 @@ from matid.classification.classifications import (
     Class3D,
 )  # pylint: disable=import-error
 
-from nomad import config
 from nomad.units import ureg
 from nomad.atomutils import Formula, get_normalized_wyckoff, search_aflow_prototype
 
@@ -43,7 +42,7 @@ from nomad.datamodel.metainfo.basesections import Entity, System
 from nomad.datamodel.metainfo.annotations import ELNAnnotation
 
 from .atoms_state import AtomsState
-from .utils import get_sibling_section, is_not_representative
+from .utils import get_sibling_section, is_not_representative, config
 
 
 class GeometricSpace(Entity):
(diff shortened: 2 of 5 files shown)


ladinesa:
imho, these are indeed schema-specific settings however, there should be a mechanism to override these values though the nomad config file. My only concern if you move it at this time is that these are also used in the legacy schema, hence you will have to update in two places. So my advice is to keep them in nomad for the mean time.

JosePizarro3:
> imho, these are indeed schema-specific settings however, there should be a mechanism to override these values though the nomad config file. My only concern if you move it at this time is that these are also used in the legacy schema, hence you will have to update in two places. So my advice is to keep them in nomad for the mean time.

Maybe I can move the default values to the `nomad.yaml` of the repo, but I am not sure how to load them then for the main nomad installation. Perhaps config_tolerances.py should be moved to its own subfolder and have some functionality to initialize the default values from the nomad.yaml?

About the legacy schema, all fine. It can stay with its own config for now. We can think on different names.

ladinesa:
> > imho, these are indeed schema-specific settings however, there should be a mechanism to override these values though the nomad config file. My only concern if you move it at this time is that these are also used in the legacy schema, hence you will have to update in two places. So my advice is to keep them in nomad for the mean time.
> 
> Maybe I can move the default values to the `nomad.yaml` of the repo, but I am not sure how to load them then for the main nomad installation. Perhaps config_tolerances.py should be moved to its own subfolder and have some functionality to initialize the default values from the nomad.yaml?

they should be in defaults.yaml I think. I am not sure about schemas but we should probably do something similar to parsers where the kwargs are loaded when module is loaded. 

> 
> About the legacy schema, all fine. It can stay with its own config for now. We can think on different names.

JosePizarro3:
The problem is if someone wants to use this plugin in an OASI how do they set up these parameters? In my head, I always imagine that people will touch anyways the functionalities and define their tolerances in Python, so they will hardly use a yaml to pass the configs of these parameters. But I'd might be wrong, you tell me.

ladinesa:
> The problem is if someone wants to use this plugin in an OASI how do they set up these parameters? In my head, I always imagine that people will touch anyways the functionalities and define their tolerances in Python, so they will hardly use a yaml to pass the configs of these parameters. But I'd might be wrong, you tell me.

The plugin has defaults, and nomad has defaults.yaml or nomad.yaml which overrides these if defined.

JosePizarro3:
> > The problem is if someone wants to use this plugin in an OASI how do they set up these parameters? In my head, I always imagine that people will touch anyways the functionalities and define their tolerances in Python, so they will hardly use a yaml to pass the configs of these parameters. But I'd might be wrong, you tell me.
> 
> The plugin has defaults, and nomad has defaults.yaml or nomad.yaml which overrides these if defined.

Ok, not sure how these defaults are defined. Do you mean in the nomad_plugin.yaml? Can I define a new key and use it for these defaults?

JosePizarro3:
Closing this in favor of #64

PR #71, opened 2024-05-16
KSpace definition

JosePizarro3:
@ndaelman-hu I would need a quick review on this to be able to keep working on #52

PR #72, opened 2024-05-21
 Add DielectricFunction properties

### LINKED ISSUE #34: Add `DielectricFunction` properties
@ndaelman-hu

We have to add the `DielectricFunction` base class with the same functionalities as in the normalizers. Important links to the old schema:

[Spectra normalizer](https://github.com/nomad-coe/nomad-normalizer-plugin-spectra)

JosePizarro3:
Another property added to the list. In this case, there is a renaming to `Permittivity` to be in line with the taxonomy. I have to check if we can define `DielectricStrength`, and how to extract it from `Permittivity`.

The design idea here is that we want to store a 3x3 tensor $\epsilon(q, \omega)$. If it does not depend on $q$, we can extract the absorption spectrum as the imaginary term like $I(\omega) = Im \epsilon(\omega)$. Like the spectrum has a `rank=[]`, and the permittivity has `rank=[3, 3]`, I just needed to create 3 sub-sections for each principal direction for the spectrum.

I also:
- Added a new utils function to extract the specific list of variables (function `get_variables()`).
- Added abstraction for spectrum, defining `AbsorptionSpectrum` as a more abstract category for XAS, EELS. I added `axis` as a MEnum
- I had to add `KMesh` (from #71), and I also defined `Frequency` for variables
- Added `Permittivity.type` for simpler categorization


JosePizarro3:
Hi @ndaelman-hu, @ladinesa 

I am not 100% who is better for reviewing this: I don't want comments on the schema, but rather on the implementation of functionalities and check that the testing is correct.

Let me know.

src/nomad_simulations/utils/utils.py
```diff
@@ -129,3 +129,26 @@ def is_not_representative(model_system, logger: BoundLogger):
...
+        variable_cls (Variables): The class of the variables to get.
+
+    Returns:
+        (List[Variables]): The list of variables which are of type `variable_cls`.
+    """
+    if variables is None:
+        return []
+    result = []
+    for var in variables:
+        if isinstance(var, variable_cls):
```
ndaelman-hu, line 152: How specific does the type checking have to be? Should child classes also return `True`?

src/nomad_simulations/utils/utils.py
```diff
@@ -129,3 +129,26 @@ def is_not_representative(model_system, logger: BoundLogger):
...
+    """
+    Get the list of variables which are of type `variable_cls` and appear under `variables`.
+
+    Args:
+        variables (List[Variables]): The list of variables to check.
+        variable_cls (Variables): The class of the variables to get.
+
+    Returns:
+        (List[Variables]): The list of variables which are of type `variable_cls`.
+    """
```
ndaelman-hu: This approach is fine for small archives.
For larger, you'd likely want a specialized container class that holds an inverse index.

tests/test_utils.py
```diff
@@ -59,3 +65,16 @@ def test_is_not_representative():
 
 
 # ! Missing test for RusselSandersState (but this class will probably be deprecated)
+
+
+def test_get_variables():
```
ndaelman-hu: Add a test for parent classes, e.g. `PotentialEnergy` inherits from `Energy`.
Check what happens when you request `Energy` then.

tests/test_utils.py
```diff
@@ -59,3 +65,16 @@ def test_is_not_representative():
...
+    """
+    Test the `get_variables` utility function
+    """
+    variables = [
+        Temperature(points=list(range(-3, 4)) * ureg.kelvin),
+        Energy(points=list(range(-3, 4)) * ureg.joule),
+    ]
+    energies = get_variables(variables, Energy)
+    assert len(energies) == 1
+    assert np.allclose(energies[0].points.magnitude, [-3, -2, -1, 0, 1, 2, 3])
```
ndaelman-hu: I think you can just check if the objects really match, i.e.
` assert energies == [variables[0]]`

src/nomad_simulations/variables.py
```diff
@@ -150,3 +152,76 @@ def __init__(
...
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
+
+
+class KMesh(Variables):
+    """
+    K-point mesh over which the physical property is calculated. This is used to define `ElectronicEigenvalues(PhysicalProperty)` and
+    other k-space properties. The `points` are obtained from a refernece to the `NumericalSettings` section, `KMesh(NumericalSettings)`.
+    """
```
ndaelman-hu: Wouldn't this be more appropriate in the KPoints PR?

src/nomad_simulations/utils/utils.py
```diff
@@ -129,3 +129,26 @@ def is_not_representative(model_system, logger: BoundLogger):
...
+    if variables is None:
+        return []
```
ndaelman-hu, line 149: The optional input is a bit non-sensical... You can just simplify this to a list and retain the logic below.

tests/test_utils.py
ndaelman-hu: You should also add empty lists (and `None`, if you decide to support `Optional`).

Btw, I would run those as separate tests from this case, using parametrize: each test should look for a different failure mode.

src/nomad_simulations/properties/permittivity.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        # If the `pemittivity` depends on the scattering vector `q`, then we cannot extract the absorption spectrum
+        q_mesh = get_variables(self.variables, KMesh)
+        if len(q_mesh) > 0:
+            logger.warning(
+                'The `permittivity` depends on the scattering vector `q`, so that we cannot extract the absorption spectrum.'
+            )
+            return None
+        # Extract the `Frequency` variable to extract the absorption spectrum
+        frequencies = get_variables(self.variables, Frequency)
```
ndaelman-hu, line 91: If this is all `get_variables` is used for, I'd just use a one-liner, or define a quick lambda function. It hardly has the complexity to merit being a utility.

Moreover, I think you're restricting the usage scope of this function: it isn't limited to just `variables`. You could use the logic in other places too.

ndaelman-hu (commented):
I only checked the `get_variables` utility, as you asked.
Check whether it's really worth having it as a utility and if so, the edge cases I mentioned.

I think this shows 1 weakness of the naive flat schema, is that we need these scans to provide lookup. For small archives, there's no real issue, but I'm unsure about search at the DB level for 2 reasons:

1. the schema does not serialize the actual class. Will we be able to read in more specialized child classes?
2. I don't know how an ES query will perform on these lists. Actually, do you have any tag to look for?

These 2 questions are more general than this PR. Just something th at I noticed.

ndaelman-hu:
> Hi @ndaelman-hu, @ladinesa
> 
> I am not 100% who is better for reviewing this: I don't want comments on the schema, but rather on the implementation of functionalities and check that the testing is correct.
> 
> Let me know.

Did a small review. Btw, you can also ask @Bernadette-Mohr. I think she's more than capable providing feedback on the utility.

src/nomad_simulations/variables.py
JosePizarro3: Yeah, but while I am waiting, I just speed up the development in this branch by copy-pasting what is in there. Skip it and simply review it when we talk in #71 👍🏻

src/nomad_simulations/properties/permittivity.py
JosePizarro3: There is also some usage in spectral_profile.py. But we will use it in other situations.

This utility just returns the list of variables which matches the cls type passed. I will double-check metainfo.py to see whether this exists or not.

src/nomad_simulations/utils/utils.py
JosePizarro3: I am not sure if I understand the point of small archives. properties tend to be only of in terms of a few variables.

src/nomad_simulations/utils/utils.py
JosePizarro3: Well, this is defined by the `variable_cls`. Do you want me to be very specific and only the specific class (and not the inherited ones) is checked?

JosePizarro3:
> I only checked the `get_variables` utility, as you asked. Check whether it's really worth having it as a utility and if so, the edge cases I mentioned.
> 
> I think this shows 1 weakness of the naive flat schema, is that we need these scans to provide lookup. For small archives, there's no real issue, but I'm unsure about search at the DB level for 2 reasons:
> 
> 1. the schema does not serialize the actual class. Will we be able to read in more specialized child classes?
> 2. I don't know how an ES query will perform on these lists. Actually, do you have any tag to look for?
> 
> These 2 questions are more general than this PR. Just something th at I noticed.

Not sure if I understand the point of the weakness. This data structure only stores the data, and in case of talking about querying, I doubt searching for specific `variables` would ever be a topic (but rather, that a property has a variable_cls defined).

>  Actually, do you have any tag to look for?

Again, I am never talking about ES in this schema.

JosePizarro3:
> > Hi @ndaelman-hu, @ladinesa
> > I am not 100% who is better for reviewing this: I don't want comments on the schema, but rather on the implementation of functionalities and check that the testing is correct.
> > Let me know.
> 
> Did a small review. Btw, you can also ask @Bernadette-Mohr. I think she's more than capable providing feedback on the utility.

Thanks! I will incorporate your feedback and merge. I think it makes sense if I actually merge after #71

src/nomad_simulations/utils/utils.py
ndaelman-hu: > Well, this is defined by the variable_cls

No, `isinstance` determines this, and it will return `True` for a parent, i.e. you will get `PotentialEnergy` sections / objects too when requesting `Energy`.

> Do you want me to be very specific and only the specific class (and not the inherited ones) is checked?

Idk, I was just asking whether this is the behavior you want. You can always add an optional toggle if you're unsure.

src/nomad_simulations/utils/utils.py
ndaelman-hu: So, if I understood the suggested schema well, you will have 1 big array under `Output`, correct?
Up till now, you've only used `get_variables` for minor checks of individual properties. Do you foresee the need to search this property array as a whole?

src/nomad_simulations/variables.py
ndaelman-hu: ... You can just set a different target branch, then this won't show up.

src/nomad_simulations/properties/permittivity.py
ndaelman-hu: > This utility just returns the list of variables which matches the cls type passed.

Yes, I got that. That's why it's kind of trivial.

ndaelman-hu:
> > I only checked the `get_variables` utility, as you asked. Check whether it's really worth having it as a utility and if so, the edge cases I mentioned.
> > I think this shows 1 weakness of the naive flat schema, is that we need these scans to provide lookup. For small archives, there's no real issue, but I'm unsure about search at the DB level for 2 reasons:
> > 
> > 1. the schema does not serialize the actual class. Will we be able to read in more specialized child classes?
> > 2. I don't know how an ES query will perform on these lists. Actually, do you have any tag to look for?
> > 
> > These 2 questions are more general than this PR. Just something th at I noticed.
> 
> Not sure if I understand the point of the weakness. This data structure only stores the data, and in case of talking about querying, I doubt searching for specific `variables` would ever be a topic (but rather, that a property has a variable_cls defined).
> 
> > Actually, do you have any tag to look for?
> 
> Again, I am never talking about ES in this schema.

You mentioned before essentially copying a filtered version of this schema over into `results`, that's why I'm asking.
If so, we should consider the points I raised, though we can discuss it outside of this PR. If searching for `variable_cls` is supported, my point 2 remains.

JosePizarro3:
> > > I only checked the `get_variables` utility, as you asked. Check whether it's really worth having it as a utility and if so, the edge cases I mentioned.
> > > I think this shows 1 weakness of the naive flat schema, is that we need these scans to provide lookup. For small archives, there's no real issue, but I'm unsure about search at the DB level for 2 reasons:
> > > 
> > > 1. the schema does not serialize the actual class. Will we be able to read in more specialized child classes?
> > > 2. I don't know how an ES query will perform on these lists. Actually, do you have any tag to look for?
> > > 
> > > These 2 questions are more general than this PR. Just something th at I noticed.
> > 
> > 
> > Not sure if I understand the point of the weakness. This data structure only stores the data, and in case of talking about querying, I doubt searching for specific `variables` would ever be a topic (but rather, that a property has a variable_cls defined).
> > > Actually, do you have any tag to look for?
> > 
> > 
> > Again, I am never talking about ES in this schema.
> 
> You mentioned before essentially copying a filtered version of this schema over into `results`, that's why I'm asking. If so, we should consider the points I raised, though we can discuss it outside of this PR. If searching for `variable_cls` is supported, my point 2 remains.

I am still not sure. In `results` there will be a reference to the specific property (e.g., `permittivity`). The variables will never be searched, but probably it will be filtered to the result which variables we want to make searchable. This is as optimal as whatever is already in place for the `results` section when it is referencing properties in `run`.

Maybe I used wrong the word "copy"?

src/nomad_simulations/properties/permittivity.py
JosePizarro3: It is just a reusable function which will be probably used when parsing and normalizing PhysicalProperty.

src/nomad_simulations/variables.py
JosePizarro3: But I don't want to show everything. Disregard it and that's it. This branch will be eventually rebased anyways.

src/nomad_simulations/utils/utils.py
JosePizarro3: > So, if I understood the suggested schema well, you will have 1 big array under Output, correct?

No. There is a list of properties as sub-sections. Under each property, we have to parse the variables as sub-sections which contain `points` and the property `value` with the shape coincide with the length of these variables (+ the rank defined for the property).

> Do you foresee the need to search this property array as a whole?

I don't understand this.

src/nomad_simulations/utils/utils.py
JosePizarro3: Yeah, it might be useful to specify only the class, and not treat inheritance or just use a toggle like you said. I will implement this.

tests/test_utils.py
JosePizarro3: I think for now I will keep it like this. I have not any case where this will happen, but for sure we will see it later. 

I added  a todo.

src/nomad_simulations/utils/utils.py
JosePizarro3: Ok, see my comment below. I added a todo to tackle this later when we have an example of a variable which inherits from another variable.

JosePizarro3:
I think I will be merging this tomorrow morning! No problem, we can revisit this as well in the future.

PR #76, opened 2024-05-28
added branch composition to the modelsystem normalizer

### LINKED ISSUE #73: ModelSystem for H5MD
Schema adjustments necessary while working on parse_system of the H5MD plugin

JFRudzinski:


JFRudzinski:
@JosePizarro3, @Bernadette-Mohr 
I added the attribute `composition_formula` to ModelSystem, which describes the children of each ModelSystem with the notation X(n)Y(m), where X, Y are the branch labels and n,m are the number of repeating branches with the same label. This is applied during normalization only for representative systems. At the lowest level of the hierarchy, i.e., when a model system has no `model_system` subsection, composition formula then corresponds to the normal chemical formula in terms of the atom_labels. 

I am not attached to this notation, but I think it is useful for understanding what each branch is containing.  Let me know what you think. There are TODOs and ?'s that warrant addressing, please have a look at those.

src/nomad_simulations/model_system.py
```diff
@@ -901,6 +901,16 @@ class ModelSystem(System):
         """,
     )
 
+    composition_formula = Quantity(
+        type=str,
+        shape=[],
```
JosePizarro3: You can delete the empty shape.

src/nomad_simulations/model_system.py
```diff
@@ -901,6 +901,16 @@ class ModelSystem(System):
...
+        description="""
+        The overall composition of the system with respect to its subsystems.
+        The syntax for a system composed of X and Y with x and y components of each,
+        respectively, is X(x)Y(y).
```
JosePizarro3: Would you mind putting some examples? Some from the testing would do.

src/nomad_simulations/model_system.py
```diff
@@ -967,6 +977,25 @@ def normalize(self, archive, logger) -> None:
         if is_not_representative(self, logger):
             return
 
+        # Traversing the System Hierarchy
+        # TODO consider moving to utils and passing generic set of functions to execute at each branch level
+        def set_branch_composition(system, subsystems, atom_labels):
```
JosePizarro3: I suggest moving these two functions, `set_branch_composition` and `traverse_system_recurs` to a special method, `resolve_composition_formula` at the same level of `resolve_system_type_and_dimensionality`.

I suggest returning the actual composition_formula, and setting (in line 997):
```python
        atom_labels = [atom.chemical_symbol for atom in self.cell[0].atoms_state]
        if self.composition_formula is None:
                self.composition_formula = self.resolve_composition_formula(atom_labels)
```

JosePizarro3: Furthermore, I know typing is annoying, but would you mind adding it?

src/nomad_simulations/utils/utils.py
```diff
@@ -125,7 +125,8 @@ def is_not_representative(model_system, logger: BoundLogger):
     if model_system is None:
         logger.warning('The `ModelSystem` is empty.')
         return None
+    #? Can we somehow only print this warning once in the case of many systems?
```
JosePizarro3: Please, delete this warning fully, no need to pop any message to the logger if it will clutter it.

src/nomad_simulations/model_system.py
```diff
@@ -967,6 +977,25 @@ def normalize(self, archive, logger) -> None:
...
+            if subsystems is None:
+                subsystem_labels = [np.array(atom_labels)[system.get("atom_indices")]]
+            else:
+                subsystem_labels = [subsystem.get("branch_label") for subsystem in subsystems]
+            system.composition_formula = get_composition(subsystem_labels)
+
+        def traverse_system_recurs(system, atom_labels):
+            subsystems = system.get("model_system")
```
JosePizarro3:
```python
subsystems = system.model_system
set_branch_composition(system, subsystems, atom_labels)
if subsystems is not None:
...
```

The "is not None" is to cover issues when it is an empty list (together with the for loop it will be handled correctly).

src/nomad_simulations/model_system.py
```diff
@@ -967,6 +977,25 @@ def normalize(self, archive, logger) -> None:
         if is_not_representative(self, logger):
             return
 
+        # Traversing the System Hierarchy
+        # TODO consider moving to utils and passing generic set of functions to execute at each branch level
+        def set_branch_composition(system, subsystems, atom_labels):
+            if subsystems is None:
+                subsystem_labels = [np.array(atom_labels)[system.get("atom_indices")]]
+            else:
+                subsystem_labels = [subsystem.get("branch_label") for subsystem in subsystems]
```
JosePizarro3:
```python
subsystem_labels = [subsystem.branch_label for subsystem in subsystems if subsystem.branch_label is not None]
```
error handling for when `branch_label` is not set.

src/nomad_simulations/model_system.py
```diff
@@ -967,6 +977,25 @@ def normalize(self, archive, logger) -> None:
...
+            system.composition_formula = get_composition(subsystem_labels)
+
+        def traverse_system_recurs(system, atom_labels):
+            subsystems = system.get("model_system")
+            set_branch_composition(system, subsystems, atom_labels)
+            if subsystems:
+                for subsystem in subsystems:
+                    traverse_system_recurs(subsystem, atom_labels)
+
+        atom_labels = [atom.chemical_symbol for atom in self.cell[0].atoms_state]
```
JosePizarro3: I think you need some error handling here, to consider if `self.cell` exists and if so, if `atoms_state` exist. Something like:
```python
atom_labels = [atom.chemical_symbol for atom in self.cell[0].atoms_state if self.cell is not None and self.cell.atoms_state is not None]
```

src/nomad_simulations/model_system.py
```diff
@@ -967,6 +977,25 @@ def normalize(self, archive, logger) -> None:
...
+        traverse_system_recurs(self, atom_labels)
```
JosePizarro3: If moved to its own class method, you will have an advantage on having here `self.resolve_composition_formula(atom_labels)`, which is more nice to read.

tests/test_model_system.py
```diff
@@ -439,3 +441,75 @@ def test_normalize(self):
         assert np.isclose(model_system.elemental_composition[0].atomic_fraction, 2 / 3)
         assert model_system.elemental_composition[1].element == 'O'
         assert np.isclose(model_system.elemental_composition[1].atomic_fraction, 1 / 3)
+
+
+    @pytest.mark.parametrize(
+        'mol_label_list, n_mol_list, atom_labels_list, composition_formula_list',
```
JosePizarro3: I wondered whether you can test this a bit differently, if you decide to move everything to its own method class `resolve_composition_formula`. In that case, you can control much more when to return none, and when to return the composition formula.

If so, the testing should change to handle this "false testing", i.e., that when the conditions you are looking for do not apply, then the method returns None.

An example (this params are for seeing whether normalization of `KSpace` would return None, compute, or return the reciprocal_lattice_vectors): https://github.com/nomad-coe/nomad-schema-plugin-simulation-data/blob/develop/tests/test_numerical_settings.py#L51

JosePizarro3 (commented):
Very good, just some minor reorganization comments.

I am mainly interested on moving the couple of functions you have under `normalize()` directly as class methods and combine them into something called `resolve_composition_formula`. Still, you will need to define a specific function inside there to do the recursion. 

I will also change the `.get('...')` methods for our typical style to resolve quantities and sub-sections, e.g., `system.model_system` instead of `system.get('model_system')`. I think the second would return a dictionary, but I'd might be wrong.

Let me know if you need help or something is not clear.

src/nomad_simulations/model_system.py
JFRudzinski: Why do we need to check `if self.composition_formula is None:`, I guess then the parser can explicitly set this quantity if we want?

JFRudzinski:
@JosePizarro3  In terms of `system.model_system` instead of `system.get('model_system')`, they return the same thing if `system.model_system` exists. However, the .get() returns None if it doesn't exist, whereas the former breaks with an error. I guess it is somehow guaranteed that these archive sections are populated with all their possible attributes, but I am just used to using .get() to be safe. 

I will change it, just wanted to clarify.

JosePizarro3:
> @JosePizarro3 In terms of `system.model_system` instead of `system.get('model_system')`, they return the same thing if `system.model_system` exists. However, the .get() returns None if it doesn't exist, whereas the former breaks with an error. I guess it is somehow guaranteed that these archive sections are populated with all their possible attributes, but I am just used to using .get() to be safe.
> 
> I will change it, just wanted to clarify.

Yeah, this is actually kind of annoying. @TLCFEM @ladinesa could this be implemented or is there any reason why we don't want `system.model_system` in the example above to return None?

src/nomad_simulations/model_system.py
JFRudzinski: Actually, what do you think about something like this:
```python
subsystem_labels = [subsystem.get("branch_label", "Unknown") for subsystem in subsystems]
```

This way you can see if branches exist even if they don't have labels

JosePizarro3: Yeah, that's another option. Then I agree, we can use `.get()` in this case to indicate a default value `"Unknown"`.

src/nomad_simulations/model_system.py
JFRudzinski: Also, the composition_formula is set for each branch and index of the tree individually, so it's much easier to set them within `traverse_system_recurs`....

Oh wait, I just realized that there may be some confusion or a problem here. I was sort of assuming that only branch_depth = 0 would be ever marked as representative. That's why I am traversing the entire tree to set the composition formula at each branch level. 

I could instead resolve the composition formula for all model systems, representative or not, then I don't need to have any recursive function, I just look at the children of the current branch. The difficulty there though is that for the bottom of the hierarchy I need to go back up the tree to find the list of atom labels :thinking:

ladinesa:
> > @JosePizarro3 In terms of `system.model_system` instead of `system.get('model_system')`, they return the same thing if `system.model_system` exists. However, the .get() returns None if it doesn't exist, whereas the former breaks with an error. I guess it is somehow guaranteed that these archive sections are populated with all their possible attributes, but I am just used to using .get() to be safe.
> > I will change it, just wanted to clarify.
> 
> Yeah, this is actually kind of annoying. @TLCFEM @ladinesa could this be implemented or is there any reason why we don't want `system.model_system` in the example above to return None?

model_system since it is repeating subsection should return an empty list right?

JFRudzinski:
> > > @JosePizarro3 In terms of `system.model_system` instead of `system.get('model_system')`, they return the same thing if `system.model_system` exists. However, the .get() returns None if it doesn't exist, whereas the former breaks with an error. I guess it is somehow guaranteed that these archive sections are populated with all their possible attributes, but I am just used to using .get() to be safe.
> > > I will change it, just wanted to clarify.
> > 
> > 
> > Yeah, this is actually kind of annoying. @TLCFEM @ladinesa could this be implemented or is there any reason why we don't want `system.model_system` in the example above to return None?
> 
> model_system since it is repeating subsection should return an empty list right?

I actually wasn't saying it doesn't. I was asking if all metainfo sections and attributes are automatically populated with None (or empty lists) so that you don't ever need to do .get() because you won't get an error. Is that true? At what point are they populated, at instantiation?

ladinesa:
> > > > @JosePizarro3 In terms of `system.model_system` instead of `system.get('model_system')`, they return the same thing if `system.model_system` exists. However, the .get() returns None if it doesn't exist, whereas the former breaks with an error. I guess it is somehow guaranteed that these archive sections are populated with all their possible attributes, but I am just used to using .get() to be safe.
> > > > I will change it, just wanted to clarify.
> > > 
> > > 
> > > Yeah, this is actually kind of annoying. @TLCFEM @ladinesa could this be implemented or is there any reason why we don't want `system.model_system` in the example above to return None?
> > 
> > 
> > model_system since it is repeating subsection should return an empty list right?
> 
> I actually wasn't saying it doesn't. I was asking if all metainfo sections and attributes are automatically populated with None (or empty lists) so that you don't ever need to do .get() because you won't get an error. Is that true? At what point are they populated, at instantiation?

yes it should be the case, I have not encoutered a case where I need to use .get

src/nomad_simulations/model_system.py
JosePizarro3: This is always the recurrent doubt I have (and that by good timing, @ladinesa just pointed out the contrary on another pr #52, he **does not** want to overwrite parsed data).

The condition of "is not None" is just because for me parsed data has priority, but maybe for you normalized one has it. We should do a decision so that we don't have conflicting parsing/normalization, and we should do it soon.

JFRudzinski: Should we maybe add a warning that the composition formula was set by the user? Or maybe it's too annoying since it will get triggered a million times.

These warnings don't have any specific information about the instance, such that repeating warning is not really useful to 99% of users. @ladinesa is there any way to execute a warning such that duplicate warnings in the logger are ignored?

tests/test_model_system.py
JFRudzinski: I definitely agree that testing can be improved, thanks for the tips.

I am a bit unsure atm how to proceed, because I do not know how to deal with the recursive setting of the formula and the idea of returning the formula at each level instead of setting it. This is what I wrote above about it but not sure if you saw:
"""
I was assuming that only branch_depth = 0 would be ever marked as representative. That's why I am traversing the entire tree to set the composition formula at each branch level.

I could instead resolve the composition formula for all model systems, representative or not, then I don't need to have any recursive function, I just look at the children of the current branch. The difficulty there though is that for the bottom of the hierarchy I need to go back up the tree to find the list of atom labels
"""

Do you think I should normalize only for representative systems and traverse the tree downwards, or normalize all systems, but have to traverse to the top of the tree for any of the bottom nodes?

JFRudzinski: Ok, I moved the normalization up to Simulation() as you suggested. I think this is nicer. I need to still work on the tests though. I will have to come back to that next week.

JosePizarro3: Ok, cool. Let me know when you need another review. I am resolving all conversations 👍🏻

JFRudzinski:
@JosePizarro3 This is ready for another review. No rush though, you can leave it till after the project meeting if you don't have time.

I did make some improvements to the testing, I know it's not exactly what you had in mind, and I still feel that it is not ideal in the sense that the generation of the hierarchy is sort of complicated (I guess I could move this to some template generator?), but I am not exactly sure an alternative.

That being said, I think the functionality is much better now in the sense that I actually test a bunch of cases where quantities might be missing, and I tried to document this in the description.

src/nomad_simulations/general.py
```diff
@@ -177,6 +179,31 @@ def _set_system_branch_depth(
             system_child.branch_depth = branch_depth + 1
             self._set_system_branch_depth(system_child, branch_depth + 1)
 
+    def resolve_composition_formula(
+        self, system_parent: ModelSystem, logger: BoundLogger
+    ) -> None:
+        """
```
JosePizarro3, line 189: There is a very nice extension in VSCode (autoDocstring) that, together with Github copilot, essentially generates the docstring of these functions for you. Maybe you want to check this out

src/nomad_simulations/general.py
```diff
@@ -177,6 +179,31 @@ def _set_system_branch_depth(
...
+        """
+        def set_branch_composition(system: ModelSystem, subsystems: List[ModelSystem], atom_labels: List[str]) -> None:
```
JosePizarro3: you can also add docstrings for this and the next function to explain usage

src/nomad_simulations/general.py
```diff
@@ -177,6 +179,31 @@ def _set_system_branch_depth(
...
+            if not subsystems:
+                atom_indices = system.atom_indices if system.atom_indices is not None else []
+                subsystem_labels = [np.array(atom_labels)[atom_indices]] if atom_labels else ['Unknown' for atom in range(len(atom_indices))]
+            else:
+                subsystem_labels = [subsystem.branch_label if subsystem.branch_label is not None else "Unknown" for subsystem in subsystems]
+            if system.composition_formula is None:
+                system.composition_formula = get_composition(subsystem_labels)
+
+        def traverse_system_recurs(system, atom_labels):
```
JosePizarro3:
```python
def traverse_system(system: ModelSystem, atom_labels: List[str]) -> None:
```

src/nomad_simulations/general.py
```diff
@@ -192,8 +219,12 @@ def normalize(self, archive, logger) -> None:
...
-            if len(system_parents.model_system) == 0:
+        for system_parent in self.model_system:
+            system_parent.branch_depth = 0
+            if len(system_parent.model_system) == 0:
                 continue
-            self._set_system_branch_depth(system_parents)
+            self._set_system_branch_depth(system_parent)
+
+            if is_not_representative(system_parent, logger):
+                return
```
JosePizarro3: Shouldn't this be `continue`? In case we have more normalization afterwards

src/nomad_simulations/utils/utils.py
```diff
@@ -126,6 +127,15 @@ def is_not_representative(model_system, logger: BoundLogger):
         logger.warning('The `ModelSystem` is empty.')
         return None
     if not model_system.is_representative:
-        logger.warning('The `ModelSystem` was not found to be representative.')
         return True
     return False
+
+# TODO Either update nomad.atomutils function and remove this one, or remove the one in atomutils if we prefer it here only
```
JosePizarro3: I think it makes more sense to have it here, so you can remove it eventually from nomad.atomutils (that's the idea, that we transfer also these utility functions).

My suggestion: change slightly this TODO to say `# TODO remove function in nomad.atomutils`

tests/test_model_system.py
```diff
@@ -439,3 +442,242 @@ def test_normalize(self):
...
+        mol_label_list: List[str],
+        n_mol_list: List[int],
+        atom_labels_list: List[str],
+        composition_formula_list: List[str],
+        custom_formulas: List[str]
+    ):
+        """
+        Test the `ModelSystem` normalization of 'composition_formula' for atoms and molecules.
+
+        Description of test parameters:
```
JosePizarro3: I was following the idea of not describing the inputs in testing, but it might be a good idea. I suggest then autoDocstring as I said

tests/test_model_system.py
```diff
@@ -439,3 +442,242 @@ def test_normalize(self):
...
+            ), # pure system
+            (
+                ['H20', 'Methane'],
+                [5, 2],
+                [['H', 'O', 'O'], ['C', 'H', 'H', 'H', 'H']],
+                ['group_H20(1)group_Methane(1)', 'H20(5)', 'H(1)O(2)', 'H(1)O(2)', 'H(1)O(2)', 'H(1)O(2)', 'H(1)O(2)', 'Methane(2)', 'C(1)H(4)', 'C(1)H(4)']
+            ), # binary mixture
+        ],
+    )
+    def test_system_hierarchy_for_molecules(
```
JosePizarro3: I think this test function is duplicated, you can delete it as the next one is the good one

JosePizarro3 (approved):
Very good, just some final minor details that will make your life easier when writing docstrings for functions.

I really liked the testing, it made it very easy to understand the implementation, and I am surprised that this actually decreased the coverage... I have to ask the details about this package, it is very strange sometimes.

src/nomad_simulations/general.py
JFRudzinski: I was making a divide, so that all normalization below was for representative systems only, but maybe that's not idea, otherwise I can do:
```python
  if not is_not_representative(system_parent, logger):
      self.resolve_composition_formula(system_parent, logger)
```
I mean in that case I guess I would make a `is_representative` that returns the negative of the other function, but you get the point. Which do you prefer?

tests/test_model_system.py
JFRudzinski: I'm fine if you don't want to have this, I just thought it might be nice to include in this case cause the testing function is a bit complicated

src/nomad_simulations/general.py
JosePizarro3: I think `continue` is safer, it will simply go back to the for loop into the next element. With the current `return`, imagine that the first ModelSystem is not representative, but the next ones are, so this return will screw the resolve_composition_formula call for the systems that are actually representative.

tests/test_model_system.py
JosePizarro3: Yes, it is very nice. Just the style we can combine it with other functions across the code

src/nomad_simulations/general.py
JFRudzinski: sorry, I somehow didn't realize this was in a loop, yes continue of course :facepalm:

JFRudzinski:
@JosePizarro3 Thanks for the tips, I think I have addressed everything. Let me know if you think we are ready to consider merging.

btw - have you applied ruff to this repo or do I need to do it manually? My auto-ruff formatting is not instigated for this repo, not sure if that is something that I need to address locally or something with the repo.

JosePizarro3:
> @JosePizarro3 Thanks for the tips, I think I have addressed everything. Let me know if you think we are ready to consider merging.
> 
> btw - have you applied ruff to this repo or do I need to do it manually? My auto-ruff formatting is not instigated for this repo, not sure if that is something that I need to address locally or something with the repo.

So in the pyproject.toml `ruff` is added, so autoformatting should work. You also have a couple of files with conflicts in develop, maybe you need to rebase.

JFRudzinski:
> > @JosePizarro3 Thanks for the tips, I think I have addressed everything. Let me know if you think we are ready to consider merging.
> > btw - have you applied ruff to this repo or do I need to do it manually? My auto-ruff formatting is not instigated for this repo, not sure if that is something that I need to address locally or something with the repo.
> 
> So in the pyproject.toml `ruff` is added, so autoformatting should work. You also have a couple of files with conflicts in develop, maybe you need to rebase.

ok rebased and applied ruff manually (not sure what's going on there)...looks good?

JosePizarro3:
Yeah, please, merge. 

Maybe there is something off with your virtual environment and the rules applied (I am guessing some conflict on that direction). We can check it out in the office this week, it should be automatically applied when saving a file.

JFRudzinski:
> Yeah, please, merge.
> 
> Maybe there is something off with your virtual environment and the rules applied (I am guessing some conflict on that direction). We can check it out in the office this week, it should be automatically applied when saving a file.

cool, thanks a lot for all your help with this!

PR #81, opened 2024-06-03
Adapting the new entrypoints plugin mechanism and changing name of repo and package

### LINKED ISSUE #64: Adapting the new entrypoints plugin mechanism and changing name of repo and package

JosePizarro3:


JosePizarro3:
Closing this in favor of #88

PR #84, opened 2024-06-05
Add unit testing

### LINKED ISSUE #9: Adding unit testing

JosePizarro3:
Fixing, improving, and adding new testing for all the sections and methods defined so far.

I am also explicitly defining inputs to methods now, to make it more clear when calling them.

Closes #9


JosePizarro3:
@JFRudzinski @ndaelman-hu @Bernadette-Mohr 

Please, be aware of this. No need to do a review of this.

Important: I created a script to generate a coverage report with some comprehensive coverage of each of the `src/` modules. You can run `./scripts/generate_coverage_txt.sh` and this will regenerate the artifact `coverage.txt` in the root folder.

I think the content of that file is self-explanatory, but I asked Ahmed Ilyas to provide more support on the current Coverage, as the reports in PR requests are pretty cryptic.

Issue #28, opened 2024-04-02
Add MD relevant outputs 


JFRudzinski:
- [ ] Migrate and harmonize MD workflow schema
- [ ] Add additional outputs
- [ ] Expand FE Calc schema

PR #87, opened 2024-06-13
28 add md relevant outputs

JFRudzinski:
new PR for #28 since the first one is somehow broken

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ share/python-wheels/
 .installed.cfg
 *.egg
 MANIFEST
+_version.py
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
--- a/README.md
+++ b/README.md
@@ -14,9 +14,14 @@ This is a plugin for [NOMAD](https://nomad-lab.eu) which contains the base secti
 
 ## Getting started
 
-### Install the dependencies
+`nomad-simulations` can be installed as a PyPI package using `pip`. We require features from the `nomad-lab` package which are not publicly available in PyPI, so an extra flag `--index-url` needs to be specified when pip installing this package:
+```sh
+pip install nomad-simulations --index-url https://gitlab.mpcdf.mpg.de/api/v4/projects/2187/packages/pypi/simple
+```
 
-Clone the project and in the workspace folder, create a virtual environment (note this project uses Python 3.9):
+## Development
+
+If you want to develop locally this package, clone the project and in the workspace folder, create a virtual environment (note this project uses Python 3.9):
 
 ```sh
 git clone https://github.com/nomad-coe/nomad-simulations.git
@@ -118,25 +123,6 @@ If using VSCode, you can add the following snippet to your `.vscode/launch.json`
 where `${workspaceFolder}` refers to the NOMAD root.
 
 The settings configuration file `.vscode/settings.json` performs automatically applies the linting upon saving the file progress.
-
-### Build the python package
-
-The `pyproject.toml` file contains everything that is necessary to turn the project
-into a pip installable python package. Run the python build tool to create a package distribution:
-
-```
-uv pip install build
-python -m build --sdist
-```
-
-You can install the package with pip:
-
-```
-uv pip install dist/nomad-simulations-0.1.0
-```
-
-Read more about python packages, `pyproject.toml`, and how to upload packages to PyPI
-on the [PyPI documentation](https://packaging.python.org/en/latest/tutorials/packaging-projects/).
 
 ### Documentation on Github pages
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -128,6 +128,9 @@ package-dir = { "" = "src" }
 [tool.setuptools.packages.find]
 where = ["src"]
 
+[tool.setuptools_scm]
+write_to = "src/nomad_simulations/_version.py"
+
 [project.entry-points.'nomad.plugin']
 
 nomad_simulations_plugin = "nomad_simulations.schema_packages:nomad_simulations_plugin"
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,0 @@
-from setuptools import setup
-
-if __name__ == '__main__':
-    setup()
--- a/src/nomad_simulations/__init__.py
+++ b/src/nomad_simulations/__init__.py
@@ -1,0 +1,4 @@
+try:
+    from ._version import version as __version__
+except ImportError:
+    __version__ = ''


JFRudzinski:
@JosePizarro3 @ndaelman-hu Could you do a preliminary brief check of the structure within `energies.py`. My main question at the moment is how we deal with the organization of "sub-contributions".

Right now, all energies fall under `total_energies.contributions`. On the classical side, I have made this a flat list, e.g., ElectrostaticEnergy, ElectrostaticShortRangeEnergy, ElectrostaticLongRangeEnergy all fall directly under TotalEnergy, even though 2 are subsets of the third. I find it a bit easier this way since it's not a given that users/codes always store all of these options.

Will this work for the quantum case? Looking at the old schema (old quantities that I have not yet converted are commented out in the file) it seems like there are a lot of sub-structures in terms of the energies. How much are these used? Can you conceive a good functionality with flat lists?

JosePizarro3:
> @JosePizarro3 @ndaelman-hu Could you do a preliminary brief check of the structure within `energies.py`. My main question at the moment is how we deal with the organization of "sub-contributions".
> 
> Right now, all energies fall under `total_energies.contributions`. On the classical side, I have made this a flat list, e.g., ElectrostaticEnergy, ElectrostaticShortRangeEnergy, ElectrostaticLongRangeEnergy all fall directly under TotalEnergy, even though 2 are subsets of the third. I find it a bit easier this way since it's not a given that users/codes always store all of these options.
> 
> Will this work for the quantum case? Looking at the old schema (old quantities that I have not yet converted are commented out in the file) it seems like there are a lot of sub-structures in terms of the energies. How much are these used? Can you conceive a good functionality with flat lists?

Ok, let me check. I'd like to see how much of the quantum energies were restructured into `ElectronicEigenvalues`. I think there is some inconsistency there (from the legacy schema), as I would consider `Energy` to be a global quantity, not refering to the eigenvalues, and it might be that in the run schema it was not like this.

src/nomad_simulations/properties/energies.py
```diff
@@ -77,3 +78,502 @@ def __init__(
...
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
+
+        if not self.type:
+            self.type == 'quantum'
+        elif self.type != 'quantum':
+            logger.error(
+                f"Misidentified type for quantum energy."
+            )
```
ndaelman-hu: Tbh, I'm not sure what would constitute the boundary between _classical_ and _quantum_ energies.
Every classical energy has a quantum variant. There are some unique quantum phenomena that don't have a classical counterpart, but it's so few that I would not distinguish at this level.

The only example I can think of now is _exchange energy_.

src/nomad_simulations/properties/energies.py
```diff
@@ -77,3 +78,502 @@ def __init__(
...
+        super().normalize(archive, logger)
+
+
+class DihedralEnergy(ClassicalEnergy):
+    """
+    Physical property section describing contributions to the potential energy from dihedral interactions of a (sub)system.
+    """
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
```
ndaelman-hu: I'm confused now, these look more like terms of the _potential_ rather than _energy_ contributions.
Instead, I'd expect _rotational_ and _vibrational_ energies (especially useful in light of the virial theorem), as subtypes of kinetic energy. These modes would classically be described in function of the (dihedral) angle.

src/nomad_simulations/properties/energies.py
```diff
@@ -77,3 +78,502 @@ def __init__(
...
+
+
+class ExternalEnergy(ClassicalEnergy):
+    """
+    Physical property section describing contributions to the potential energy from external interactions of a (sub)system.
+    """
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
```
ndaelman-hu: I would consider this more of an energy flux, so the dimensionality is slightly different.

src/nomad_simulations/properties/energies.py
```diff
@@ -77,3 +78,502 @@ def __init__(
...
+class ZeroPointEnergy(QuantumEnergy):
+    """
+    Physical property section describing the zero-point vibrational energy of a (sub)system,
+    calculated using the method described in zero_point_method.
+    """
+    # ! Someone check this description!
+    # ? Do we really want to specify the method here? This can't be user-defined?
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
```
ndaelman-hu: Quantum-specific subtype of vibrational energy.

src/nomad_simulations/properties/energies.py
```diff
@@ -77,3 +78,502 @@ def __init__(
...
+
+class ZeroTemperatureEnergy(QuantumEnergy):
+    """
+    Physical property section describing the total energy of a (sub)system extrapolated to $T=0$, based on a free-electron gas argument.
+    """
+    # ! Someone check this description!
+    # ? Do we really want to specify the method here? This can't be user-defined?
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
```
ndaelman-hu: I think this should be introduced in the context of a workflow. Would remove it here.

src/nomad_simulations/properties/energies.py
```diff
@@ -77,3 +78,502 @@ def __init__(
...
+#! allowed_contributions = ['PotKin'] Not sure how to deal with sub-contributions... - I lean towards keeping a flat list but I am not sure of all the usages
+
+
+class ElectronicKineticEnergy(QuantumEnergy):
+    """
+    Physical property section describing the electronic kinetic energy of a (sub)system.
+    """
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
```
ndaelman-hu: Okay, this is an interesting take. In Solid State Physics, the energies below also typically refer to electrons, but technically may be defined for any part of the system.
I would think of a way for distinguishing the kind of energy contribution from the system it applies to. That will keep the wildgrowth manageable.

src/nomad_simulations/properties/energies.py
```diff
@@ -77,3 +78,502 @@ def __init__(
...
+    i.e., the integral of the first order derivative of the functional
+        stored in XC_functional (integral of v_xc*electron_density), i.e., the component
+        of XC that is in the sum of the eigenvalues. Value associated with the
+        configuration, should be the most converged value.          |
+    """
+    # ! Someone check this description!
+    # ? Do we really want to specify the method here? This can't be user-defined?
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
```
ndaelman-hu: This is the XC potential, something distinct from the energy. This is already defined in the DFT name.

src/nomad_simulations/properties/energies.py
```diff
@@ -77,3 +78,502 @@ def __init__(
...
+
+
+# ? XCCorrelationEnergy?
```
ndaelman-hu: C in XC already stands for Correlation, so this would be a repition.

src/nomad_simulations/properties/energies.py
```diff
@@ -77,3 +78,502 @@ def __init__(
...
+        description="""
+        The value of the correlation energy.
+        """,
+    )
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
+
+
+# ? XCExchangeEnergy?
```
ndaelman-hu: X in XC already stands for Exchange, so this would be a repitition.

src/nomad_simulations/properties/energies.py
```diff
@@ -77,3 +78,502 @@ def __init__(
...
+        super().normalize(archive, logger)
+
+
+class NuclearRepulsionEnergy(QuantumEnergy):
+    """
+    Physical property section describing the nuclear-nuclear repulsion energy of a (sub)system.
+    """
+
+    def normalize(self, archive, logger) -> None:
+        super().normalize(archive, logger)
```
ndaelman-hu: This is just electrostatics for the nuclei. See my take above.

src/nomad_simulations/properties/energies.py
```diff
@@ -77,3 +78,502 @@ def __init__(
...
+    # )
+
+    # van_der_waals = SubSection(
+    #     sub_section=EnergyEntry.m_def,
+    #     description="""
+    #     Contains the value and information regarding the Van der Waals energy. A multiple
+    #     occurence is expected when more than one van der Waals methods are defined. The
+    #     van der Waals kind should be specified in Energy.kind
+    #     """,
+    # )
```
ndaelman-hu: This can both refer to a set of methodologies and the Casimir effect.

src/nomad_simulations/properties/energies.py
```diff
@@ -77,3 +78,502 @@ def __init__(
...
+
+    # double_counting = SubSection(
+    #     sub_section=EnergyEntry.m_def,
+    #     categories=[FastAccess],
+    #     description="""
+    #     Double counting correction when performing Hubbard model calculations.
+    #     """,
+    # )
```
ndaelman-hu: While not a physical energy, per se, it is a term in the Hamiltonian used for corrections. This begs the question: do we want to distinguish Hamiltonian terms or physical phenomena?

src/nomad_simulations/properties/energies.py
```diff
@@ -77,3 +78,502 @@ def __init__(
...
+
+    # lowest_unoccupied = Quantity(
+    #     type=np.dtype(np.float64),
+    #     unit='joule',
+    #     shape=[],
+    #     description="""
+    #     The lowest unoccupied energy.
+    #     """,
+    # )
```
ndaelman-hu: These are already accommodated else where in the schema. May be removed here.

src/nomad_simulations/properties/energies.py
```diff
@@ -77,3 +78,502 @@ def __init__(
...
+
+    # # TODO I suggest ewald is moved to "long range" under electrostatic->energyentry, unless there is some other usage I am misunderstanding
+    # ewald = SubSection(
+    #     sub_section=EnergyEntry.m_def,
+    #     description="""
+    #     Contains the value and information regarding the Ewald energy.
+    #     """,
+    # )
```
ndaelman-hu: Subtype of electrostatic energy.

ndaelman-hu (commented):
Just some quick takes, not a thorough review.

My main impression is that there's some confusion between classical vs quantum and potential vs energy. Regarding the former, it's very simple: all classical terms have a quantum counterpart. I think it makes more sense to distinguish the systems for which we specify the energy, e.g. electrons, nuclei, both, external.

Also note that magnetism is completely left out.

JosePizarro3:
> @JosePizarro3 @ndaelman-hu Could you do a preliminary brief check of the structure within `energies.py`. My main question at the moment is how we deal with the organization of "sub-contributions".

Ok, I took a look on the code. I can send later on some notes on the DFT total energy, and how this could be corrected in other approaches. I also agree with what Nathan said on the confusion of `TotalEnergy` vs `CoulombPotential`; I think this can be "easily" solved writing down the DFT  equations for `TotalEnergy`, but also the ones for the `CoulombPotential`. `hubbard` or `double_counting` could be left out for a later feature in which we include other contributions

@JFRudzinski your approach is good, so here I am just giving an alternative idea, also for @ndaelman-hu @Bernadette-Mohr about how to deal with contributions to PhysicalProperties.

Regarding class definitions, these changes show how annoying is to define multiple contributions using inheritance. In my opinion, I think contributions could be simplified to an abstract class and some identifier (a string/MEnum) used both for 
the `TotalEnergy` and the sub_section that repeats. Like we have classical and quantum, I decided to keep them separated, but you could also abstract this to a string/MEnum. My reasoning on doing it this ways is that:
- Contributions do not have specialized quantities or normalizations.
- It gets complicated to maintain (400 lines of code, 20 classes VS. 60 lines, 3-4 classes)

My proposal would be thus something simpler, a string or MEnum (see [the abstract class `DOSProfile` and how I handled `ElectronicDOS(DOSProfile)` and its contributions](https://github.com/nomad-coe/nomad-simulations/blob/develop/src/nomad_simulations/schema_packages/properties/spectral_profile.py#L101)) to identify each contribution:
```python

class BaseTotalEnergy(PhysicalProperty):
    """
    A base section class defining the abstract concept of `TotalEnergy`. This section is useful when defining contribution to the `TotalEnergy`, as well as the `TotalEnergy` section itself.
    """

    value = Quantity(
        type=np.float64,
        unit='joule',
        description="""
        The value of the energy.
        """,
    )

class ClassicalTotalEnergy(BaseTotalEnergy):
    """
    A base section defining the concept of classical `TotalEnergy`. This section is useful for categorizing TotalEnergy in terms of classical or quantum approaches.
    """

    name = Quantity(type=MEnum('kinetic', 'potential', 'intermolecular'...),)  # more strings here

class QuantumTotalEnergy(BaseTotalEnergy):
    """
    A base section defining the concept of classical `TotalEnergy`. This section is useful for categorizing TotalEnergy in terms of classical or quantum approaches.
    """

    name = Quantity(type=MEnum('kinetic', 'Hartree', 'xc', ...),)  # more strings here

class TotalEnergy(BaseTotalEnergy):
    """
    ... TAKE DESCRIPTION FROM TAXONOMY
    """

    iri = '...'

    classical_contributions = SubSection(
        sub_section=ClassicalTotalEnergy.m_def, repeats=True
    )

    quantum_contributions = SubSection(
        sub_section=QuantumTotalEnergy.m_def, repeats=True
    )

```

The question here is also, at the end of the day, how important are these contributions? I think we can ease the taxonomy by not overclassifying these kind of things. 

Ah, and I edit to say that the description of the property should then reflect the contributions and what are these. I think we should bring this last point (if agreed) to the taxonomy TF.

src/nomad_simulations/properties/energies.py
JFRudzinski: I definitely understand your point. My reasons for proposing to distinguish between the two were:

1. To help clarify the origin of certain energy contributions in the case of a hybrid model. E.g., you may have a QM/MM model and want to store a particular contribution to the total energy from the QM and MM parts of your model separately.
    - I will argue against myself here and say that the proper way to do this would be with the method_ref, which is true, but this is a bit opaque for a user. Having the `type` quantity makes it easier to see quickly what level of model one used to get this output.
2. There are certain energy contributions, e.g., kinetic energy, where the standard method of calculation is fundamentally different from a classical or a quantum model.
    - Again maybe it is not a good idea to conflate the output and the method too much. In fact this goes against a bunch of my comments where I asked if we should take the reference to a specific calculation method out of the description. So, from this perspective I am on the fence.
3. There are many examples where there is not a quantum counterpart to the classical energy. In addition to a few quantum phenomena as you mentioned, in classical simulations it is common to store the contribution to the total energy (usually per particle) from a particular term of the force field. 
    - E.g., you may want to know the total energy on a particle in a configuration coming from some dihedral potential only. 
    - In the previous schema it became unclear at times the usage of certain contributions. This is partially due to the inclusion of specific methodology for calculation within the description, but also because I think in some cases people had a certain class of models in mind. So, I was trying to prevent this a little. 
    - I think this is why you thought there is a confusion between energy and potential. The potential is indeed stored within the FF section of method. But the corresponding energy is the evaluation of that potential or set of potentials for a particular configuration.


All that being said, I definitely am not set on this distinction. Curious to hear your rebuttal given the additional information, and also the thoughts of @Bernadette-Mohr and @JosePizarro3

JosePizarro3:
@ndaelman-hu I found this video which might help as a first approach: https://youtu.be/5mxVR1ikQQo?si=Cfxzs0FlKyqrkFuv&t=1424

But I am sure there are better notes somewhere, I just casually browsed Youtube.

src/nomad_simulations/properties/energies.py
JFRudzinski: Just to be clear I paste my explanation from above:

In classical simulations it is common to store the contribution to the total energy (usually per particle) from a particular term of the force field. E.g., you may want to know the total energy on a particle in a configuration coming from some dihedral potential only. I think this is why you thought there is a confusion between energy and potential. The potential is indeed stored within the FF section of method. But the corresponding energy is the evaluation of that potential or set of potentials for a particular configuration.

src/nomad_simulations/properties/energies.py
JFRudzinski: This term is similar to the contributions coming from specific terms in the force field as defined above. Sometimes you define an external force within the force field (i.e., not linked to a particular atom type + config). Then you may want to again know the contribution to the total energy (e.g., per particle) from this external potential. Perhaps it should be renamed?

So, it's not a flux. Maybe there is a different energy contribution that you are describing?

src/nomad_simulations/properties/energies.py
JFRudzinski: Can you be more specific what you suggest changing in this case? I need your guidance for the quantum contributions (maybe it needs to wait until we figure out whether or not to distinguish Quant/Class though)

src/nomad_simulations/properties/energies.py
JFRudzinski: Just repeating myself from above: We should definitely NOT store the Hamiltonian terms here. That goes in the method. However, we may store the evaluation of the Hamiltonian for particular configurations of particles. This should be more clear though for sure.

src/nomad_simulations/properties/energies.py
JFRudzinski: This is a major question that I have: How should we deal with sub-contributions? One option is to keep a flat list (flat inside total_energy.contributions at least), and somehow distinguish sub contributions with names and descriptions.

The other option is to nest. If we do this, I think we should be restrictive about the nesting, i.e., specify energy sub-contributions that are relevant for a particular contribution. We should not allow free recursive nesting or sub-contributions that don't make sense theoretically/methodologically

JFRudzinski:
> ```python
> ClassicalTotalEnergy
> ```

I definitely like the simplicity of this approach. I guess my only concern is losing the descriptions of the contributions. Could these actually simply be set during normalization? Since there will be no description coming from the taxonomy for these contributions? 

Oh ok, I see now that you suggested adding these to the taxonomy. This would work for physical contributions but not really specific model contributions (see my descriptions about evaluation of terms in the Hamiltonian above). I think for the later we need to add descriptions within the schema.

(tbh I also do not know how important these contributions are in general. But I think we still want to have some sort of robust scheme for dealing with them because they will be used at least in some niche cases)

JFRudzinski:
Sorry if this is confusing, but the easiest way to apply the new structure was to simply manually edit/add the files starting from the head of develop. So, this PR got killed in the process. We can keep discussing the energies here to close out the open discussion. But after that I created a new PR: [https://github.com/nomad-coe/nomad-simulations/pull/91](https://github.com/nomad-coe/nomad-simulations/pull/91)

JosePizarro3:
> Sorry if this is confusing, but the easiest way to apply the new structure was to simply manually edit/add the files starting from the head of develop. So, this PR got killed in the process. We can keep discussing the energies here to close out the open discussion. But after that I created a new PR: #91

Yeah sorry, indeed it might be better to do that.

JFRudzinski:
> > Sorry if this is confusing, but the easiest way to apply the new structure was to simply manually edit/add the files starting from the head of develop. So, this PR got killed in the process. We can keep discussing the energies here to close out the open discussion. But after that I created a new PR: #91
> 
> Yeah sorry, indeed it might be better to do that.

it was actually quite easy to amend the structure, it's just that I lost the connection to my other changes, but tbh i don't really care since all this is very preliminary anyway

src/nomad_simulations/properties/energies.py
ndaelman-hu: > the contribution to the total energy (usually per particle) from a particular term of the force field

Yes, in QM we do the same with the Hamiltonian (which may encode more than just the potential). Up to a certain extent, QM codes also report the energy decomposition as such, though not always. The crux is that this is model-specific, by definition. I.e. you're generating a semantic language to describe all physical models. This may be fine, but should then tie back to the `method` section.

What I was suggesting is to focus instead on _physical processes_. I think those are more interoperable and of general interest i. E.g. _kinetic energy_ which breaks apart into _translational_, _rotational_, and _vibrational_ is relevant for the virial theorem, entropy, etc. I would therefore make these into subsections.

> total energy on a particle in a configuration

Well, this is firstly partitioned by the structure, so you should decide on a mapping there. The rest is method specific data, so I would see to tie to the method definition. Still, one can also analyze (but maybe not easily extract) the dihedral contribution and others in a rotational and vibrational mode.

src/nomad_simulations/properties/energies.py
ndaelman-hu: Regarding your points:

1. QM/MM are indeed a mapping between methods and structure. How would you, using the distinction above, distinguish the energy contributions in QM/QM/MM (e.g. CCSD/DFT/Amber) or QM/MM/MM?
2. ... QM in the limit of heavy particles converges to CM, **without exception**. The main difference in maths is that variables and functions in CM become functions (in the Hilbert space) and operators in QM. That is to say, the same physics is presented in higher-order structures. Some new physical phenomena emerge, but all the old ones remain. More importantly, **why does the mathematical representation matter here?** We're just getting scalars (maybe tensors) in our data.
3. Well, that's just plain wrong. At best, these communities are not interested in these contributions, but that's another story. As for your example: the dihedral potential is captured by DFT, just not named as such. ML potentials actually use DFT data to reformulate the physics in terms of the (local) neighborhood.

I agree with your sentiment that a lot of these contributions are model-specific, and there's value in extracting those. Then we need a system to annotate the method terms. I would mentally distinguish it from the contributions of physical phenomena, which are not subordinate to the method used to simulate them. Let's reserve the energies here for what experimentalists could also request.

src/nomad_simulations/properties/energies.py
ndaelman-hu: Well, it all depends on where you draw the boundary between your system and the outer world. If the latter interacts with the system, we can describe that as an energy flux, e.g.:

- a thermal bath: heat (a flux of kinetic and radiation energy) is being exchanged.
- an external electric or magnetic field: fluctuations enter the system as EM-waves, i.e. a flux of radiation energy.

Given these examples, we can represent the outer-world in simpler models, abstracting away a lot of complexity. However, I don't think most people consider them part of the system, so, in line with conservation of energy, they interact via a flux.

src/nomad_simulations/properties/energies.py
ndaelman-hu: The name is fine, I'd alter the supertype to `ZeroPointEnergy(VibrationalEnergy)`.
This is the lowest vibrational state, which has a non-zero contribution.

src/nomad_simulations/properties/energies.py
ndaelman-hu: > Just repeating myself from above: We should definitely NOT store the Hamiltonian terms here

I agree with you (see above), but I have the impression that you were doing that anyhow here.

src/nomad_simulations/properties/energies.py
ndaelman-hu: Or you can have both options at once if you use inheritance.

src/nomad_simulations/properties/energies.py
Bernadette-Mohr: @JFRudzinski: With "external force," do you mean something like, e.g., a biasing potential or a wall in your box? Maybe this makes what we're talking about a bit clearer to @ndaelman-hu.

ndaelman-hu: Ow, that makes more sense! Yes, these can't be modeled via flux, as they are more so numerical / statistical tricks instead of physical forces (well, the biasing potential, at least).

I would definitely seek to associate these approaches with the method itself.

src/nomad_simulations/properties/energies.py
Bernadette-Mohr: Ok, here's my unqualified opinion:

1. Making the different energy contributions storable/accessible is valid, in my opinion, either because they were relevant for the original study or because somebody is interested in that particular aspect when conducting secondary analysis.
2. While I agree with @ndaelman-hu that, theoretically, especially QM  and MM/MD describe the same physics, each approach has it's strengths and weaknesses.
- We should absolutely provide that information somewhere, but maybe more with methods than with outputs?
One use case that comes to mind is training data for an ML model. If you want to have any faith in your predictions, you should be interested in the underlying distribution or potential issues.
3. @JFRudzinski, I'm not sure why you think including the methodology makes things confusing here; streamlining what information we provide where will probably make things clearer, though.

Generally, it seems that the slightly different usage of "potential" vs. "energy" in the different methods is a source of confusion even among us, so I think we should try to prevent this with the users. I remember us discussing something like a "derived" flag, to differentiate quantities that are calculated and aspects that are directly part of or automatically generated by the method. Could something along this line help?

JFRudzinski:
@Bernadette-Mohr Thank you very much for your input, it is very valuable.

Nathan and I just had a meeting in an attempt to settle some aspects of the above discussions. Here is a short outline of our proposal:


```python
class Energy(PhysicalProperty):
    """
    Base class for all energy properties
    """

    value = Quantity(
        type=np.float64,
        unit='joule',
        description="""
        The value of the energy.
        """,
    )

class TotalEnergy(Energy):
    """
    Just the total energy
    """

class PropertyContribution(PhysicalProperty):
    """
    Base class for defining contribution link to some method. Will be used at least for energies and forces
    """
    method_contribution_ref = Quantity(
        type=MethodContribution,
        description="""
        """,
        a_eln=ELNAnnotation(component='ReferenceEditQuantity'),
    )

class EnergyContribution(PropertyContribution, Energy):
    """
    Energy property linked to method contribution
    """

    name = Quantity(type=str,)  # may be set explicitly or via method_contr_ref

```

Let me know if there are any comments or questions. My plan is to implement this structure today and then switch to #91 for another review

PR #93, opened 2024-06-27
Further abstraction in ModelMethod

### LINKED ISSUE #92: Further abstraction in ModelMethod
@ndaelman-hu @Bernadette-Mohr 

After talking with @JFRudzinski, it seems we should be more consistent on treating contributions of physical properties, more importantly when talking about models (either FF or TB). There are cases where physical property contributions (e.g., the total energy of a system) should have refs to specific terms in the Hamiltonian or model solved.

For this, I think it makes sense to add a new layer of abstraction for `ModelMethod`, and add a repeated subsection called `terms` under it. This will allow us to add a `method_ref` to the base class of physical properties/contributions so that it points either to the total Hamiltonian or to the specific part of the Hamiltonian.

JosePizarro3:
Here you have the changes. I took the opportunity and added `model_method_ref` to `Outputs` as well.

I think one reviewer is enough, but if any of you want (@ndaelman-hu @Bernadette-Mohr) let me know and I add you as reviewers.

Closes #92

### DIFF
--- a/src/nomad_simulations/schema_packages/model_method.py
+++ b/src/nomad_simulations/schema_packages/model_method.py
@@ -982,8 +982,6 @@ class BSE(ExcitedStateMethodology):
 
             H_BSE = H_diagonal + 2 * gx * Hx - gc * Hc
 
-        where gx, gc specifies the `kind`.
-
         Online resources for the theory:
         - http://exciting.wikidot.com/carbon-excited-states-from-bse#toc1
         - https://www.vasp.at/wiki/index.php/Bethe-Salpeter-equations_calculations
--- a/src/nomad_simulations/schema_packages/model_method.py
+++ b/src/nomad_simulations/schema_packages/model_method.py
@@ -23,10 +23,10 @@ import numpy as np
 
 from nomad.datamodel.data import ArchiveSection
 from nomad.datamodel.metainfo.annotations import ELNAnnotation
-from nomad.metainfo import MEnum, Quantity, SubSection
+from nomad.metainfo import MEnum, Quantity, SubSection, URL, Section
 
 if TYPE_CHECKING:
-    from nomad.metainfo import Section, Context
+    from nomad.metainfo import Context
     from nomad.datamodel.datamodel import EntryArchive
     from structlog.stdlib import BoundLogger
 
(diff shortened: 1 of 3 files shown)


src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -36,37 +36,50 @@
...
+    `ModelMethod`, while its contributions are defined in `ModelMethod.terms`.
+
+    Example: a custom model Hamiltonian containing two terms:
+        $H = H_{V_{1}(r)}+H_{V_{2}(r)}$
+    where $H_{V_{1}(r)}$ and $H_{V_{2}(r)}$ are the contributions of two different potentials written in
+    real space coordinates $r$. These potentials could be defined in terms of a combination of parameters
+    $(a_{1}, b_{1}, c_{1}...)$ for $V_{1}(r)$ and $(a_{2}, b_{2}, c_{2}...)$ for $V_{2}(r)$. If we name the
+    total Hamiltonian as `'FF1'`:
+        `ModelMethod.name = 'FF1'`
+        `ModelMethod.contributions = [BaseModelMethod(name='V1', parameters=[a1, b1, c1]), BaseModelMethod(name='V2', parameters=[a2, b2, c2])]`
```
JFRudzinski, line 54: from this example it is unclear to me if parameters will be defined at the level of BaseModelMethod() or only within subclasses?

I had done something like this before for Force Fields:
```python
class Potential(ArchiveSection):
    """
    Section containing information about an interaction potential.
    """

    type = Quantity(
        type=str,
        shape=[],
        description="""
        Specifies the type (i.e., functional form) of the interaction potential.
        """,
    )

    parameters = Quantity(
        type=dict[str, np.float64], 
        shape=[],
        description="""
        Dictionary of label and parameters of the interaction potential.
        """,
    )

class BondPotential(Potential):
    """
    Section containing information about bond potentials.

    Suggested types are: harmonic, cubic, Morse, fene, tabulated
    """

    equilibrium_value = Quantity(
        type=np.float64,
        unit='m',
        shape=[],
        description="""
        Specifies the equilibrium bond distance.
        """,
    )

    force_constant = Quantity(
        type=np.float64,
        shape=[],
        unit='J / m **2',
        description="""
        Specifies the force constant of the bond potential.
        """,
    )

```

where I define some default structure for parameters, but then define specific, named parameters for particular types of potentials. But in this case `parameters` becomes obsolete, so now I don't think it really makes sense to define it at this abstract level (or even at the level of `Potential()`). If you do define it at this abstract level I think you need a structure with labels/descriptors and values, right?

JFRudzinski (approved):
structure makes sense to me, fits perfectly with what I started creating for FFs before :+1: 

Just the one small question about the definition of parameters in general

src/nomad_simulations/schema_packages/model_method.py
JosePizarro3: I think indeed the parameters could be first defined in the specific classes that inherit from `BaseModelMethod`, and then we can check later on if there are commonalities or a way of standardizing these into the abstract class.

So in your example, you just need to define `BondPotential(BaseModelMethod)` (and the parameters as a `type=np.float64`, I guess and if you want, or only the `equilibrium_value` and `force_constant`) and `ForceField(ModelMethod)`.

Does this sound good?

JFRudzinski: yeah, sounds good.

I will actually just delete parameters inside of Potential() and then only have specific named parameters defined in the individual specific classes

PR #102, opened 2024-07-18
101 complete basis set migration

ndaelman-hu:
Migrate the basis set description to `numerical_settings.py`.

The basis set as whole is contained in `BasisSetContainer`, at the same level as other settings. Different electronic structures, e.g. wavefunction, density, Hamiltonian matrices, each have their own `BasisSetContainer`.

Different regions (distinguished by space, energy, or Hamiltonian) has its own set of parameters, described in `BasisSet`.
Basis sets localized around an atomic nucleus references a species (i.e. `AtomsState`) are classified as `LocalizedBasisSet(BasisSet)`.


ndaelman-hu:
@JosePizarro3 I think a first review is in place. I'll see whether there's anything missing, then moving on to testing.
Tbh, I think there will be little to test here, but feel free to make suggestions.

ndaelman-hu:
Might modify `BasisSetContainer` and `BasisSet` to `FullBasisSet` and `BasisSetPortion`, respectively.

JosePizarro3:
> Might modify `BasisSetContainer` and `BasisSet` to `FullBasisSet` and `BasisSetPortion`, respectively.

You can also call them `BasisSet` and `BasisSetContribution`. And we can standardize this way of calling contributions to a total section.

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -56,13 +61,29 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
 
 class Mesh(ArchiveSection):
     """
-    A base section used to specify the settings of a sampling mesh. It supports uniformly-spaced
-    meshes and symmetry-reduced representations.
+    A base section used to specify the settings of a sampling mesh.
+    It supports uniformly-spaced meshes and symmetry-reduced representations.
     """
 
+    type = Quantity(
```
JosePizarro3: For the `GeometricalSpace` in model_system.py, I was calling this `coordinates_system`. Also, check the MEnum there please :)

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -85,7 +106,7 @@ class Mesh(ArchiveSection):
         | `'Monkhorst-Pack'` | Regular mesh with an offset of half the reciprocal lattice vector. |
         | `'Gamma-offcenter'` | Regular mesh with an offset that is neither `'Gamma-centered'`, nor `'Monkhorst-Pack'`. |
         """,
-    )
+    )  # ! This should go under KMesh
```
JosePizarro3: Yeah, please move it.

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -115,20 +136,20 @@ class Mesh(ArchiveSection):
     )
 
     dimensionality = Quantity(
-        type=np.int32,
+        type=MEnum(1, 2, 3),
```
JosePizarro3: Does this thing work?

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -890,15 +911,295 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
         super().normalize(archive, logger)
 
 
-class BasisSet(NumericalSettings):
-    """"""
-
-    # TODO work on this base section (@ndaelman-hu)
+class BasisSet(ArchiveSection):
+    """A type section denoting a basis set component of a simulation.
```
JosePizarro3: """
A base section used as an abstract class for all the different types of basis set components.
""""

You can also add a couple of examples, but I think is enough to say that it is an abstract class.

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -890,15 +911,295 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
         super().normalize(archive, logger)
 
 
-class BasisSet(NumericalSettings):
-    """"""
-
-    # TODO work on this base section (@ndaelman-hu)
+class BasisSet(ArchiveSection):
```
JosePizarro3: Another alternative name would be: `BasisSetComponent`

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -890,15 +911,295 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+
+    smearing_method = Quantity(
+        type=MEnum('Fermi-Dirac', 'Gaussian', 'Methfessel-Paxton'),
+        description="""
+        Method used to smear the occupation numbers in the plane-wave basis set.
+        """,
+    )
+
+    def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
+        super().normalize(archive, logger)
```
JosePizarro3: Just a small note here: watch out for super normalizations when doing multiple inheritance. The logic might get very complex (although in this case it is trivial, just a note for the future).

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -890,15 +911,295 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
 
     def __init__(self, m_def: 'Section' = None, m_context: 'Context' = None, **kwargs):
         super().__init__(m_def, m_context, **kwargs)
         # Set the name of the section
         self.name = self.m_def.name
 
+
+class PlaneWaveBasisSet(BasisSet, Mesh):
+    """
+    Defines a plane-wave basis set over a singular mesh.
```
JosePizarro3: Ok, I have some suggestions for the descriptions of this and the other different types: can you add a couple of phrases describing what each basis is, and one ref if possible? Nothing too complex, just something that connects everything nicely (you are welcome to add some formulation, like the actual basis set $\mathcal{B}=\{G_{i}\}$ with i=1,2...N or something like that)

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -890,15 +911,295 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+
+    cutoff_energy = Quantity(
+        type=np.float64,
+        unit='joule',
+        description="""
+        Cutoff energy for the plane-wave basis set. The simulation uses plane waves with energies below this cutoff.
+        """,
+    )
+
+    smearing_method = Quantity(
```
JosePizarro3: Is this not something typical from the `NumericalSettings` and not from `BasisSet`? Maybe we should have a `Smearing(NumericalSettings)` class?

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -890,15 +911,295 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        super().normalize(archive, logger)
+
+
+class LocalizedBasisSet(BasisSet):
+    """
+    Section that connects a basis set to a localized unit, often an atom.
+    """
+
+    species_definition = SubSection(sub_section=AtomsState.m_def)
+    # ? @JosePizarro3: are we ensure that the atoms_state is always set? can you point me to any normalizer?
```
JosePizarro3: Well, you can use the `normalize` function to check this.

In any case, `species_definition` should be a ref to `ModelSystme.cell.atoms_state`, and you can use `AtomsState` to be populated in case the basis is localized as in this case. If using references:
```python
    species_definition = Quantity(type=AtomsState, shape=['*'], description="""Reference to the section `AtomsState` containing the information of the states of the atoms used for the localized basis set.""")
```

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -890,15 +911,295 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+
+    # TODO: add atom index-based instantiator for species if not present
+
+
+class AtomCenteredFunction(ArchiveSection):
```
JosePizarro3: I guess this is a place holder for later implementing something else? If so, would you mind adding a TODO?

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -890,15 +911,295 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    """
+
+    functional_composition = SubSection(
+        sub_section=AtomCenteredFunction.m_def, repeats=True
+    )
+
+    def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
+        super().normalize(archive, logger)
+        # self.name = self.m_def.name
+        # TODO: set name based on basis functions
```
JosePizarro3: I guess you can check some standard namings. I am not sure what you wrote on "? use naming BSE", but are not the names in FHIaims precisely ids for these basis sets?

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -890,15 +911,295 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        # ? use naming BSE
+
+
+class APWWavefunction(ArchiveSection, ABC):
```
JosePizarro3: You don't need double inheritance to `ABC` here, nor to define an `@abstractmethod:normalize()`. Just inherit from `ArchiveSection` and define the normalize() and super().normalize().

Also, you could specify that this is just an abstract class of `APWOrbital` and `APWLocalOrbital` used to be able to add them as sub-sections of `APWLChannel` depending on the problem at hand.

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -890,15 +911,295 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        super().normalize(archive, logger)
+        if np.any(np.isneginf(self.differential_orders)):
+            logger.error('`APWLOrbital.differential_order` must be non-negative.')
+        if np.any(np.isneginf(self.boundary_orders)):
+            logger.error('`APWLOrbital.boundary_orders` must be non-negative.')
+
+
+class APWLChannel(ArchiveSection):
+    """
+    Channel of the angular momentum ($l$),
```
JosePizarro3: This description is a bit criptic.

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -890,15 +911,295 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        # self.name = self.m_def.name
+        self.n_wavefunctions = len(self.orbitals) * (2 * self.name + 1)
+
+
+class MuffinTinRegion(LocalizedBasisSet, Mesh):
+    """
+    Muffin-tin region around atoms, containing the augmented part of the APW basis set.
+    """
+
+    # there are 2 main ways of structuring the APW basis set
```
JosePizarro3: I actually think this should go in the description

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -890,15 +911,295 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+
+    l_channels = SubSection(sub_section=APWLChannel.m_def, repeats=True)
+
+    def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
+        super().normalize(archive, logger)
+        self.type = 'Spherical'
+
+
+class BasisSetContainer(NumericalSettings):
+    """Section providing an overview of the full basis set.
```
JosePizarro3: """
A base section used to define the basis set used during the diagonalization of the Hamiltonian defined as `ModelMethod`. This section may contain multiple basis set specifications under the `basis_sets` (rename here to `basis_set_components` or something similar), each with their own parameters.
"""

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -890,15 +911,295 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    The basis set may contain multiple basis sets specifications `BasisSet`,
+    each with their own parameters."""
+
+    native_tier = Quantity(
+        type=str,
+        shape=[],
+        description="""
+        Code-specific tag indicating the precision used
```
JosePizarro3: Maybe specify a couple of examples?

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -890,15 +911,295 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        for the basis set and meshes of numerical routines.
+        """,
+    )
+
+    electronic_structure = Quantity(
```
JosePizarro3: You can delete this. The `NumericalSettings` is a sub-section of the `ModelMethod`, so it will have this information above anyways.

JosePizarro3 (commented):
Minor restructuring, and I left some questions opened.

More importantly, I saw some normalization in the method normalizer (nomad/normalizing/method.py) in which I think you are resolving the type. Is this under control? I think we need to have under control these normalizations that are generating extra information, so my rule would be:
- If the MethodNormalizer generates a new identifier or new information, you could move it here.
- If it is just copying stuff into results, you can forget about it (but probably have it as a note to make sure we don't forget things).

ndaelman-hu:
> More importantly, I saw some normalization in the method normalizer (nomad/normalizing/method.py) in which I think you are resolving the type. Is this under control? I think we need to have under control these normalizations that are generating extra information, so my rule would be:

>   - If the MethodNormalizer generates a new identifier or new information, you could move it here.
>   - If it is just copying stuff into results, you can forget about it (but probably have it as a note to make sure we don't forget things).

Thx for pointing this out. There is a small piece of code that assigns the overall basis set type for APW. This should be migrated over.
There is also some legacy normalization at the bottom of `normalizing/method.py`, but that's irrelevant to our purposes.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: I end up using only 1 of them (don't want parallel schemas). I can add that to the description.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: Yeah, I can definitely add it to the description. Unless there are some technical issues, I prefer keeping `ABC` to ensure that the abstract nature is really enforced.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: BSE (Basis Set Exchange) is an online DB for Gaussian and Slater basis sets.
FHI-aims uses NAO (numeric atomic orbitals) which have a different description.
Both are centered around atom nuclei, though.

src/nomad_simulations/schema_packages/numerical_settings.py
JosePizarro3: Please, delete ABC. We don't really need it.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: I added some now. I used the common terminology, but could also refer to the section names directly.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: Sure, this one also crossed my mind. The issue is that in common language, the context defines to which basis set is referred to.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: If both classes have `normalize`, it should call the first one, right? There is a way of stating that it should choose the 2nd one: `super(Mesh, self).normalize()`.

I'll see to be more explicit here.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: Applied. Tbh, at first I was somewhat hesitant to use `AtomsState`. It contains all the relevant information, but I'm unsure in how far it's supposed to represent an actual "physical" system. Since Wannier uses it too, I guess other basis sets should be fine.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: > BSE (Basis Set Exchange) is an online DB for Gaussian and Slater basis sets.

Note that I did not yet implement this part, am just considering whether it's worth it. I would keep it for a separate PR.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: Done. I left out the part on `APWOrbital` and `APWLocalOrbital`, though: parent classes should not introduce their children. I do add a throwback in the descriptions of the child sections.

Note that explicit mentions in the descriptions are hard to maintain, as they don't get type-checked.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: Added with only slight modifications

ndaelman-hu: Btw, I would reserve renaming `BasisSetContainer` and `BasisSet` to `FullBasisSet`/`BasisSet` and `BasisSetComponent` until we migrate the parsers.
Else the naming may become confusing.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: I'm not finding the section nor quantity... I checked, and I did branch of the most recent version of `develop`

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: I think you're interpreting this quantity differently than what I'm suggesting:
basis sets may vary for the wavefunction, electron density, and other subroutines that involve specific integrals. This just helps distinguishing them.

This used to be a string, but should really become a section by its own right. You mentioned in the issue that no such quantity exists yet. So, I'll keep this to a TODO for now.

EDIT: 
> other subroutines that involve specific integrals

I guess the Hamiltonian contributions exist now, but here we're referring to the vectors that go into the integral, not the kernel.

src/nomad_simulations/schema_packages/numerical_settings.py
JosePizarro3: https://github.com/nomad-coe/nomad-simulations/blob/develop/src/nomad_simulations/schema_packages/model_system.py#L153

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: They're very code-dependent, which I would reserve for the parser plugins.
I added this to the description with a generic example.

src/nomad_simulations/schema_packages/numerical_settings.py
JosePizarro3: Not sure if I understand. Can you give me some examples of this string?

src/nomad_simulations/schema_packages/numerical_settings.py
JosePizarro3: The thing is more involved: https://www.geeksforgeeks.org/method-resolution-order-in-python-inheritance/

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: Okay, I misread the filename. Thx for the link.
I'm still unsure of where to specify the coordinates themselves, though:

`Mesh` refers to the `points` which are dependent on the coordinate system. `GeometricalSpace` only uses the coordinate system in the normalization.
The current implementation always assumes `x, y, z` coordinates: it does not automatically convert to `r, theta, phi` when changing to `spherical` for example...

We can likely retain the original distinction/semantics, but I suggest _dependency injection_ for any geometric functionality, meaning that we pass a `CoordinateSystem` class implementing a set of routines to the section initializer. Further normalization is then handled by the section calling those routines, as usual.
Meanwhile, we use _polymorphism_ to specialize `CoordinateSystem` into its various kinds. (Note that Cartesian should be clarified where it assumes 90 angles + we're missing ellipsoidal.)

This is going quite outside this PR. For now, I'll leave `Mesh` as was.
I can open a new issue, assign myself, and have you review it. This would be for after the holidays, though. Just give it a thought.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: As mentioned below, I'm very much for the name change, but it's better to perform it when migrating the parsers. This will prevent any compatibility issues.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: Regarding the FHI-aims basis set, these indeed have names, but they are code-specific. I stored them under `native_tier`, as they also fully determine the precision. Note that a user can always customize the basis set in this code.

The names in the BSE, e.g. STO-3G or 6-31++G, are a community standard and employed in all molecular codes.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: Indeed, similar to NAO and LAPW, these atomic orbitals are vectors of fundamental functions (typically Gaussians).
While there are lot combinations that have become standard, one can always construct them freely. This is where the BSE might come in handy.

This part never got implemented in the old schema. It should be after this migration.

src/nomad_simulations/schema_packages/numerical_settings.py
JosePizarro3: I think you are over-thinking this. When you are doing parsers, we will know that it is the new way. Otherwise, and immediately when parsing, one would realize that the naming of the legacy style is wrong.

Simply change it here and that's it. Forget about the old schema. Let it burn.

984b7148-792c-4132-9ed6-3c6728084e98_942x646

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: Those implemented in the parsers are `wavefunction` or `electron_density`. They may also cover [_augmentation charges_](https://www.vasp.at/wiki/index.php/ADDGRID), [exact exchange](https://www.vasp.at/wiki/index.php/PRECFOCK) integrals, etc.
Essentially, a code can freely generate grids and basis sets for any specific case, e.g. https://wiki.fysik.dtu.dk/gpaw/documentation/core.html#uniform-grids

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: Lol

I'm really not trying to preserve the old schema conventions. It's just that if I shift the name `BasisSet` around, it'll get confusing for when I hook up the parsers, e.g. this was the `basis_set`, but now it's a component that goes into the `basis_set` that used to be a container. Have mercy on my poor dev soul

src/nomad_simulations/schema_packages/numerical_settings.py
JosePizarro3: Ok, I might understand that you wanna differentiate something like `scf_observable` or something like that for WFs or densities, but augmentation charges could go into its own quantity, and the same for how to treat the Hartree part, no?

My main problem is on the name of the quantity, "electronic_structure"

src/nomad_simulations/schema_packages/numerical_settings.py
JosePizarro3: This is true if you take months to implement the new parsers, but you will see it is not the case.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: I've already commented it out (will push soon), so we can return to it later on.

I'm fine with a different name, just not sure what to call it. The main examples are representations of the electronic structure, but the more exotic cases are indeed vaguer.

The reason I try to place them all here, is that they can be represented by the same (basis set) approach.
In the VASP cases that I linked, they are just grids. The HF part (_overlap densities_ actually) is still handled as plane waves, just with a coarser spacing. It even has it own tier system! Some codes do something similar for the XC density too. Augmentation charges are a necessary addition to the basis set when dealing with PAW pseudopotentials.

Don't get me wrong, though, each addition of a representation should be considered by itself.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: ...
I'll be out on holidays for the next 2 weeks, so there's that.
After, I'd like to start migrating the parsers to the new entry-points system, with each their own repo. Migrating means, retaining the old parser, while also have the setup ready to add the new one. You off-handly suggested this last week, and I agree that it is best to decouple the parsers before deprecating the old ones.

Exciting is definitely on the top of the list, but I'm less sure about Abinit or Wien2k. Regardless, I'd handle the basis set migration for all of them at once. I'm not sure about the timeframe of when we'll get around to it, but we can bump up priority there.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: I extended the description for plane-waves and the LAPW consituents (though their types already contained some formulas). I'll hold off on `AtomCentered` for now, which is kind of a stub anyhow.

src/nomad_simulations/schema_packages/numerical_settings.py
JosePizarro3: Please, just change it to the new naming. You are not the only one transfering parsers, and better to have the schema naming under control so the definitions don't changr in 1 or 2 months.

Furthermore, this plan you have in mind I think is not entirely correct. The idea would be to, first, take over one parser and transfer to the new schema. We do not need to put parsers in the new entry points mechanism with the old schema.

But you go on holidays now. Did you change everything you wanted? I can takr over and merge next week if you can't.

ndaelman-hu: I finished the majority, just some normalization and testing. Unless anyone is touching `numerical_settings.py`, just leave it and I will finish it once I'm back. You'll be back the week after, so then you review for the last time and we merge.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: No, `MEnum` only supports `str`. I traced the code back to `Enum.__init__()` in `data_type.py`.
I could extend it to include other types, but it looks like a rabbit hole. Probably not a good time investment.

JosePizarro3:
hey, just back from vacation

@ndaelman-hu how is this going? Is it ready?

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+        description="""
+        Reference to the section `AtomsState` specifying the localization of the basis set.
+        """,
+        a_eln=ELNAnnotation(components='ReferenceEditQuantity'),
+    )
+
+    # TODO: add atom index-based instantiator for species if not present
+
+    hamiltonian_scope = Quantity(
+        type=EnergyContribution,
```
JosePizarro3: This should point to `BaseModelMethod`, and the info of the Hamiltonian contribution should appear under `model_method[...].contributions[i]`. `EnergyContribution` can use a `model_method_ref` to point to that as well.

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+
+from nomad_simulations.schema_packages.atoms_state import AtomsState
+from nomad_simulations.schema_packages.numerical_settings import (
+    Mesh,
+    NumericalSettings,
+)
+from nomad_simulations.schema_packages.properties.energies import EnergyContribution
+
+
+def check_normalized(func: Callable):
```
JosePizarro3: Maybe you can move this to `general.py` so we can all use it.

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+    cutoff_energy = Quantity(
+        type=np.float64,
+        unit='joule',
+        description="""
+        Cutoff energy for the plane-wave basis set.
+        The simulation uses plane waves with energies below this cutoff.
+        """,
+    )
+
+    @property  # ? keep, or convert to `Quantity`?
```
JosePizarro3: You can define the quantity. The radius cutoff is interesting as well, and even if the info is repeated, I argue that this won´t clutter the archive.

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+        Compute the cutoff radius for the plane-wave basis set, expressed in reciprocal coordinates.
+        """
+        if self.cutoff_energy is None:
+            return None
+        m_e = const.m_e * ureg(const.unit('electron mass'))
+        h = const.h * ureg(const.unit('Planck constant'))
+        return np.sqrt(2 * m_e * self.cutoff_energy) / h
+
+    def normalize(self, archive: EntryArchive, logger: BoundLogger) -> None:
+        super(BasisSet, self).normalize(archive, logger)
```
JosePizarro3: I am not 100% that the MRO is as you expect here. This is a bit of a pain logic, but can you double check and confirm to me? Is testing going through as expected?

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+        This cutoff has no units, referring to the product of the smallest muffin-tin radius
+        and the length of the cutoff reciprocal vector ($r_{MT} * |K_{cut}|$).
+        """,
+    )
+
+    def set_cutoff_fractional(
+        self, mt_r_min: pint.Quantity, logger: BoundLogger
+    ) -> None:
+        """
+        Compute the fractional cutoff parameter for the interstitial plane waves in the LAPW family.
```
JosePizarro3, line 156: Please, add Args definitions and what this are, e.g., what is `mt_r_min`.

Also, why not being more verbose?

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+        of the Hamiltonian term to which the basis set applies.
+        """,
+        a_eln=ELNAnnotation(components='ReferenceEditQuantity'),
+    )
+
+    # ? band_scope or orbital_scope: valence vs core
+
+    def normalize(self, archive, logger):
+        super().normalize(archive, logger)
+        self.name = self.m_def.name
```
JosePizarro3, line 79: Is this assigning `BasisSet` to all the other basis set classes? I think you should write the `normalize` function on each of the basis set so it assigns properly `'PlaneWaveBasisSet'`, etc.

src/nomad_simulations/schema_packages/basis_set.py
JosePizarro3: You can then define a function and a call in the `normalize` to extract this.

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+    def wrapper(self, archive: EntryArchive, logger: BoundLogger) -> None:
+        if self._is_normalized:
+            return None
+        func(self, archive, logger)
+        self._is_normalized = True
+
+    return wrapper
+
+
+class BasisSet(NumericalSettings):
```
JosePizarro3: Would you mind adding a TODO here:
```python
# TODO check implementation of `BasisSet` for Wannier and Slater-Koster orbitals
```

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+
+    functional_composition = SubSection(
+        sub_section=AtomCenteredFunction.m_def, repeats=True
+    )  # TODO change name
+
+    def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
+        super().normalize(archive, logger)
+        # self.name = self.m_def.name
+        # TODO: set name based on basis functions
+        # ? use naming BSE
```
JosePizarro3: It might conflict with elasticsearch and BetheSalpeterEquation, so we have to carefully check this

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+        shape=['n_terms'],
+        description="""
+        Derivative order of the radial wavefunction term.
+        """,
+    )  # TODO: add check non-negative # ? to remove
+
+    def _get_open_quantities(self) -> set[str]:
+        """Extract the open quantities of the `APWBaseOrbital`."""
+        return {
+            k for k, v in self.m_def.all_quantities.items() if self.m_get(v) is not None
```
JosePizarro3, line 265: you cannot do `getattr(self, k)` directly and forget about the values?

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+                )
+            self.n_terms = None
+
+        # enforce differential order constraints
+        if np.any(np.isneginf(self.differential_order)):
+            logger.error(
+                '`APWBaseOrbital.differential_order` must be completely non-negative.'
+            )
+
+    def __init__(self, *args, **kwargs):
```
JosePizarro3, line 334: just a picky comment: would you mind moving this above other functions?

JosePizarro3: now that I think about it, and seeing your decorator, do you need this?

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+
+    def normalize(self, archive: EntryArchive, logger: BoundLogger) -> None:
+        super().normalize(archive, logger)
+
+        # enforce quantity length (will be used for type assignment)
+        new_n_terms = self.get_n_terms()
+        if self.n_terms is None:
+            self.n_terms = new_n_terms
+        elif self.n_terms != new_n_terms:
+            if logger is not None:
```
JosePizarro3: please, take out this condition, logger should be set anyways

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+                logger.error(
+                    f'Inconsistent lengths of `APWBaseOrbital` quantities: {self.m_def.quantities}. Setting back to `None`.'
+                )
+            self.n_terms = None
+
+        # enforce differential order constraints
+        if np.any(np.isneginf(self.differential_order)):
```
JosePizarro3: `if self.differential_order is not None and...`

JosePizarro3: or something more careful regarding the length of the array

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+    """
+    Implementation of `APWWavefunction` capturing the foundational (S)(L)APW basis sets, all of the form $\sum_{lm} \left[ \sum_o c_{lmo} \frac{\partial}{\partial r}u_l(r, \epsilon_l) \right] Y_lm$.
+    The energy parameter $\epsilon_l$ is always considered fixed during diagonalization, opposed to the original APW formulation.
+    This representation then has to match the plane-wave $k_n$ points within the muffin-tin sphere.
+
+    * D. J. Singh and L. Nordström, \"INTRODUCTION TO THE LAPW METHOD,\" in Planewaves, pseudopotentials, and the LAPW method, 2nd ed. New York, NY: Springer, 2006, pp. 43-52.
+    """
+
+    type = Quantity(
+        type=MEnum('apw', 'lapw', 'slapw'),  # ? where to put 'spherical_dirac'
```
JosePizarro3: well, it is given as an option in the description. No idea what this is, so you can delete from the description or add it in the MEnum

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+        # assign a APW orbital type
+        # this snippet works of the previous normalization
+        new_type = self.n_terms_to_type(self.n_terms)
+        if self.type is None:
+            self.type = new_type
+        elif self.type != new_type:
+            logger.error(
+                f'Inconsistent `APWOrbital` type: {self.type}. Setting back to `None`.'
+            )
+            self.type = None
```
JosePizarro3, line 395: so I am not sure deleting info from the schema is a good idea

I think a logger.error is more than enough to understand this situations, so please, can you delete this assignment and the one in `APWBaseOrbital`?

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+            return 'lo'
+        elif sorted(bo_terms) == [0, 0, 1]:  # ! double-check
+            return 'LO'
+        else:
+            return 'custom'
+
+    @check_normalized
+    def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
+        super().normalize(archive, logger)
+        if np.any(np.isneginf(self.boundary_order)):
```
JosePizarro3: `if self.boundary_order is not None and len(self.boundary_order) != 0 and...`

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+
+    orbitals = SubSection(sub_section=APWBaseOrbital.m_def, repeats=True)
+
+    def _determine_apw(self) -> dict[str, int]:
+        """
+        Produce a count of the APW components in the l-channel.
+        Invokes `normalize` on `orbitals`.
+        """
+        for orb in self.orbitals:
+            orb.normalize(None, None)
```
JosePizarro3: please, pass the arguments

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+    n_orbitals = Quantity(
+        type=np.int32,
+        description="""
+        Number of wavefunctions in the l-channel, i.e. $(2l + 1) n_orbitals$.
+        """,
+    )
+
+    orbitals = SubSection(sub_section=APWBaseOrbital.m_def, repeats=True)
+
+    def _determine_apw(self) -> dict[str, int]:
```
JosePizarro3, line 448: I am not sure why this is private

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+        for orb in self.orbitals:
+            if isinstance(orb, APWOrbital) and orb.type.lower() in type_count.keys():
+                type_count[orb.type] += 1
+            elif isinstance(orb, APWLocalOrbital):
+                type_count['lo'] += 1
+            else:
+                type_count['other'] += 1  # other de facto operates as a catch-all
+        return type_count
+
+    def __init__(self, *args, **kwargs):
```
JosePizarro3, line 469: 2 init functions

And check my comment above

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+        super().__init__(*args, **kwargs)
+        self._is_normalized = False
+
+    @check_normalized
+    def normalize(self, archive: EntryArchive, logger: BoundLogger) -> None:
+        super(BasisSet).normalize(archive, logger)
```
JosePizarro3: again, not sure whether this is producing the desired result

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+        type=np.int32,
+        description="""
+        Maximum angular momentum quantum number that is sampled.
+        Starts at 0.
+        """,
+    )
+
+    l_channels = SubSection(sub_section=APWLChannel.m_def, repeats=True)
+
+    def _determine_apw(self) -> dict[str, int]:
```
JosePizarro3, line 508: again, not sure why this is private

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+            for channel in self.l_channels[1:]:
+                type_count.update(channel._determine_apw())
+        return type_count
+
+    @check_normalized
+    def normalize(self, archive, logger):
+        super().normalize(archive, logger)
+
+
+class BasisSetContainer(NumericalSettings):
```
JosePizarro3, line 534: I am confused: `BasisSet` is also a `NumericalSettings`. So this might get confusing for people trying to understand what info to store under `model_method.numerical_settings`. 

Before, we decided that only the container inherits from it so that people can append it to the `ModelMethod` archive. What has changed?

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+    native_tier = Quantity(
+        type=str,  # to be overwritten by a parser `MEnum`
+        description="""
+        Code-specific tag indicating the overall precision based on the basis set parameters.
+        The naming conventions of the same code are used. See the parser implementation for the possible values.
+        The number of tiers varies, but a typical example would be `low`, `medium`, `high`.
+        """,
+    )
+
+    # TODO: add reference to `electronic_structure`,
```
JosePizarro3, line 551: Delete this block comment

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+
+class BasisSetContainer(NumericalSettings):
+    """
+    A section defining the full basis set used for representing the electronic structure
+    during the diagonalization of a Hamiltonian (component), as defined in `ModelMethod`.
+    This section may contain multiple basis set specifications under the basis_set_components,
+    each with their own parameters.
+    """
+
+    native_tier = Quantity(
```
JosePizarro3, line 542: why not `code_specific_precision`?

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+        """,
+    )
+
+    # TODO: add reference to `electronic_structure`,
+    # specifying to which electronic structure representation the basis set is applied
+    # e.g. wavefunction, density, grids for subroutines, etc.
+
+    basis_set_components = SubSection(sub_section=BasisSet.m_def, repeats=True)
+
+    def _determine_apw(self) -> Optional[str]:
```
JosePizarro3, line 557: why private?

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+    def normalize(self, archive: EntryArchive, logger: BoundLogger) -> None:
+        super().normalize(archive, logger)
+        pws = [
+            comp
+            for comp in self.basis_set_components
+            if isinstance(comp, PlaneWaveBasisSet)
+        ]
+        if len(pws) > 1:
+            logger.warning('Multiple plane-wave basis sets found were found.')
+        self.name = self._determine_apw(logger)
```
JosePizarro3: this function does not accept a `logger`

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+        mt_min = None
+        for comp in self.basis_set_components:
+            if isinstance(comp, MuffinTinRegion):
+                if mt_min is None or comp.radius < mt_min.radius:
+                    mt_min = comp
+        return mt_min
+
+    def normalize(self, archive: EntryArchive, logger: BoundLogger) -> None:
+        super().normalize(archive, logger)
+        pws = [
```
JosePizarro3: pws, comp... you can use the longer name here

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -0,0 +1,642 @@
...
+        if len(pws) > 1:
+            logger.warning('Multiple plane-wave basis sets found were found.')
+        self.name = self._determine_apw(logger)
+        try:
+            pws[0].set_cutoff_fractional(self._smallest_mt(), logger)
+        except (IndexError, AttributeError):
+            logger.error('Expected a `APWPlaneWaveBasisSet` instance, but found none.')
+
+
+def generate_apw(
```
JosePizarro3, line 621: What is this for? I only see it in the tests being used.

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -838,7 +838,7 @@ class ExcitedStateMethodology(ModelMethodElectronic):
     A base section used to define the parameters typical of excited-state calculations. "ExcitedStateMethodology"
     mainly refers to methodologies which consider many-body effects as a perturbation of the original
     DFT Hamiltonian. These are: GW, TDDFT, BSE.
-    """
+    """  # Note: we don't really talk about Hamiltonians in DFT: their physics is accommodated in the functional itself
```
JosePizarro3, line 841: what?

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -151,12 +151,20 @@ class GeometricSpace(Entity):
...
+        space. Defaults to 'cartesian'.
+
+        | name       | description | dimensionalities | coordinates |
+        |------------|-------------|------------------|-------------|
+        | cartesian  | coordinate system with fixed angles between the axes (not necessarily 90°) | 1, 2, 3 | x, y, z |
+        | cylindrical| cylindrical symmetry | 3 | r, theta, z |
+        | spherical  | spherical symmetry | 3 | r, theta, phi |
+        | ellipsoidal| spherically elongated system | 3 | r, theta, phi |
+        | polar      | spherical symmetry | 2 | r, theta |
+        """,  # ? could this not be extended to the k-space
```
JosePizarro3, line 167: yes, that's why this is an abstract class, but in k-space we are anyways talking about cartesian space

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -54,15 +54,29 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        """,
+    )
+
+
 class Mesh(ArchiveSection):
     """
-    A base section used to specify the settings of a sampling mesh. It supports uniformly-spaced
-    meshes and symmetry-reduced representations.
+    A base section used to specify the settings of a sampling mesh.
+    It supports uniformly-spaced meshes and symmetry-reduced representations.
```
JosePizarro3, line 74: why moved down?

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -309,14 +310,27 @@ class KMesh(Mesh):
     """
 
     label = Quantity(
-        type=MEnum('k-mesh', 'q-mesh'),
+        type=MEnum('k-mesh', 'q-mesh'),  # ? add g-mesh for LAPW
```
JosePizarro3: go ahead, this is a simple identifier and g-mesh is a good idea

tests/conftest.py
```diff
@@ -402,3 +402,129 @@ def k_line_path() -> KLinePathSettings:
 @pytest.fixture(scope='session')
 def k_space_simulation() -> Simulation:
     return generate_k_space_simulation()
+
+
+refs_apw = [
```
JosePizarro3, line 408: ouch 😅 

now I understand why it took you so long to prepare this testing

JosePizarro3 (approved):
Good. Only a small clarification for `BasisSet` and `BasisSetContainer` both inheriting from `NumericalSettings`.

After clarifying, you can merge at will.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: It should only be called by the normalizer. Testing may call it standalone.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: This is something that will be for another PR, given that we don't have any place yet for the actual eletronic object.
It looks like with the C4 task (@EBB2675), this may be incoporated. It is even possible that the referencing direction is the inverse of how I stipulate it here.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: As a name? I'm fine changing it, but this has been the term used in all the communication material.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: Done.

src/nomad_simulations/schema_packages/basis_set.py
JosePizarro3: What is an "electronic object"?

ndaelman-hu:
> Good. Only a small clarification for `BasisSet` and `BasisSetContainer` both inheriting from `NumericalSettings`.
> 
> After clarifying, you can merge at will.

You're right, I fixed it now.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: done

ndaelman-hu: Note, though, that this now introduces the `EntryArchive` and `BoundLogger` dependencies in `general.py`, where before they were only called by `mypy`. We should keep an eye out for potential circular imports.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: done

src/nomad_simulations/schema_packages/basis_set.py
JosePizarro3: For typing, you can just put them between `'EntryArchive'` and `'BoundLogger'` so we avoid that.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: > ? use naming BSE

This refers to the formatting in the Basis Set Exchange, not changing the name to BSE. I've modified the comment for clarity.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: Honestly, idk what the proper NOMAD section interface is: it changes all the time...
Your suggestion is appropriate for vanilla classes, yes.

src/nomad_simulations/schema_packages/model_method.py
ndaelman-hu: I'd change the phrasing from "DFT Hamiltonian" to "density functional", since the former is not really used.

Overall, I associate "Hamiltonian" more so with operators acting upon vectors / quantum states.
DFT, meanwhile, rewrites the Hamiltonian to be a functional of another function, typically the external potential $v\left(\vec{r}\right)$. To rephrase it in terms of the electronic density function $n\left(\vec{r}\right)$, it uses a Legendre transform. Therefore, I'd consider it a different mathematical object.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: It's the relativistic correction to the basis set. I'd revisit this some other time.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: > so I am not sure deleting info from the schema is a good idea

Yeah, this is a general policy choice that we ought to discuss in Area C, heck even with the other Areas.

Here's my reasoning for setting it back: the logs are not hard-tied to the `archive.json`, i.e. we cannot guarantee that a user will see them together. So a user may naively assume the data to be correct, while a check will reveal an inconsistency. I'd rather have them skip this entry than consume incorrect data.

This is also related to the storage of the original data:
- for parsers, the user can always go to the original upload. Inconsistencies are more so a sign of bad parser logic.
- for ELNs, the user inputs the data, so overwriting it there may lead to data loss and is not advisable.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: Because it should be called by `normalize`. The LAPW community uses a variety of abbreviations to refer to the basis set as a whole (i.e. our `BasisSetContainer`). `_determine_apw` just provides this name.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: Yes, I spotted this and removed it.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: You're right, this goes outside the interface specification. Done

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: This was a mistake. I had to restructure a couple of times. Has been corrected.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: ... I think the definition made it quite clear in this case, but alright.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: Might also be used by the parsers: while APW is highly nested, the input typically contains only some general parameters with minor tweaks. This bridges both styles of writing.

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: I need a k-vector denoting the radius of the sphere.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: The description seems a bit too specific for linear response, I'd see to extend that.

src/nomad_simulations/schema_packages/model_method.py
JosePizarro3: > Therefore, I'd consider it a different mathematical object.

It is not.

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: I'm not following

JosePizarro3: As EntryArchive and BoundLogger are just for typing, you can avoid importing them directly, and rather add at the beginning of the file:
```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from ... import EntryArchive, BoundLogger
```

And in the variables typing just put them between quoutes.

ndaelman-hu: > And in the variables typing just put them between quoutes.

Okay, this is an important detail. Else it'll raise a `NameSpaceError`.

JosePizarro3: https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

understanding why types and only run when type checking: https://vickiboykis.com/2023/12/11/why-if-type_checking/

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: Nope, `self` refers to the calling class.

src/nomad_simulations/schema_packages/basis_set.py
JosePizarro3: It is ok, if you don't want to change it is fine. When first seeing this quantity I had to read the description to read what was referring to, but if you want you can leave it as `native_tier`.

In the front-end there is a mix: `code_specific_tier` :)

ndaelman-hu:
@JosePizarro3 is the expected use of `AtomicCell` to also point to individual species / atoms?

src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu: > In the front-end there is a mix: code_specific_tier :)

Honestly, who decides in this stuff? No other area but C1 uses this, so why does this get changed without any heads-up?

JosePizarro3:
> @JosePizarro3 is the expected use of `AtomicCell` to also point to individual species / atoms?

No, for that you point to AtomsState

ndaelman-hu:
> > @JosePizarro3 is the expected use of `AtomicCell` to also point to individual species / atoms?
> 
> No, for that you point to AtomsState

Okay, that's what I thought. Thx for confirming.
This might raise issues, as @EBB2675 brought up some examples of atom groups that all share the same basis set description.
Anyhow, we can look into that extension during another PR.

JosePizarro3:
> > > @JosePizarro3 is the expected use of `AtomicCell` to also point to individual species / atoms?
> > 
> > 
> > No, for that you point to AtomsState
> 
> Okay, that's what I thought. Thx for confirming. This might raise issues, as @EBB2675 brought up some examples of atom groups that all share the same basis set description. Anyhow, we can look into that extension during another PR.

You can always reference a list of `AtomsState`, I think. ~~Never tried before tho.~~ Ok I am correcting myself immediately, as I added this for Green's functions, but I still need to test it in the w2dynamics parser.

PR #114, opened 2024-09-10
added previous ff schema from older version

### LINKED ISSUE #113: Add Force Field Schema
To replace #18

JFRudzinski:
- lays basic foundation for FF schema
- adds back ForceCalculations() as a numerical setting
- no effect to nomad-parser-plugins-simulation


JFRudzinski:
@Bernadette-Mohr You could already take a look at this. I have finished implementing bond potentials, and also added a more general test suite that may be useful to you in other contexts.

I will continue expanding to the other potential types, but let me know if you have any major feedback about the direction this is going.

JFRudzinski:
@JosePizarro3 @ndaelman-hu @EBB2675 @Bernadette-Mohr  FYI - I implemented a function here in `test_force_field.py` for comparing 2 nested dictionaries, that deals with some typical data types that we have in the archive. This way I just take a section that I populated in the archive and do `<section>.m_to_dict()>`, and then compare with the expected results that I input in a pytest fixture.

I'm sure it could be improved, and maybe you already have such tools in your testing, but perhaps we can collect these into one file in `tests/` so that devs can use them, and it encourages consistency in testing.

Also, I am definitely open to feedback concerning the structure of the tests.

ndaelman-hu:
> @JosePizarro3 @ndaelman-hu @EBB2675 @Bernadette-Mohr FYI - I implemented a function here in `test_force_field.py` for comparing 2 nested dictionaries, that deals with some typical data types that we have in the archive. This way I just take a section that I populated in the archive and do `<section>.m_to_dict()>`, and then compare with the expected results that I input in a pytest fixture.
> 
> I'm sure it could be improved, and maybe you already have such tools in your testing, but perhaps we can collect these into one file in `tests/` so that devs can use them, and it encourages consistency in testing.
> 
> Also, I am definitely open to feedback concerning the structure of the tests.

Regarding `assert_dict_equal`, perhaps it's easier if you list out the kind of comparison you want?
Python comes with a native implementation of `==` for most data structures, including `dict`:

```python
>> {'a': {'b': 2}, 'c': 3} == {'a': {'b': 2}, 'c': 3}
True
```

So, if all elements have to match in value (but not necessarily order), you can use the out-of-the-box definition.

If instead only a certain numerical threshold is needed, or only certain properties should be compared, I'd recursively apply a transformation / filtering to the `dict`s before comparison.
If you also want to perform such comparisons on the matching `Section`s too (e.g. as with `ModelSystem`), then I'd overwrite its `__eq__`.

P.s.: that isn't to say that your implementation is bad, just that it may already be covered. In Haskell, your approach would be called _defining the initial algebra to an F-algebra_. Given the nice maths of F-algebras, it's quite popular. For initial algebras constructed from algebraic data types, `Eq` can be derived algorithmically. While recursion is far less powerful in Python, it essentially uses the same trick here.

JFRudzinski:
@ndaelman-hu thanks for your feedback!

I did originally try to use the native python comparisons, the original function I implemented a while ago, but I believe I could not simply compare 2 dictionaries with arbitrarily nested structures and containing various and a priori unknown types...(I will have to take a look again)

Maybe I over complicated things now, but a couple examples that I needed to treat in a special way:
1. floats need to be compared with the `approx` function, and arrays of floats with `np.isclose()` 
2. np.inf is stored as a string in the archive -- I realize now that this is more of a problem with how I managed the data that I was comparing

I guess these things fall under: "If instead only a certain numerical threshold is needed, or only certain properties should be compared, I'd recursively apply a transformation / filtering to the dicts before comparison." -- what exactly do you mean by this?

ndaelman-hu:
> @ndaelman-hu thanks for your feedback!
> 
> I did originally try to use the native python comparisons, the original function I implemented a while ago, but I believe I could not simply compare 2 dictionaries with arbitrarily nested structures and containing various and a priori unknown types...(I will have to take a look again)
> 
> Maybe I over complicated things now, but a couple examples that I needed to treat in a special way:
> 
>     1. floats need to be compared with the `approx` function, and arrays of floats with `np.isclose()`
> 
>     2. np.inf is stored as a string in the archive -- I realize now that this is more of a problem with how I managed the data that I was comparing
> 
> 
> I guess these things fall under: "If instead only a certain numerical threshold is needed, or only certain properties should be compared, I'd recursively apply a transformation / filtering to the dicts before comparison." -- what exactly do you mean by this?

Indeed, in that case I'd use a recursive function, much as you are doing now, but with the type signature `dict -> dict`.
In that recursive function, I'd just catch the cases you mentioned above, convert them and leave the rest intact.

There shouldn't be any issue overwriting the values in-place, given that the object where `m_to_dict` originated from is stored separately. Else, you can use `deepcopy`.

Lastly, Python has a max. stack frames (1000). If you encounter a stack overload, feel free to reach out, and we look into it.

tests/test_force_field.py
```diff
@@ -0,0 +1,888 @@
...
+        d1 (dict): First dictionary to compare.
+        d2 (dict): Second dictionary to compare.
+    """
+
+    assert isinstance(d1, dict), f'Expected dict, got {type(d1)}'
+    assert isinstance(d2, dict), f'Expected dict, got {type(d2)}'
+    assert d1.keys() == d2.keys(), f'Keys mismatch: {d1.keys()} != {d2.keys()}'
+
+    def compare_unknown(key, value1, value2):
+        assert value1 is None, f"Unknown types'{key}': {type(value1)} != {type(value2)}"
```
Bernadette-Mohr, line 65: Are we sure here that `value1 = None` is the only problem we can ever encounter here, or should we assert that `value1` and `value2` have both the same type more generally?

tests/test_force_field.py
```diff
@@ -0,0 +1,888 @@
...
+    TabulatedPotential,
+    UreyBradleyAngle,
+)
+from nomad_simulations.schema_packages.general import Simulation
+from nomad_simulations.schema_packages.numerical_settings import ForceCalculations
+
+# from structlog.stdlib import BoundLogger
+from . import logger
+
+MOL = 6.022140857e23
```
Bernadette-Mohr, line 41: I'll research if we can extend the definition in `ureg`. I think it would be nice if we could get rid of `MOL` as a hard-coded global variable, right now I see it in many NOMAD classes.

Bernadette-Mohr (commented):
I don't have a lot to say for a review. We can discuss the scope and direction of further development in our meeting.

src/nomad_simulations/schema_packages/force_field.py
```diff
@@ -0,0 +1,1224 @@
...
+        """,
+    )
+
+    type = Quantity(
+        type=MEnum(
+            'bond',
+            'angle',
+            'bond-angle',
+            'dihedral',
+            'angle-dihedralimproper dihedral',
```
EBB2675: is this `'angle-dihedralimproper dihedral'` intended?

src/nomad_simulations/schema_packages/force_field.py
```diff
@@ -0,0 +1,1224 @@
...
+        super().normalize(archive, logger)
+
+        self.name = 'TabulatedBond'
+
+
+class AnglePotential(Potential):
+    """
+    Section containing information about angle potentials.
+
+    Suggested types are: harmonic, cosine, restricted_cosinse, fourier_series, urey_bradley, polynomial, tabulated
```
EBB2675: restricted_cosine

src/nomad_simulations/schema_packages/force_field.py
```diff
@@ -0,0 +1,1224 @@
...
+        if not self.functional_form:
+            self.functional_form = 'fourier_series'
+        elif self.functional_form != 'fourier_series':
+            logger.warning('Incorrect functional form set for FourierSeriesAngle.')
+
+
+# TODO I think we should name these more generally and then say that AKA
+class UreyBradleyAngle(AnglePotential):
+    r"""
+    Section containing information about a Urey-Bradly angle potential:
```
EBB2675: Bradley

src/nomad_simulations/schema_packages/force_field.py
```diff
@@ -0,0 +1,1224 @@
...
+
+        # set the dimensions based on the particle indices, if stored
+        if not self.n_interactions:
+            self.n_interactions = (
+                len(self.particle_indices)
+                if self.particle_indices is not None
+                else None
+            )
+        if not self.n_particles:
+            self.n_interactions = (
```
EBB2675: we are setting `self.n_interactions` again here, should the second assignment be `self.n_particles`

src/nomad_simulations/schema_packages/force_field.py
```diff
@@ -0,0 +1,1224 @@
...
+        unit='J',
+        shape=['*'],
+        description="""
+        List of energy values associated with each bin.
+        """,
+    )
+
+    forces = Quantity(
+        type=np.float64,
+        shape=[],
```
EBB2675: should it be of shape=['*'] as well?

src/nomad_simulations/schema_packages/force_field.py
```diff
@@ -0,0 +1,1224 @@
...
+    k_2 \left[ 1 + \cos(2(\theta - \theta_0)) \right] + \dots + C$,
+    where $\{k_1, k_2, \dots}$ are the `fourier_force_constants` for each term in the series
+    and $\theta_0$ is the `equilibrium_value` of the angle $\theta$.
+    $C$ is an arbitrary constant (not stored).
+    """
+
+    fourier_force_constants = Quantity(
+        type=np.float64,
+        shape=['*'],
+        unit='J / degree**2',
```
EBB2675: im a bit confused, sorry if this is a ridiculous comment. each cosine is dimensionless. The bracketed terms are dimensionless too. I had the feeling that each `k_m` must carry the units of energy so that `V(\theta)` has units of energy, ` unit='J'`

EBB2675 (changes requested):

src/nomad_simulations/schema_packages/force_field.py
EBB2675: also, isnt k_m technically a fourier coefficient?

EBB2675:
one last _rookie_ question, just for my own understanding

is it the domain convention to store angles as degrees? My thinking: trig and derivatives are defined for radians, so otherwise one needs hidden pi/180 factors in every case. 

If SI units: torques become cleanly J/rad, harmonic constants J/rad**2, and Fourier coefficients  in J

src/nomad_simulations/schema_packages/force_field.py
JFRudzinski: you are correct, it was just a mistake by me, I re-implemented this as RyckaertBellemansDihedral (Can you check this one and confirm it makes sense to you), I will remove this FourierSeriesAngle

src/nomad_simulations/schema_packages/force_field.py
```diff
@@ -0,0 +1,1191 @@
...
+        """,
+    )
+
+    type = Quantity(
+        type=MEnum(
+            'bond',
+            'angle',
+            'bond-angle',
+            'dihedral',
+            'angle-dihedral,improper dihedral',
```
EBB2675: i think its still here

JFRudzinski: I put the comma in the wrong place :facepalm:

src/nomad_simulations/schema_packages/force_field.py
```diff
@@ -0,0 +1,1192 @@
...
+            if self.forces is None and self.energies is not None:
+                try:
+                    # generate forces from energies numerically using spline
+                    self.forces = (
+                        self.compute_forces(
+                            self.bins.magnitude,
+                            self.energies.magnitude,
+                            smoothing_factor=smoothing_factor,
+                        )  # ? How do I deal with units here?
+                    )
```
EBB2675: indeed it crashes with `AttributeError: 'numpy.ndarray' object has no attribute 'magnitude'` here.

Would it be safer here to avoid `.magnitude` on bins/forces, since in the base class they may be plain ndarrays?

src/nomad_simulations/schema_packages/force_field.py
EBB2675: looks good from my side :+1:

src/nomad_simulations/schema_packages/force_field.py
ndaelman-hu: Is not a bad idea, but this means that you have to handle the stripping and reapplying of units externally. Or you ensure they're always provided in SI.

Personally, I'd use type hinting (`ureg.Quantity`) and handle the stripping-reapplying internally.
Granted, Python isn't statically type-checked, but at least we're using `mypy` in the CI/CD.

ndaelman-hu:
> one last _rookie_ question, just for my own understanding
> 
> is it the domain convention to store angles as degrees? My thinking: trig and derivatives are defined for radians, so otherwise one needs hidden pi/180 factors in every case.
> 
> If SI units: torques become cleanly J/rad, harmonic constants J/rad**2, and Fourier coefficients in J

It's a good question that has come up several times. Overall, NOMAD allows for users to choose their units.
By convention we use SI in the `archive` (does not store units directly), so it would end up being `rad`, I guess.

ndaelman-hu:
Was still receiving notifications for this PR. Didn't mean to butt in, but thought these replies might be helpful.
Will sign out from notifications now.

JFRudzinski:
> one last _rookie_ question, just for my own understanding
> 
> is it the domain convention to store angles as degrees? My thinking: trig and derivatives are defined for radians, so otherwise one needs hidden pi/180 factors in every case.
> 
> If SI units: torques become cleanly J/rad, harmonic constants J/rad**2, and Fourier coefficients in J

Not necessarily, it is common to use either depending on the case. I guess rad are more common in the technical usage but deg are typically used when plotting the potentials, etc.

In terms of the schema definitions, NOMAD will automatically convert for you in terms of parsing, but I guess the question is when users grab the data will they want it in radian, so I think you are right and I switched to rad now

src/nomad_simulations/schema_packages/force_field.py
JFRudzinski: I avoid the normalization/derivation now if no units + added a TODO to somehow prevent or discourage use of the base classes that don't have units

JFRudzinski:
@EBB2675 could you check my changes and approve this if OK now

src/nomad_simulations/schema_packages/force_field.py
```diff
@@ -0,0 +1,1212 @@
...
+                            smoothing_factor=smoothing_factor,
+                        )
+                        * self.energies.units
+                    )
+
+                    energies_diff = energies.to('kJ').magnitude * MOL - (
+                        self.energies.to('kJ').magnitude * MOL
+                        - np.min(self.energies.to('kJ').magnitude * MOL)
+                    )
+                    if np.all([x < tol for x in energies_diff]):
```
EBB2675: only one minor question: should the consistency check use an absolute comparison with `np.abs` ? With `x<tol`, large negative deviations would pass here

EBB2675 (approved):

PR #117, opened 2024-09-13
Added check_simulation_cell in utils

JosePizarro3:
@JFRudzinski @ladinesa I added this utils function to check if a simulation cell is the same. "Same" here means that I check the positions in the lattice + the chemical symbol to be the same in each position. I am sure this can be extended, but for now, I need this when resolving workflows. 

@ladinesa would you mind taking a quick look? I am pretty sure the check of the positions can be optimized, but I was lazy + ChatGPT is suggesting the wrong stuff (basically using numpy functions with some args that cannot be passed...).


src/nomad_simulations/schema_packages/utils/utils.py
```diff
@@ -168,3 +170,54 @@ def get_composition(children_names: 'list[str]') -> str:
...
+    # The `positions` should have the same length (same number of positions)
+    if len(cell_1.positions) != len(cell_2.positions):
+        return False
+    n_positions = len(cell_1.positions)
+
+    # Check that all the `positions`` of `cell_1` match with the ones in `cell_2`
+    check_positions = []
+    for i1, pos1 in enumerate(cell_1.positions):
+        for i2, pos2 in enumerate(cell_2.positions):
+            if np.allclose(pos1, pos2):
```
ladinesa: are you not supposed to return False here is they are not close?

src/nomad_simulations/schema_packages/utils/utils.py
```diff
@@ -168,3 +170,54 @@ def get_composition(children_names: 'list[str]') -> str:
...
+    if len(check_positions) != n_positions:
+        return False
+
+    # If the cells are not `AtomicCell`, finish function (using m_def.name to avoid circular import problems)
+    if cell_1.m_def.name == 'Cell' and cell_2.m_def.name == 'Cell':
+        return True
+    if cell_1.m_def.name != 'AtomicCell' or cell_2.m_def.name != 'AtomicCell':
+        return False
+
+    # Check that the `chemical_symbol` of the atoms in `cell_1` match with the ones in `cell_2`
```
ladinesa: test for chem symbols is easier so it should precede positions

src/nomad_simulations/schema_packages/utils/utils.py
```diff
@@ -168,3 +170,54 @@ def get_composition(children_names: 'list[str]') -> str:
...
+    n_positions = len(cell_1.positions)
+
+    # Check that all the `positions`` of `cell_1` match with the ones in `cell_2`
+    check_positions = []
+    for i1, pos1 in enumerate(cell_1.positions):
+        for i2, pos2 in enumerate(cell_2.positions):
+            if np.allclose(pos1, pos2):
+                check_positions.append([i1, i2])
+                break
+    if len(check_positions) != n_positions:
```
ladinesa: easier to check length so it should go before.

src/nomad_simulations/schema_packages/utils/utils.py
ladinesa: add tolerance here for control

src/nomad_simulations/schema_packages/utils/utils.py
```diff
@@ -168,3 +170,54 @@ def get_composition(children_names: 'list[str]') -> str:
     children_count_tup = np.unique(children_names, return_counts=True)
     formula = ''.join([f'{name}({count})' for name, count in zip(*children_count_tup)])
     return formula if formula else None
+
+
+def check_simulation_cell(cell_1: 'Cell', cell_2: 'Cell') -> bool:
```
ladinesa: i would rename it is_same or is_equal_cell

src/nomad_simulations/schema_packages/utils/utils.py
```diff
@@ -168,3 +170,54 @@ def get_composition(children_names: 'list[str]') -> str:
...
+    for i1, pos1 in enumerate(cell_1.positions):
+        for i2, pos2 in enumerate(cell_2.positions):
+            if np.allclose(pos1, pos2):
+                check_positions.append([i1, i2])
+                break
+    if len(check_positions) != n_positions:
+        return False
+
+    # If the cells are not `AtomicCell`, finish function (using m_def.name to avoid circular import problems)
+    if cell_1.m_def.name == 'Cell' and cell_2.m_def.name == 'Cell':
```
ladinesa: these should go up , they are easier

src/nomad_simulations/schema_packages/utils/utils.py
```diff
@@ -168,3 +170,54 @@ def get_composition(children_names: 'list[str]') -> str:
...
+        return False
+    if cell_1.positions is None or cell_2.positions is None:
+        return False
+
+    # The `positions` should have the same length (same number of positions)
+    if len(cell_1.positions) != len(cell_2.positions):
+        return False
+    n_positions = len(cell_1.positions)
+
+    # Check that all the `positions`` of `cell_1` match with the ones in `cell_2`
```
ladinesa: are you not going to check for translated postions for periodic systems? mybe with ase atoms wrap?

src/nomad_simulations/schema_packages/utils/utils.py
ladinesa: i think this is generic enough that we can put in atomutils. I have a similar function in simulationworkflowschema

src/nomad_simulations/schema_packages/utils/utils.py
```diff
@@ -168,3 +170,54 @@ def get_composition(children_names: 'list[str]') -> str:
...
+    check_positions = []
+    for i1, pos1 in enumerate(cell_1.positions):
+        for i2, pos2 in enumerate(cell_2.positions):
+            if np.allclose(pos1, pos2):
+                check_positions.append([i1, i2])
+                break
+    if len(check_positions) != n_positions:
+        return False
+
+    # If the cells are not `AtomicCell`, finish function (using m_def.name to avoid circular import problems)
```
ladinesa: m_def name is unreliable, i still like type check maybe local import?

src/nomad_simulations/schema_packages/utils/utils.py
JosePizarro3: The logic here goes first with `Cell` and `Cell.positions`, and then it checks the chemical symbols afterward. The crystal lattice and a material is not the same thing

ladinesa: checking positions is more time consuming than symbols right?

JosePizarro3: but the cell only contains positions, not symbols

ladinesa: ah in atoms_state, but why dont you check all atoms in atoms_state?

src/nomad_simulations/schema_packages/utils/utils.py
JosePizarro3: Well, this is checking a newly created list. 

What I have is 2 lists of positions for each of the cells -> I store in a list the index at which the first and second positions coincide -> if the new list is of length `n_positions` it means that the both cells contain the same positions -> otherwise, there are some missing positions from one to another cell and I should return False

src/nomad_simulations/schema_packages/utils/utils.py
JosePizarro3: This goes with the logic of "first check `Cell`, then check if they are `AtomicCell`". These ifs are to control the cases where the cells are `Cell` and then return True if so (after all the checks before).

src/nomad_simulations/schema_packages/utils/utils.py
JosePizarro3: I am thinking you are right... with ASE this would be simpler so I'll check if they have the == logic done to compare `ase.Atoms`

src/nomad_simulations/schema_packages/utils/utils.py
JosePizarro3: I am checking one position of the `cell_1` with respect to all the other positions in `cell_2`, so no, I can only continue.

This implementation I did is not optimized at all... I would work on it, but first, I'll check ase for their implementation.

JosePizarro3:
Hi, so I checked ASE and the only option I saw is to directly check the positions and chemical symbols by using the functions `get_positions` and `get_chemical_symbols`. But this would only work for AtomicCell; our Cell is a more abstract object than Atoms, but it is more correct in our situation imo (we are considering the lattice to be an abstraction to the material, which is what is formally correct).

Do you mind doing a final swap and telling me if something is off?

ladinesa:
> Hi, so I checked ASE and the only option I saw is to directly check the positions and chemical symbols by using the functions `get_positions` and `get_chemical_symbols`. But this would only work for AtomicCell; our Cell is a more abstract object than Atoms, but it is more correct in our situation imo (we are considering the lattice to be an abstraction to the material, which is what is formally correct).
> 
> Do you mind doing a final swap and telling me if something is off?

Looks good but I thought you would implement `__eq__` for Cell? Also I would really use typing checking instead of checking for m_def.name, use local import to rid of cyclic de.

JosePizarro3:
> Looks good but I thought you would implement `__eq__` for Cell? Also I would really use typing checking instead of checking for m_def.name, use local import to rid of cyclic de.

Ah true, sorry I didn't understand this part before. I can implement that and it makes all the sense.

JosePizarro3:
Much better now. What do you think?

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -296,6 +296,32 @@ class Cell(GeometricSpace):
...
+        for i1, pos1 in enumerate(positions_1):
+            for i2, pos2 in enumerate(positions_2):
+                if np.allclose(
+                    pos1, pos2, atol=configuration.equal_cell_positions_tolerance
+                ):
+                    check_positions.append([i1, i2])
+                    break
+        return check_positions
+
+    def __eq__(self, other: 'Cell') -> bool:
```
ladinesa: you would also define `__ne__` which is simply not `self.__eq__`

src/nomad_simulations/schema_packages/utils/utils.py
```diff
@@ -168,3 +175,25 @@ def get_composition(children_names: 'list[str]') -> str:
...
+        cell_2 (Cell): The second `Cell` to compare.
+
+    Returns:
+        bool: True if the cells are the same, False otherwise.
+    """
+    # TODO extend this function to compare more information of the cells (`lattice_vectors`) and check ase.Atoms functions
+    # If any of the cells is None or the positions are empty, return False
+    if cell_1 is None or cell_2 is None:
+        return False
+    if cell_1.positions is None or cell_2.positions is None:
```
ladinesa: why not put this also in `__eq__` so you completely get rid of this function,

ladinesa:
> Much better now. What do you think?

indeed just a couple more suggestions.

src/nomad_simulations/schema_packages/model_system.py
JosePizarro3: You mean `return not self.__eq__` in the __ne__?

ladinesa: yes indeed `return not self.__eq__(other)`, this is required when doing !=

JosePizarro3:
Done! Many thanks for the help 🙂

PR #118, opened 2024-09-13
Add protection to `model_system` normalization in case the list is empty

ndaelman-hu:
This error was triggered when testing the mapping annotation, and is a valid security feature.


JosePizarro3:
`if not self.model_system:`

ndaelman-hu:
The same correction likely applies to a bunch of other cases. I can see to correct all of them, next week.

ndaelman-hu:
> `if not self.model_system:`

I take it that with this correction, it has your implicit approval. Will bypass the security to merge.

PR #122, opened 2024-09-20
Extend equalities with comparison

ndaelman-hu:
### Changes

This is an extension to commit 96f4a91, defining also `<`, `>`, `<=`, `<=` for `Cell` and `AtomicCell`.
It builds the definition up functor-wise, and introduces `PartialOrderElement` as the foundational comparison type, to power partial orders. Note that incomparable elements of the same class return `False` instead of `NotImplemented`.

### Background and motivation
 
@JFRudzinski knows this, but early on this year I've been working on sorting hierarchies like `ModelSystem`. This work wasn't merged due to incompleteness and changing priorities. The comparison metrics matched those applied in commit 96f4a91. The other operators are necessary for sorting, though.

### DIFF
--- a/src/nomad_simulations/schema_packages/__init__.py
+++ b/src/nomad_simulations/schema_packages/__init__.py
@@ -31,8 +31,8 @@ class NOMADSimulationsEntryPoint(SchemaPackageEntryPoint):
         description='Limite of the number of atoms in the unit cell to be treated for the system type classification from MatID to work. This is done to avoid overhead of the package.',
     )
     equal_cell_positions_tolerance: float = Field(
-        1e-12,
-        description='Tolerance (in meters) for the cell positions to be considered equal.',
+        12,
+        description='Decimal order or tolerance (in meters) for comparing cell positions.',
     )
 
     def load(self):
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -225,25 +225,25 @@ class GeometricSpace(Entity):
             return
 
 
+def _check_implemented(func: 'Callable'):
+    """
+    Decorator to restrict the comparison functions to the same class.
+    """
+
+    def wrapper(self, other):
+        if not isinstance(other, self.__class__):
+            return NotImplemented
+        return func(self, other)
+
+    return wrapper
+
+
 class PartialOrderElement:
     def __init__(self, representative_variable):
         self.representative_variable = representative_variable
 
     def __hash__(self):
         return self.representative_variable.__hash__()
-
-    @staticmethod
-    def _check_implemented(func: 'Callable'):
-        """
-        Decorator to restrict the comparison functions to the same class.
-        """
-
-        def wrapper(self, other):
-            if not isinstance(other, self.__class__):
-                return NotImplemented
-            return func(self, other)
-
-        return wrapper
 
     @_check_implemented
     def __eq__(self, other):
@@ -398,7 +398,7 @@ class Cell(GeometricSpace):
 
     def is_ne_cell(self, other) -> bool:
         # this does not hold in general, but here we use finite sets
-        return (not self.is_equal_cell(other))
+        return not self.is_equal_cell(other)
 
     def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
         super().normalize(archive, logger)
@@ -644,7 +644,10 @@ class Symmetry(ArchiveSection):
     )
 
     def resolve_analyzed_atomic_cell(
-        self, symmetry_analyzer: 'SymmetryAnalyzer', cell_type: str, logger: 'BoundLogger'
+        self,
+        symmetry_analyzer: 'SymmetryAnalyzer',
+        cell_type: str,
+        logger: 'BoundLogger',
     ) -> 'Optional[AtomicCell]':
         """
         Resolves the `AtomicCell` section from the `SymmetryAnalyzer` object and the cell_type
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -24,6 +24,7 @@ from nomad.units import ureg
 if TYPE_CHECKING:
     from nomad.datamodel.datamodel import EntryArchive
     from nomad.metainfo import Context, Section
+    from nomad.ureg import pint
     from structlog.stdlib import BoundLogger
 
 from nomad_simulations.schema_packages.atoms_state import AtomsState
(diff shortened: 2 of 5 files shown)


ndaelman-hu:
I commented out some ill-defined test cases. These should either return `False` or `NotImplemented` across the board, though there are various subcases.

I can elaborate further in a later post, but 1 example would be `None`: should this always be interpreted as `[]` for repeatable quantities / sections? If so, are we ensured that this is always enforced?

JosePizarro3:
I am not sure I understand the logic of `<` and `>` here... how did you decide that a Cell is larger/smaller than another by just looking at how many positions are defined in the lattice?

Also, the test you deleted should be added back.

I will review this in some time. If @JFRudzinski or @ladinesa want to do it before me, feel free.

ndaelman-hu:
> I am not sure I understand the logic of < and > here... how did you decide that a Cell is larger/smaller than another by just looking at how many positions are defined in the lattice?

No, the comparison follows the definition of _subsets_, with `positions` vectors defined as individual elements (in the case of `Cell`).
E.g. `[2, 0, 0]` is not gt `[1, 0, 0]`. One cell is gt another, if it contains all the former's positions.

JosePizarro3:
> > I am not sure I understand the logic of < and > here... how did you decide that a Cell is larger/smaller than another by just looking at how many positions are defined in the lattice?
> 
> No, the comparison follows the definition of _subsets_, with `positions` vectors defined as individual elements (in the case of `Cell`). E.g. `[2, 0, 0]` is not gt `[1, 0, 0]`. One cell is gt another, if it contains all the former's positions.

Yeah, but I would not use inequalities here. I would naively think that Cell > Cell means the volume is bigger or something...

What is this useful for? Please, give me a concrete and easy-to-read example, no need to go into a deep detail if it is not needed.

ndaelman-hu:
> Also, the test you deleted should be added back.

No tests were deleted, some are just commented out pending a discussion. I kind of touched on this in my previous comment, but I'll elaborate. Essentially, there are 3 cases where I don't see the sense of forcing a comparison. Once we reach consensus, we can then decide how to treat such cases, i.e. `False`, `None`, or `NotImplemented`. These are different aspects.

### 1. Comparing `Cell` with `AtomicCell`.

Since `chemical_symbols` are relevant when comparing two `AtomicCell`s, a `Cell` can never equate with an `AtomicCell`. I think there's little dispute here.

I currently enforce that the comparisons are all homogeneous operators, so defaulting to a `False` response falls outside that scope. Would `NotImplemented` not be more in line with the Python convention here?

If they need so, the user should first convert `AtomicCell` to a `Cell` (or vice versa) and then compare.

### 2. Comparing two `Cell`s, where one has `Cell.positions is None`.

The current definition does support `Cell.positions==[]`. I'm just not sure whether `None` should be treated the same.
A `None` scenario is indicative of only a partial instantiation or a missing normalization. `[]`, meanwhile, shows that the parser populated `Cell.positions` and is a completely valid case.

### 3. Both `Cell.positions is None`

This is an extension to point 2. Python's native (and often naive) implementation would be that these are indeed the same.
In that case, this would be an exception to the exception... I think `NotImplemented` conveys the message better: populate or normalize first before comparing.

JosePizarro3:
> > Also, the test you deleted should be added back.
> 
> No tests were deleted,...

All these are just to ensure there were no errors during parsing or at any other moment, and hence one Cell is missing information (like `positions`) or there is a weird comparison of Cell with AtomicCell.

Furthermore, I disagree with `Cell.positions=[]`; you cannot do that due to validation of the shape of the Quantity.

ndaelman-hu:
> > > I am not sure I understand the logic of < and > here... how did you decide that a Cell is larger/smaller than another by just looking at how many positions are defined in the lattice?
> > 
> > 
> > No, the comparison follows the definition of _subsets_, with `positions` vectors defined as individual elements (in the case of `Cell`). E.g. `[2, 0, 0]` is not gt `[1, 0, 0]`. One cell is gt another, if it contains all the former's positions.
> 
> Yeah, but I would not use inequalities here. I would naively think that Cell > Cell means the volume is bigger or something...

Good point, though the same would also apply to `==`, right?

The choice of what attributes to compare eventually depends on the use case (see below). In and of itself, there's no issue adding the volume to that list.
If a custom schema requires a different definition, then it can overwrite the default.
If we support multiple use cases that have clashing attributes, we could step away from the dunder methods and just define regular methods.

> What is this useful for? Please, give me a concrete and easy-to-read example, no need to go into a deep detail if it is not needed.

Automatically building a system hierarchy tree.

Given `[AtomicCell(<TiO2 slab>), AtomicCell(<Al2O3 slab>), AtomicCell(<full system: TiO2-Al2O3 interface>)]` automatically generate the tree

```
AtomicCell(<full system: TiO2-Al2O3 interface>)
├── AtomicCell(<TiO2 slab>)
└── AtomicCell(<Al2O3 slab>)
```

Of course, if the previous implementation of `==` serves some particular cases, pls mention them too. Those are surely relevant.

JosePizarro3:
> > Yeah, but I would not use inequalities here. I would naively think that Cell > Cell means the volume is bigger or something...
> 
> Good point, though the same would also apply to `==`, right?
>
> The choice of what attributes to compare eventually depends on the use case (see below). In and of itself, there's no issue adding the volume to that list. If a custom schema requires a different definition, then it can overwrite the default. If we support multiple use cases that have clashing attributes, we could step away from the dunder methods and just define regular methods.

Yes: `positions` and `lattice_vectors` are key here. I wanted to add only `positions` to check the first cases, but probably I will add the volume via the `lattice_vectors` as well.


> > What is this useful for? Please, give me a concrete and easy-to-read example, no need to go into a deep detail if it is not needed.
> 
> Automatically building a system hierarchy tree.
> 
> Given `[AtomicCell(<TiO2 slab>), AtomicCell(<Al2O3 slab>), AtomicCell(<full system: TiO2-Al2O3 interface>)]` automatically generate the tree
> 
> ```
> AtomicCell(<full system: TiO2-Al2O3 interface>)
> ├── AtomicCell(<TiO2 slab>)
> └── AtomicCell(<Al2O3 slab>)
> ```

But then you should add the operators in `ModelSystem`, not in `Cell`. And if it is just a mater of a hierarchical tree, why not simply comparing lengths of positions across cells?

ndaelman-hu:
> > > Also, the test you deleted should be added back.
> > 
> > 
> > No tests were deleted,...
> 
> All these are just to ensure there were no errors during parsing or at any other moment, and hence one Cell is missing information (like `positions`) or there is a weird comparison of Cell with AtomicCell.

But wouldn't a failing comparison just be indicative of on an underlying bug or "weird"/inappropriate usage on the dev's side? Returning `False` won't amend that.

Btw, `NotImplemented` will prompt Python to consider alternative approaches that are covered, e.g. `not ==` if `!=` is not defined. If all else fails, it will return a `NotImplementedError`.

> Furthermore, I disagree with `Cell.positions=[]`; you cannot do that due to validation of the shape of the Quantity.

Sure, but it's not up to the comparison operators to enforce `shape`.
Regarding the choice of example, everything I said applied also to `[[]]` or `[[], []]`. Both will be considered equal, btw.

ndaelman-hu:
> Yes: positions and lattice_vectors are key here. I wanted to add only positions to check the first cases, but probably I will add the volume via the lattice_vectors as well.

I can add them in this PR, no prob.

> But then you should add the operators in ModelSystem, not in Cell.

Okay, I'll take a look at that. Still, I use comparisons between smaller components to build up the larger ones, so the extension to `Cell` may still be relevant.

> And if it is just a mater of a hierarchical tree, why not simply comparing lengths of positions across cells?

No, that's not good, it really has to work like `subset`: we only want systems that are fully contained within another to be hierarchically linked.
This implies that not all cases can be resolved / are comparable. This is indeed what created issues for me the 1st time around.
In the meantime, I've been learning about the maths of "partial orders", of which subset is indeed the most prominent example.

Anyhow, I should've merged by original implementation for comparing back then, regardless. That's why I'm correcting this now.
The sorting will be a separate commit.

ndaelman-hu:
@JosePizarro3 You mentioned some use cases for your comparison operator last week. Could you pls repeat them here, so I account for them in the testing?

@JFRudzinski @JosePizarro3 pls review again.

ndaelman-hu:
Feedback round is closed. I will make 1 amendment and merge tomorrow.

JosePizarro3:
Hi, I can give you a final review tonight. But, did you rebase?

ndaelman-hu:
> Hi, I can give you a final review tonight.

Thx.

>  But, did you rebase?

Yes, I did.

JosePizarro3:
Capture

Clearly, you did not.

JFRudzinski (approved):

JFRudzinski:
In case it's not clear: my approval is for the code changes, obviously the conflicts need to resolved before merging.

ndaelman-hu:
Conflicts resolved.

PR #125, opened 2024-09-23
Fix equal operator in cell

JosePizarro3:
Following @TLCFEM advice, I changed `__eq__` (and `__ne__`) to a non-dunder method, `is_equal_cell`


JosePizarro3:
Closed in favor of #127

PR #134, opened 2024-09-25
131 add mkdocstrings to docu

### LINKED ISSUE #131: Add mkdocstrings to docu
https://mkdocstrings.github.io/usage/

JosePizarro3:
@JFRudzinski @Bernadette-Mohr or @EBB2675 @ndaelman-hu, would you mind taking a look?

You can locally launch the MKDocs page (check the README instructions), and see the new `References` section. Bear in mind we can change the style, but I do not want to spend to much time on this before deciding how the layout of the references should look like.


JosePizarro3:
@JFRudzinski maybe you want to take a look on this and give improvement feedback.

I was also thinking that this docu generated shows that we need to describe better our objects.

JFRudzinski:
just a note to the others: after I did 
```sh
uv pip install -r requirements_docs.txt
```
I get `ERROR   -  Config value 'plugins': The "mkdocstrings" plugin is not installed` when I try to execute mkdocs. I just had to deactivate my env, and then reactivate and then it worked.

JFRudzinski:
- [ ] All of the `http://purl.obolibrary.org/obo` links are broken

JFRudzinski:
Screenshot from 2024-10-04 13-17-58

~~Is this the intended effect?  Is it possible to change the format of the overview, e.g., display the overall type and then attributes in some sort of table or something. ATM it is easier to just read the source code.~~

~~Also, can we reduce the font size of the quantities? It is considerably larger than the normal font size throughout the docs~~

Sorry, I didn't see at first your comment about formatting.

I would suggest maybe something like:

name (Quantity(str)): 
    \<description\>

Not sure about for the ELN stuff

JFRudzinski:
Screenshot from 2024-10-04 13-21-48

This is potentially very useful if we can leverage the admonisions in our descriptions. I wonder if we can also integrate  some sort of equation display? That would be very nice

JFRudzinski:
fixed a few things: https://github.com/nomad-coe/nomad-simulations/pull/142

mostly I just added open is new browser tab for links to external sites

JFRudzinski:
> improvement feedback.
> 
> I was also thinking that this docu generated shows that we need to describe b

I definitely agree about improving the descriptions. I think for me it would be easiest to adjust the format first so that the description is easier to read, and then I could start going through and trying to make things more consistent/descriptive and tagging things that are unclear.

I could start this sooner, but then I would probably just go through the code and not use the docs. I think the former is better though cause then we can start utilizing the admonitions within the descriptions as mentioned above

What do you think?

JosePizarro3:
> > improvement feedback.
> > I was also thinking that this docu generated shows that we need to describe b
> 
> I definitely agree about improving the descriptions. I think for me it would be easiest to adjust the format first so that the description is easier to read, and then I could start going through and trying to make things more consistent/descriptive and tagging things that are unclear.
> 
> I could start this sooner, but then I would probably just go through the code and not use the docs. I think the former is better though cause then we can start utilizing the admonitions within the descriptions as mentioned above
> 
> What do you think?

Yeah, agreed. I will take a look on formatting better this mkdocs page. And then we improve the descriptions in a subsequent pr so the docu will show nicely the results.

JosePizarro3:
@JFRudzinski I was playing with MkDocs a bit more in another context, I think I can improve the documentation in this pr.

However, I wondered whether we are going to change the classes too much to make even sense. I guess we could discuss this next year in a meeting with the others like you suggested.

JFRudzinski:
> @JFRudzinski I was playing with MkDocs a bit more in another context, I think I can improve the documentation in this pr.
> 
> However, I wondered whether we are going to change the classes too much to make even sense. I guess we could discuss this next year in a meeting with the others like you suggested.

Oh ok, nice. I would very much like to learn about this (for other plugin docs as well). But yeah I agree we should wait till things are a bit more stable. 

And yes, let's set a meeting starting next year, I will contact you.

JFRudzinski:
closing in favor of #267

PR #140, opened 2024-09-27
Add some methods for the cell

### LINKED ISSUE #136: Add some methods for the cell
I am going to add some methods to `ModelSystem` and/or `Cell`/`AtomicCell` to make our lives easier during parsing:

- `from_ase()`: right now we have `to_ase`, but we can define the "from" one.
- `get_atom_symbols()`: this one is to pass the `AtomsState(chemical_symbol)` information to a flat list (as it was originally in the schema).

@ndaelman-hu @EBB2675 please, feel free to suggest other methods that you think might be interesting as well

JosePizarro3:
@ndaelman-hu You can check it out, but being this small, I think I can merge without a review.

### DIFF
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -391,16 +391,26 @@ class AtomicCell(Cell):
             return False
         return True
 
-    def get_chemical_symbols(self) -> list:
+    def get_chemical_symbols(self, logger: 'BoundLogger') -> list[str]:
         """
         Get the chemical symbols of the atoms in the atomic cell. These are defined on `atoms_state[*].chemical_symbol`.
+
+        Args:
+            logger (BoundLogger): The logger to log messages.
 
         Returns:
             list: The list of chemical symbols of the atoms in the atomic cell.
         """
         if not self.atoms_state:
             return []
-        return [atom_state.chemical_symbol for atom_state in self.atoms_state]
+
+        chemical_symbols = []
+        for atom_state in self.atoms_state:
+            if not atom_state.chemical_symbol:
+                logger.warning('Could not find `AtomsState[*].chemical_symbol`.')
+                return []
+            chemical_symbols.append(atom_state.chemical_symbol)
+        return chemical_symbols
 
     def to_ase_atoms(self, logger: 'BoundLogger') -> Optional[ase.Atoms]:
         """
@@ -414,7 +424,7 @@ class AtomicCell(Cell):
             (Optional[ase.Atoms]): The ASE Atoms object with the basic information from the `AtomicCell`.
         """
         # Initialize ase.Atoms object with labels
-        atoms_labels = self.get_chemical_symbols()
+        atoms_labels = self.get_chemical_symbols(logger=logger)
         ase_atoms = ase.Atoms(symbols=atoms_labels)
 
         # PBC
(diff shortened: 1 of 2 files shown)


src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -391,6 +391,17 @@ def is_equal_cell(self, other) -> bool:
             return False
         return True
 
+    def get_chemical_symbols(self) -> list:
```
ndaelman-hu: I think that the chemical symbols are always strings, so the output type signature can be `list[str]` even

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -391,6 +391,17 @@ def is_equal_cell(self, other) -> bool:
...
+        """
+        Get the chemical symbols of the atoms in the atomic cell. These are defined on `atoms_state[*].chemical_symbol`.
+
+        Returns:
+            list: The list of chemical symbols of the atoms in the atomic cell.
+        """
+        if not self.atoms_state:
+            return []
+        return [atom_state.chemical_symbol for atom_state in self.atoms_state]
```
ndaelman-hu: Wouldn't this raise an `AttributeError` if `atom_state.chemical_symbol` is not set? It could also result in `None`, which should be filtered out too.

Overall, I'm also quite a fan list comprehensions. When dealing with edge cases in Python, however, they often fall short... There are languages that do a better job at combining both.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -436,6 +447,35 @@ def to_ase_atoms(self, logger: 'BoundLogger') -> Optional[ase.Atoms]:
...
+        """
+        Parses the information from an ASE Atoms object to the `AtomicCell` section.
+
+        Args:
+            ase_atoms (ase.Atoms): The ASE Atoms object to parse.
+            logger (BoundLogger): The logger to log messages.
+        """
+        # `AtomsState[*].chemical_symbol`
+        chemical_symbols = ase_atoms.get_chemical_symbols()
+        for symbol in chemical_symbols:
```
ndaelman-hu: if the return type `ase_atoms.get_chemical_symbols` is `list[str]`, then you can contract these lines into 1: no need for single-use varirable assignment

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -436,6 +447,35 @@ def to_ase_atoms(self, logger: 'BoundLogger') -> Optional[ase.Atoms]:
...
+            atom_state = AtomsState(chemical_symbol=symbol)
+            self.atoms_state.append(atom_state)
+
+        # `periodic_boundary_conditions`
+        self.periodic_boundary_conditions = ase_atoms.get_pbc()
+
+        # `lattice_vectors`
+        cell = ase_atoms.get_cell()
+        self.lattice_vectors = ase.geometry.complete_cell(cell) * ureg('angstrom')
```
ndaelman-hu, line 478: same comment, you can contract into 1 line: the single-use varaible doesn't help the semantics, so it's irrelevant

ndaelman-hu (commented):
Pls double-check how `get_chemical_symbols` handles missing attributes and `None` values in `atom_state.chemical_symbol`.

That point aside, no need for further review.

JosePizarro3:
Thanks for the comments!

PR #144, opened 2024-10-16
Remove check for `n_` quantities

ndaelman-hu:


ndaelman-hu:
Cloased in favor of #146

PR #145, opened 2024-11-01
Fix shape mismatch of bond_list in model_system

Bernadette-Mohr:
The bond list is a list of lists or tuples with two elements each, containing atom indices. It was erroneously set to a scalar.


JFRudzinski (approved):
Looks good, you can merge this :+1:

PR #148, opened 2024-11-22
Cg particle support

Bernadette-Mohr:
Addition of a class `ParticleCell` to `model_system` for the parsing of coarse-grained representations.
The normalizer still calls `atomic_cell.to_ase_atoms`; somewhere, `particle_cell` must be automatically changed to `atomic_cell.`

### DIFF
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -27,6 +27,7 @@ if TYPE_CHECKING:
     from structlog.stdlib import BoundLogger
 
 from nomad_simulations.schema_packages.atoms_state import AtomsState
+from nomad_simulations.schema_packages.particles_state import ParticlesState
 from nomad_simulations.schema_packages.utils import (
     get_sibling_section,
     is_not_representative,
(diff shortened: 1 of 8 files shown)


src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -492,6 +493,168 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+            - [0, 0, 0, 3] three equivalent atoms and one non-equivalent.
+        """,
+    )
+
+    def __init__(self, m_def: 'Section' = None, m_context: 'Context' = None, **kwargs):
+        super().__init__(m_def, m_context, **kwargs)
+        # Set the name of the section
+        self.name = self.m_def.name  #! self.name here is 'ParticleCell'
+
+    def is_equal_cell(self, other) -> bool:
```
JFRudzinski: What exactly is this used for in atomic_cell? Is it for defining different reference frames?

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -492,6 +493,168 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+                type_2 = other.particles_state[particle[1]].particle_type
+                if type_1 != type_2:
+                    return False
+        except Exception:
+            return False
+        return True
+
+    def get_particle_types(self, logger: 'BoundLogger') -> list[str]:
+        """
+        Get the chemical symbols of the atoms in the atomic cell. These are defined on `atoms_state[*].chemical_symbol`.
```
JFRudzinski: maybe make a TODO that you need to go through and adjust all the atom->particle wording

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -492,6 +493,168 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+
+        Returns:
+            (Optional[Particles]): The Partilces object with the basic information from the `ParticleCell`.
+        """
+        # Initialize Partilces object with labels
+        particle_labels = self.get_particle_types(logger=logger)
+        particles = Particles(symbols=particle_labels)
+
+        # PBC
+        if self.periodic_boundary_conditions is None:
```
JFRudzinski, line 657: I am assuming a lot of these functionalities are shared with atomic_cell. I am wondering if they can be taken up to cell, or defined consistently between the classes in some other way?

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1017,6 +1181,7 @@ class ModelSystem(System):
     # TODO improve description and add an example using the case in atom_indices
     bond_list = Quantity(
         type=np.int32,
+        shape=['*', 2],
```
JFRudzinski, line 1250: I thought this was already fixed? You could maybe just make another quick PR to deal with this separately and merge immediately.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1137,11 +1305,16 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
                     sec_symmetry = self.m_create(Symmetry)
                     sec_symmetry.normalize(archive, logger)
 
+        print(self.model_system)
```
JFRudzinski: Do these mean that you have tested this with your parser? If not how are you testing? And with normalization?

JFRudzinski (changes requested):
Upon initial glance, I think the direction is looking good. Could you maybe just list out here any major changes that you have made compared with AtomicCell? It would be easy for me to look into those directly once I know.

Another thing that comes to mind is units. For gromacs or other sims using physical units we can get away with this for now, but we need to figure out a bigger solution for reduced units and such. This is not really on you, we need to have a discussion with Area D eventually. For now, just think about the different situations with CG units that you know.

Perhaps as a minimum effort to indicate that units may not be meaningful, using ParticleCell could trigger some sort of warning, or populate some metadata upstream, i.e., in Simulation(), about this :thinking:

src/nomad_simulations/schema_packages/model_system.py
Bernadette-Mohr: I don't think it is even used in `nomad` `AtomicCell`. I don't see an issue with removing it.

src/nomad_simulations/schema_packages/model_system.py
Bernadette-Mohr: That would make sense as far as possible. To prevent errors with CG systems, we have to check carefully for all situations where functions from `ase` are called and keep those duplicates.

JFRudzinski:
Covered now in #148 #211 #216 #222 etc

PR #150, opened 2024-12-04
initial investigation into harmonizing particles and atoms

JFRudzinski:
@Bernadette-Mohr

### DIFF
--- a/src/nomad_simulations/schema_packages/atoms_state.py
+++ b/src/nomad_simulations/schema_packages/atoms_state.py
@@ -552,7 +552,17 @@ class HubbardInteractions(ArchiveSection):
                 )
 
 
-class AtomsState(Entity):
+class State(Entity):
+    """
+    A base section to define the state information of the system.
+    """
+
+    def __init__(self, m_def: 'Section' = None, m_context: 'Context' = None, **kwargs):
+        super().__init__(m_def, m_context, **kwargs)
+        self.labels = None
+
+
+class AtomsState(State):
     """
     A base section to define each atom state information.
     """
@@ -641,3 +651,7 @@ class AtomsState(Entity):
             self.chemical_symbol = self.resolve_chemical_symbol(logger=logger)
         if self.atomic_number is None:
             self.atomic_number = self.resolve_atomic_number(logger=logger)
+
+        # Set the labels
+        if self.chemical_symbol is not None:
+            self.labels = self.chemical_symbol
--- a/src/nomad_simulations/schema_packages/atoms_state.py
+++ b/src/nomad_simulations/schema_packages/atoms_state.py
@@ -559,7 +559,6 @@ class State(Entity):
 
     def __init__(self, m_def: 'Section' = None, m_context: 'Context' = None, **kwargs):
         super().__init__(m_def, m_context, **kwargs)
-        self.labels = None
 
 
 class AtomsState(State):
@@ -651,7 +650,3 @@ class AtomsState(State):
             self.chemical_symbol = self.resolve_chemical_symbol(logger=logger)
         if self.atomic_number is None:
             self.atomic_number = self.resolve_atomic_number(logger=logger)
-
-        # Set the labels
-        if self.chemical_symbol is not None:
-            self.labels = self.chemical_symbol
--- a/src/nomad_simulations/schema_packages/general.py
+++ b/src/nomad_simulations/schema_packages/general.py
@@ -225,7 +225,9 @@ class Simulation(BaseSimulation, Schema):
             )
 
     #! Generalize from checks for atomic systems, error with CG input
-    def resolve_composition_formula(self, system_parent: ModelSystem) -> None:
+    def resolve_composition_formula(
+        self, system_parent: ModelSystem, logger: 'BoundLogger'
+    ) -> None:
         """Determine and set the composition formula for `system_parent` and all of its
         descendants.
 
@@ -246,13 +248,15 @@ class Simulation(BaseSimulation, Schema):
                 to the atom indices stored in system.
             """
             if not subsystems:
-                atom_indices = (
-                    system.atom_indices if system.atom_indices is not None else []
+                particle_indices = (
+                    system.particle_indices
+                    if system.particle_indices is not None
+                    else []
                 )
                 subsystem_labels = (
-                    [np.array(labels)[atom_indices]]
+                    [np.array(labels)[particle_indices]]
                     if labels
-                    else ['Unknown' for atom in range(len(atom_indices))]
+                    else ['Unknown' for atom in range(len(particle_indices))]
                 )
             else:
                 subsystem_labels = [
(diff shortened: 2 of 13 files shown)


src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -552,7 +552,17 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
 
-class AtomsState(Entity):
+class State(Entity):
+    """
+    A base section to define the state information of the system.
+    """
+
+    def __init__(self, m_def: 'Section' = None, m_context: 'Context' = None, **kwargs):
+        super().__init__(m_def, m_context, **kwargs)
+        self.labels = None
```
JFRudzinski: This is one approach to simplifying the logic in functions like the composition formula ones, but it seems maybe not the best for typing purposes (I can't make it a protected variable because we intend to use it outside of this class). 

Also, the approach doesn't really work for atoms_state vs. particles_state so I am still thinking about that.

src/nomad_simulations/schema_packages/general.py
```diff
@@ -1,10 +1,17 @@
-from typing import TYPE_CHECKING
+#! TODO: Why is TYPE_CHECKING False?
+from typing import TYPE_CHECKING, List, Iterable, Union
 
-if TYPE_CHECKING:
-    from collections.abc import Callable
-
-    from nomad.datamodel.datamodel import EntryArchive
-    from structlog.stdlib import BoundLogger
+if not TYPE_CHECKING:
```
JFRudzinski, line 4: This seems to be a general problem, I need to ask someone

src/nomad_simulations/schema_packages/general.py
```diff
@@ -269,23 +276,28 @@ def get_composition_recurs(system: ModelSystem, atom_labels: list[str]) -> None:
...
-        )
-        atom_labels = (
-            [atom.chemical_symbol for atom in atoms_state]
-            if atoms_state is not None
-            else []
-        )
-        get_composition_recurs(system=system_parent, atom_labels=atom_labels)
+        labels = []
+        if system_parent.cell is not None:
+            if system_parent.cell[0].name == 'AtomicCell':
```
JFRudzinski: This is just a temporary solution to make sure the rest works. 

I think now that we can use some method of an archive section called m_xpath to grab the section by name of the parent class...see get_sibling_section() in utils.py

More generally I am also considering making an intermediate parent for AtomicCell and ParticleCell to share particle-specific quantities and functionalities that we don't want in cell...

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1372,10 +1372,12 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
         #! ChemicalFormula calls `ase_atoms = atomic_cell.to_ase_atoms(logger=logger)` and `ase_atoms.get_chemical_formula()`
         # Creating and normalizing ChemicalFormula section
-        # TODO add support for fractional formulas (possibly add `AtomicCell.concentrations` for each species)
-        sec_chemical_formula = self.m_create(ChemicalFormula)
-        sec_chemical_formula.normalize(archive, logger)
-        if sec_chemical_formula.m_cache:
-            self.elemental_composition = sec_chemical_formula.m_cache.get(
-                'elemental_composition', []
-            )
+        if any(cell.name == 'AtomicCell' for cell in self.cell):
```
JFRudzinski, line 1313: This approach seems fine to me, I just generalized in case later we have some case with both atomic and particle cells (although this seems far away at this point)

JFRudzinski:
We do also need to consider whether a straight copy of the ase Atoms class is a sustainable/robust approach. There are many things to consider. I started investigating, but there is more work to do before I have a clear idea

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -402,6 +411,117 @@ def is_ne_cell(self, other) -> bool:
...
+        if isinstance(self, AtomicCell):
+            return self.atoms_state
+        elif hasattr(self, 'particles_state'):
+            return self.particles_state
+        else:
+            raise AttributeError(
+                'The class does not have atoms_state or particles_state'
+            )
+
+    def get(self, key, logger=None):
```
JFRudzinski, line 424: This is the crux of the implementation of getting the labels from either type of cell. Still not sure if there is a better way

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -552,7 +552,16 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
                 )
 
 
-class AtomsState(Entity):
+class State(Entity):
```
JFRudzinski, line 555: Didn't end up using this after all, but I left it for now in case we need it. But will remove before merge if not

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -476,86 +596,14 @@ def get_chemical_symbols(self, logger: 'BoundLogger') -> list[str]:
...
-        self.positions = positions * ureg('angstrom')
-
     def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
         super().normalize(archive, logger)
 
         # Set the name of the section
         self.name = self.m_def.name if self.name is None else self.name
 
 
+# TODO Consider changing name to BeadCell or CGBeadCell, only using "particle" for the more abstract reference to atoms or beads
```
JFRudzinski, line 616: @Bernadette-Mohr how would you feel about this type of naming?

JFRudzinski:
@Bernadette-Mohr This is mostly working now (in terms of the tests for atomic cell), but I haven't adjusted the parser branch yet. I will try to do that tomorrow.

Just take a look what I did and let me know if you think it makes sense

src/nomad_simulations/schema_packages/model_system.py
Bernadette-Mohr: That makes a lot of sense!
Distinctive, clear naming goes a long way toward well-structured, understandable code.

src/nomad_simulations/schema_packages/model_system.py
Bernadette-Mohr: I think we do need to be able to differentiate at various points, e.g., to selectively skip certain normalizations. I've been trying to come up with a more elegant way than constantly looking for the cell type, but I'm not positive there is one at this point.

src/nomad_simulations/schema_packages/model_system.py
Bernadette-Mohr: I agree. Keeping it as general as possible for as long as possible will make possible extensions later easier, even if we don't have concrete ideas about them now.

Bernadette-Mohr (approved):
Looks good as far as I can see.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -402,6 +411,117 @@ def is_ne_cell(self, other) -> bool:
...
+        if labels is None:
+            logger.error('Could not find `Cell.state.labels`.')
+            labels = ['X'] * len(self.positions) if self.positions is not None else None
+        # Initialize ase.Atoms object with labels
+        # ! We need to make sure that the labels from ase.Atoms are not being used downstream!
+
+        try:
+            symbols2numbers(labels)
+        except KeyError:
+            logger.error('Non chemical symbols in `Cell.state.labels`.')
```
Bernadette-Mohr: Shouldn't this be a `logger.warning`? The `KeyError` is already being handled, and the message seems purely informative at this point.

Bernadette-Mohr (approved):
I missed this one in the first round.

src/nomad_simulations/schema_packages/model_system.py
JFRudzinski: yes! :pray:

JFRudzinski:
covered by #216 and #222

PR #162, opened 2025-01-24
Remove editable annotations

### LINKED ISSUE #161: Remove Str Edit Quantities from schema

ndaelman-hu:
Aside from removing editable quantities, this PR also removes redundant normalizers

closes #161

### DIFF
--- a/docs/normalize.md
+++ b/docs/normalize.md
@@ -25,7 +25,7 @@ class Section1(ArchiveSection):
 class Section2(ArchiveSection):
     normalizer_level = 0
 
-    def normalize(self, achive, logger):
+    def normalize(self, archive, logger):
         super().normalize(archive, logger)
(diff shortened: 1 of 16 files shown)


ndaelman-hu:
@JFRudzinski let me know if I should invite somebody else to review

JFRudzinski (approved):
Looks good, thank you @ndaelman-hu !

I guess the normalize function with super().normalize is a part of the parent classes through inheritance, so they are automatically included, is that right?

ndaelman-hu:
> Looks good, thank you @ndaelman-hu !
> 
> I guess the normalize function with super().normalize is a part of the parent classes through inheritance, so they are automatically included, is that right?

Indeed, that's standard behaviour in OOP, and NOMAD does not modify this. Unless you overwrite it: `self.normalize = super().normalize`

PR #167, opened 2025-02-12
Remove validations

ladinesa:


ladinesa:
@JFRudzinski so I want to retain chemas physical property structure, I like it tbh. I simply removed the unnecessary validations. I tried it on the mapping parser and it seems to work just with the usual m_from_dict.

JFRudzinski:
Let's discuss this at today's meeting. The following questions already come to mind: 
Are you proposing to
- apply this to all properties (even simple properties that don't require additional metadata)?
- also keep the flexible structure and variables functionality?
- treat different types within the same repeating subsection under Outputs?
- treat decomposable types with a contributions subsection?

ndaelman-hu:
> @JFRudzinski so I want to retain chemas physical property structure, I like it tbh. I simply removed the unnecessary validations. I tried it on the mapping parser and it seems to work just with the usual m_from_dict.

@ladinesa If this passes testing, which aspects are blocking this PR? Is just the git pipeline and my feedback?

ndaelman-hu:
@ladinesa You mentioned that this setup worked for your new parser. Is this still the case?
I think that extracting `value` should yield an error for any arrays now with the extra typing restrictions.

ndaelman-hu:
Honestly, I don't see the point in `PhysicalProperty.source`.

@JFRudzinski @ladinesa

JFRudzinski:
> Honestly, I don't see the point in `PhysicalProperty.source`.
> 
> @JFRudzinski @ladinesa

Can you elaborate on your thoughts? I guess the point is that our PP is living under Simulation() anyway, so the context is assumed, and something like `source` was probably intended for the more general PP class, i.e., when a simulation property gets lifted to results where it is placed in an equivalent section/quantity as the corresponding experimental data?

ndaelman-hu:
@ladinesa This is ready for testing now with a parser. Note that you may have to modify the shape of a few cases. `PhysicalProperty.variables` is no longer valid, btw. Where necessary, there will be semantically denoted subsections.

I also enforce shape checking as much as possible. There are 2 instances where the shape may really become more nebulous, maybe I can pin them down in the near future.

ndaelman-hu:
@JFRudzinski Barring any comments from Alvin, I'd say that this is ready for merging.
There is still further cleaning possible, and we should decide on what we do with `variables.py`, but this already fixes most issues + allows Alvin to continue with the parsers.

ladinesa:
> @ladinesa This is ready for testing now with a parser. Note that you may have to modify the shape of a few cases. `PhysicalProperty.variables` is no longer valid, btw. Where necessary, there will be semantically denoted subsections.
> 
> I also enforce shape checking as much as possible. There are 2 instances where the shape may really become more nebulous, maybe I can pin them down in the near future.

i will test this over the weekend.

src/nomad_simulations/schema_packages/outputs.py
```diff
@@ -66,6 +66,8 @@ class Outputs(ArchiveSection):
         """,
     )
 
+
+class ElectronicStructureOutputs(Outputs):
```
ladinesa: Maybe redo this change, i remember we agreed that we, for the moment simply put all properties in Outputs

JFRudzinski: @ladinesa I made this change, please check #175 

I also tested this branch with my h5md parser (both mapping annotations and the old way), and had no issues. I suggest that I merge #175 here and then merge this one. Please confirm.

ladinesa: ok, i am also adapting the changes to the parsers.

JFRudzinski:
@ladinesa Can you have a look at why these tests are failing? Everything works for me locally, and I did not change any dependencies or anything in the toml when I merged my branch. Also, my branch was building correctly on its own....so I have no idea why everything fails now :grimacing:

ladinesa:
> @ladinesa Can you have a look at why these tests are failing? Everything works for me locally, and I did not change any dependencies or anything in the toml when I merged my branch. Also, my branch was building correctly on its own....so I have no idea why everything fails now 😬

i am not sure, but maybe use another nomad-lab version 1.3.14 or latest 1.3.16.dev100

JFRudzinski:
tried:

nomad-lab>=3.14.0
nomad-lab==3.14.0
nomad-lab>=1.3.16.dev100

will post on discord issues

JFRudzinski:
> @ladinesa Can you have a look at why these tests are failing? Everything works for me locally, and I did not change any dependencies or anything in the toml when I merged my branch. Also, my branch was building correctly on its own....so I have no idea why everything fails now 😬

Fix is here on [Discord](https://discordapp.com/channels/1201445470485106719/1356611349685211137)

JFRudzinski (approved):

JFRudzinski:
closes #143

Issue #159, opened 2024-12-13
Splitting `ModelSystem` and `Cell`


ndaelman-hu:
This issue was prompted by the Area C meeting on 13.12.2024

## System Hierarchies and Particle Definitions

The current structure of `ModelSystem` leaves some unclarities open wrt structuring. Any `ModelSystem` instance contains subsections `model_system` and `cell`.

- The intended use of `model_system` is to cover _subsystems_ in the system hierarchy. What may constitute a subsystem is completely context-dependent. Atoms, monomers, material phases, device components, coarse-grained systems, ONIOM shells are equally valid, and may even be used intermittently, e.g. first by material phase, then by monomer: `system -> slab + [polymer -> monomer]` (where `->` stands for "subsystem of"). 
-  `cell` specifies the positions, velocities, etc. of whatever kind of _particle_ it refers to. It may also employ other sections, like `AtomState`, for more contextual definitions. 

## Structuring Options

This schema is now ambiguous regarding where to place the particle definitions (i.e. populate `cell`):

- **top level**: all particles are defined at the top-level system. Subsystems only refer back to the particles via their indices.
- **bottom level**: each leaf node in the system hierarchy defines its own particles. Supersystem nodes (i.e. nodes that aren't leafs) simply stay blank.
- **maximally referential**: similar to "bottom level", but each supersystem node repeats the particle definitions of its own subsystems.
- **minimalistic referential**: similar to "bottom level", but each supersystem node refers to the origins of its own subsystems.
- **free-style**: no clear structure, fully up to the whims of the parser dev.

Under the current schema, all options are equally valid. I think we can all agree that we'd like to set a clearer guideline of which approach to use. We seem to prefer the "top level" approach overall.
This assumes, of course, that we won't run into any cases where we need the more generic "bottom level" and its alternatives. I **think** that we can safely split such cases over multiple entries, as they typically involve very different kinds of calculations. Feel free to provide counterexamples.

## Suggested Solution
 
If we end up going with "top-level", we either:

1. decouple `cell` from `ModelSystem`.
2. have `model_system` be of a simpler type than `ModelSystem`.

Regardless of the implementation, we should still retain some common (intrinsic) properties for any subsystem, e.g. total charge, total mass. We should also support references to `ModelMethod`, to cover cases like ONIOM.

Sketch of the problem statement and suggested solutions.
[...]

PR #170, opened 2025-03-24
Refactoring model system

### LINKED ISSUE #149: Generalize 'model_system'
Could we rename, e.g., `atom_index` or `atomic_system` to `index` and `system,` and isolate calls to specific `ase` functionalities to appropriate cases?
Would this also make sense for electronic-calculations-specific methods, e.g. `Symmetry`?

### LINKED ISSUE #171: revise `AtomsState` to avoid repeated atomic fields
Currently we create an `AtomsState` instance for each atom in the simulation. When there are many identical atoms (e.g. 200 carbons), the fields like `chemical_symbol` and `atomic_number` are repeated in every single `AtomsState`.

**Proposal:** A new concept (e.g., an "AtomDefinition" class or `Element` in basesections.v2) to store _once-per-species data_ (chemical symbol, atomic number, etc.) and reference it from each atom.

I am working on an example implementation in PR #170

EBB2675:
A summary of this PR is given in the following issue: #159. 

The main points are as follows:

- Particle definitions at the **top-level**: all particles are defined at the top-level system. Subsystems only refer back to the particles via their indices.
- Use the basesections.v2.System, and leverage the hierarchical structure via the  “sub_systems.” This would solve the previous child archives of the `ModelSystem` issue,


EBB2675:
- **`AtomsState` Tests**
  - [x] Verify elemental data resolution using `AtomDefinition` (e.g. for Fe, H, Cu, O).

- **`ModelSystem` Basic Tests**
  - [x] Test that `from_ase_atoms()` correctly populates top‑level `positions` and `atom_states`.
  - [x] Test that `to_ase_atoms()` builds an ASE Atoms object using:
    - Chemical symbols from `atom_states` (via `AtomDefinition`)
    - Positions from the `ModelSystem`
    - Lattice vectors and periodic boundary conditions from the first cell.
  - [x] Test roundtrip consistency between `from_ase_atoms()` and `to_ase_atoms()`.
  - [x] Verify that normalization does not update type/dimensionality when:
    - The `ModelSystem` is not representative.
    - No cell is provided.

- **Hierarchical `ModelSystem` Tests**
  - [x] Test a two‑level hierarchy with children having specific `atom_indices`.
  - [x] Test a deep hierarchy with a grandchild to verify:
    - Correct branch labels and branch depths.
    - Preservation of the parent's top‑level `positions` and `atom_states`.

- **Update and Integration Tests**
  - [x] **Re‑normalization / Update Behavior:**  
    Ensure that re‑calling `normalize()` after updating positions and cell (simulating a change from molecule to bulk) updates the computed type/dimensionality.
  - [x] **`Symmetry` & `ChemicalFormula` Integration:**  
    Validate that when a `ModelSystem` has a valid cell (using a dummy subclass for testing), both the `ChemicalFormula` and `Symmetry` sections normalize correctly.
  - [x] **Re‑parenting / Hierarchy Updates:**  
    verify that updating a child’s `atom_indices` is reflected in the parent's hierarchy.
  - [x] **Deep Hierarchy Stress Test:**  
    create a 10‑level deep hierarchy and ensure that recursive hierarchy collection correctly returns the expected branch depths, labels, and `atom_indices`.

JFRudzinski:
From our meeting today (I forget how references are currently working, but the second subsection should be built of refs to the first one):

MS
 |----particle_defns=SubSection(ParticleDefn, repeats=True) [length == n_particle_types]
 |----particle_types=SubSection(Ref(ParticleDefn), repeats=True) [length == n_particles]
 
 ParticleDefns()
     label: str
     charge: float
     ...

ndaelman-hu:
Thank you for all the work. Additional thoughts (from or immediately after our meeting today):

- [x] Pls add a provisional spin class. No need to write an actual implementation, this will likely follow over the course of several PRs.
- [ ] `AtomicCell` can from my side just be usurped into `Symmetry`.
- [ ] Since certain numerical properties are defined in reference to `AtomsState`, it could make sense to have a concept like a "reference state", i.e. one without core-hole excitation, Hubbard onsite interaction 0, etc. This will often be the case for most pseudopotentials. Think on how explicitly you'd like this reference state to be denoted: should an empty state be interpreted as the reference (then the docs should clarify this), or should it normalize to the more verbose representation?
- [ ] While the docs per each class serve their purpose, I feel that this structure has become big enough that we should also add the global overview in our docs. This should just be high-level and focus on where each part is placed. May also be a follow-up PR assigned to somebody else.

EBB2675:
I am a bit confused again:

> MS 
|----particle_defns=SubSection(ParticleDefn, repeats=True) [length == n_particle_types] 
|----particle_types=SubSection(Ref(ParticleDefn), repeats=True) [length == n_particles]

what does the subsection particle_types refer to exactly?

EBB2675:
I understand that there is no particular size restriction on defining a particle, and on a broad context, we wanted to fit `Atoms` under `Particle`. But while implementing it, it occurred to me that we should refrain from that

Atoms have unique properties and behaviors that warrant a dedicated `Atom-something` structure such as `AtomDefn` and `AtomsState` (including how it should be normalized), while non-atomic particles can be described separately. I think the schema is growing deep enough to keep these detached. Also, while keeping things separate, i feel like we can move faster on schema development. I am finishing up the other points we have discussed, and it will be ready for a review very soon @JFRudzinski 

A provisional spin description is something I will add right away, but the other points you have made may be suitable for future PRs indeed :thumbsup: @ndaelman-hu

EBB2675:
I have added a draft spin class at the atom level, which can easily be integrated to the orbital and global levels.

This base is constructed while taking into account multiplicity for molecules, magnetization for materials. Later on: if only collinear data is stored, the vector may effectively have only z-component and others zero by convention.

This is consistent with ASE 's `Atoms` object, and OpenBabel as well.

Currently, an atom's state can look like this:

```
AtomsState:
  atom_definition_ref:
    chemical_symbol: "Fe"
    atomic_number: 26
    charge: 0
  spin_definition_ref:
    spin_type: collinear
    atom_spin_moment: 2.2  # a placeholder in Bohr magnetons
    atom_spin_vector: null  # not used for collinear spin
  orbitals_state:
    - n_quantum_number: 3
      l_quantum_number: 2
      l_quantum_symbol: "d"
      ml_quantum_number: 0
      ml_quantum_symbol: "d_z2"
      j_quantum_number: [2.5]
      mj_quantum_number: [0.5, -0.5]
      ms_quantum_number: 0.5
      ms_quantum_symbol: "up"
      degeneracy: 5
      occupation: 5.0
```

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -358,55 +364,58 @@ class Cell(GeometricSpace):
...
-    )
+    # this is moved to ModelSystem.
+    # n_atoms = Quantity(
+    #     type=np.int32,
+    #     description="""
+    #     Number of atoms in the atomic cell.
+    #     """,
+    # )
 
     equivalent_atoms = Quantity(
```
JFRudzinski, line 351: Do you know in what context this is used? I presume that it has something to do with calculating some minimal cell representation?

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -429,109 +439,109 @@ def __init__(self, m_def: 'Section' = None, m_context: 'Context' = None, **kwarg
...
-        self.lattice_vectors = ase.geometry.complete_cell(cell) * ureg('angstrom')
-
-        # `positions`
-        positions = ase_atoms.get_positions()
-        if (
-            not positions.tolist()
-        ):  # ASE assigns a shape=(0, 3) array if no positions are found
-            return None
-        self.positions = positions * ureg('angstrom')
+    # @staticmethod
```
JFRudzinski: What happened to all these functions? Are they moved, temporarily removed, no longer valid, ...?

It's fine with me if you want to leave some commented out for now to make sure we don't need them later. I would just add some additional comments for context of your intent

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1061,6 +1061,24 @@ class ModelSystem(System):
         """,
     )
 
+    n_atoms = Quantity(
```
JFRudzinski: I would like to make all quantities that live directly in ModelSystem non-atom-specific. Not sure if we even need a `n_atoms`/`n_particles`. But l1052, can we simply change this from `atom_indices` to `indices`? Similar to `positions`. We can try to avoid using "particles" also if you prefer.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1095,7 +1113,142 @@ class ModelSystem(System):
...
+    )
+
+    # New subsection: list of atomic state entries
+    atom_states = SubSection(
+        section_def=AtomsState.m_def,
+        repeats=True,
+        description='List of atomic elements (and their electronic state) for the simulation.',
+    )
+
+    particle_states = SubSection(
```
JFRudzinski, line 979: I might be fine with this, since I guess there will only ever be either atomic or non-atomic...

but why not just have 
class ParticleState():
   ...some generic stuff
class AtomsState(ParticleState):
   ...atom specific stuff
class CoarseGrainedState(ParticleState):
   ...for a CG model

particle_states = SubSection(...ParticleState.m_def)

Then if a new sub-class comes along it's easier to extend. Is there some obvious downside?

Same for the Defs().

I guess from the normalizing standpoint it's a little more difficult because you might have to check the child class type or something to execute some logic. I think I found this when working with the trial implementation of ParticleCell, but I can't exactly remember.

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -633,3 +666,35 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    orbitals_state = SubSection(sub_section=OrbitalsState.m_def, repeats=True)
+
+    core_hole = SubSection(sub_section=CoreHole.m_def, repeats=False)
+
+    hubbard_interactions = SubSection(
+        sub_section=HubbardInteractions.m_def, repeats=False
+    )
+
+
+class ParticleState(Entity):
```
JFRudzinski: I will create a new MR to integrate/migrate @Bernadette-Mohr 's previous work on the particle/CG representation after this one is done.

JFRudzinski (approved):
@EBB2675 Great MR, thanks for all your work!

There are just some relatively minor comments or possible refinements from my side.

If you want to discuss the Atom/Particle implementation, we could do that today < 12:30 or directly after our update meeting.

My suggested approach would be:

-  refine based on people's comments and then just go ahead and merge it, even if there are some smaller open questions. 
- I would continue my assessment by migrating the CG representation stuff next week and at the same time actually testing it with my parser. Alvin can also test it out and give further feedback. 
- After this round of implementation tests, I think we should make a release v0.4.0 to make clear the major change in structure. We could include the preliminary PhysicalProperty changes in the initial release or later depending on the progress (This is mostly for us at this point anyway).

src/nomad_simulations/schema_packages/model_system.py
EBB2675: Thank you very much for pointing it out, I temporarily disabled them. I will add a comment on why

EBB2675:
>     Since certain numerical properties are defined in reference to `AtomsState`, it could make sense to have a concept like a "reference state", i.e. one without core-hole excitation, Hubbard onsite interaction 0, etc. This will often be the case for most pseudopotentials. Think on how explicitly you'd like this reference state to be denoted: should an empty state be interpreted as the reference (then the docs should clarify this), or should it normalize to the more verbose representation?


I have played around with this a bit, and i think it is best if we can deal with the reference state mainly through ModelMethod channel (while referring to the AtomsState), since the structure of core/particle states are strictly bound to the underlying method

i have two options to do this in mind, if you can create an issue and tag me, we can continue this point from there @ndaelman-hu

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -545,9 +547,48 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    # For collinear spin, a single value may suffice.
+    # For non-collinear, you might store a vector.
+    spin_type = Quantity(
+        type=MEnum('collinear', 'non-collinear'),
+        default='collinear',
+        description='Specifies whether the spin is treated as a scalar (collinear) or vector (non-collinear).',
+    )
+
+    # Typical per-atom spin value in collinear calculations (in Bohr magnetons)
+    atom_spin_moment = Quantity(
```
ladinesa: i suggest having a vector value for both spin types.

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -545,9 +547,48 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
     """
-    A base section to define each atom state information.
+    Defines spin properties for a single atomic species.
+    This can include per-atom or per-orbital spin parameters.
+    """
+
+    # For collinear spin, a single value may suffice.
+    # For non-collinear, you might store a vector.
+    spin_type = Quantity(
+        type=MEnum('collinear', 'non-collinear'),
```
ladinesa: do we expect other spin types in the future? if not could this be also a bool quantity, say, is_non_collinear?

src/nomad_simulations/schema_packages/model_system.py
EBB2675: I am on it

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -545,9 +547,48 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        shape=['3'],
+        unit='bohr_magneton',
+        description='3D spin vector per atom for non-collinear calculations.',
+    )
+
+
+class AtomDefn(Entity):
+    """
+    A base section that defines a single atomic species or element, to be referenced
+    by `AtomsState` or other sections. It stores the essential properties of an
```
ladinesa: update AtomsState?

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -545,9 +547,48 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        type=np.float64,
+        shape=['3'],
+        unit='bohr_magneton',
+        description='3D spin vector per atom for non-collinear calculations.',
+    )
+
+
+class AtomDefn(Entity):
+    """
+    A base section that defines a single atomic species or element, to be referenced
```
ladinesa: Just in general, maybe simply start with Defines

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -633,3 +666,35 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+
+class AtomsState(Entity):
+    """
+    A base section to define each atom state information.
+    """
+
+    atom_definition_ref = Quantity(
+        type=AtomDefn,
+        description="""
+        A reference to the `ParticleDefn` section that describes this atom's species. This
```
ladinesa: is this ParticleDefn?

src/nomad_simulations/schema_packages/general.py
```diff
@@ -249,20 +249,26 @@ def get_composition_recurs(system: ModelSystem, atom_labels: list[str]) -> None:
...
                 system=system, subsystems=subsystems, atom_labels=atom_labels
             )
             if subsystems:
                 for subsystem in subsystems:
                     get_composition_recurs(system=subsystem, atom_labels=atom_labels)
 
-        atoms_state = (
-            system_parent.cell[0].atoms_state if system_parent.cell is not None else []
-        )
+        # atoms_state = (
```
ladinesa: why are these commented out?

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -653,47 +663,34 @@ def resolve_analyzed_atomic_cell(
...
+        Note: In the new design, the cell contains only geometric data.
+        The per-atom electronic/state information is stored at the ModelSystem level (atom_states).
         """
-        if cell_type not in ['primitive', 'conventional']:
-            logger.error(
-                "Cell type not recognized, only 'primitive' and 'conventional' are allowed."
-            )
+        try:
+            wyckoff = symmetry_analyzer.get_wyckoff_letters(cell_type)
+            equivalent_atoms = getattr(
```
ladinesa: this is not very pythonic, i suggest having a map of cell type and get system function

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1095,7 +1113,142 @@ class ModelSystem(System):
         """,
     )
 
-    model_system = SubSection(sub_section=SectionProxy('ModelSystem'), repeats=True)
+    total_charge = Quantity(
```
ladinesa, line 962: in general, is the idea having `total_charge` quantity or `charge` section to also include contributions? or are these atomic quantites defined in atom_states?

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1095,7 +1113,142 @@ class ModelSystem(System):
...
+    total_spin = Quantity(
+        type=np.int32,
+        description="""
+        Total spin state of the system.
+        Not to be confused with the spin multiplicity 2S + 1.
+        """,
+    )
+
+    # New subsection: list of unique atom definitions
+    atom_types = SubSection(
```
ladinesa: i am not sure if the benefit that we get from introducing this list of unique atom types outweighs the additional complexity

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1150,46 +1303,94 @@ def resolve_system_type_and_dimensionality(
...
+    def populate_atom_types(self, logger: 'BoundLogger') -> None:
+        """
+        Populates the `atom_types` subsection with unique AtomDefn entries
+        extracted from the `atom_states` subsection.
+        """
+        unique_types = {}
+        for state in self.atom_states:
+            atom_def = state.atom_definition_ref
+            if atom_def is None:
+                logger.warning('Missing atom_definition_ref in one of the atom_states.')
```
ladinesa: this could result in a lot of warning, i suggest, putting the warning outside the loop if warning is found

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1150,46 +1303,94 @@ def resolve_system_type_and_dimensionality(
...
         return system_type, dimensionality
 
+    def populate_atom_types(self, logger: 'BoundLogger') -> None:
+        """
+        Populates the `atom_types` subsection with unique AtomDefn entries
+        extracted from the `atom_states` subsection.
+        """
+        unique_types = {}
+        for state in self.atom_states:
+            atom_def = state.atom_definition_ref
```
ladinesa: i am not sure i follow, you want to populate atom_types but you read them from state.atom_definition_ref but this also references the one in atom_types?

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1095,7 +1113,142 @@ class ModelSystem(System):
...
+            self.atom_states.append(state)
+
+        positions = ase_atoms.get_positions()
+        if not positions.tolist():
+            logger.error('ASE Atoms has no positions.')
+            return
+        self.positions = positions * ureg('angstrom')
+        self.n_atoms = len(self.positions)
+
+        # Optionally, update cell information from ASE
```
ladinesa: should this not be mandatory and read from ase.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1095,7 +1113,142 @@ class ModelSystem(System):
...
+        ase_atoms.set_positions(self.positions.to('angstrom').magnitude)
+        return ase_atoms
+
+    def from_ase_atoms(self, ase_atoms: ase.Atoms, logger: 'BoundLogger') -> None:
+        """
+        Populates ModelSystem from an ASE Atoms object.
+        Replaces the atom_states subsection with new entries based on the ASE chemical symbols,
+        using AtomDefinition to store elemental data, and assigns ASE positions to the top-level positions quantity.
+        """
+        self.atom_states.clear()
```
ladinesa: is it only atoms_states that you update, also cell info.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1095,7 +1113,142 @@ class ModelSystem(System):
...
+                )
+            else:
+                logger.warning('No lattice_vectors found in cell[0].')
+        else:
+            logger.warning('No cell section available in ModelSystem.')
+
+        if self.positions is None:
+            logger.error('ModelSystem.positions is not available.')
+            return None
+        if len(self.positions) != len(symbols):
```
ladinesa: i am not sure if this possible as there are internal consistency checks for model system

src/nomad_simulations/schema_packages/atoms_state.py
EBB2675: I dont. How do we benefit from using boolean instead of MEnum?

src/nomad_simulations/schema_packages/atoms_state.py
EBB2675: I will add ParticleDefn as Joe suggested:

> 
> class ParticleState():
> ...some generic stuff
> class AtomsState(ParticleState):
> ...atom specific stuff
> class CoarseGrainedState(ParticleState):
> ...for a CG model

src/nomad_simulations/schema_packages/general.py
EBB2675: atoms_state is not under cell anymore

src/nomad_simulations/schema_packages/atoms_state.py
ladinesa: if you use it later u dont have to do comparisons like if spin_type == collinear just makes it a bit simpler. in general you use MEnum if you expect more than two options

EBB2675: thank you very much, i will switch to boolean

ladinesa: only if you dont expect other spin types later.

src/nomad_simulations/schema_packages/model_system.py
EBB2675: i have moved these comparison functions from Cell to ModelSystem

src/nomad_simulations/schema_packages/model_system.py
EBB2675: later on, when we add the atomic partial charge schemes (through PhysicalProperty), it should be tied to the total_charge

src/nomad_simulations/schema_packages/model_system.py
EBB2675: i am removing it for now

src/nomad_simulations/schema_packages/model_system.py
EBB2675: @ndaelman-hu

src/nomad_simulations/schema_packages/model_system.py
EBB2675: done

src/nomad_simulations/schema_packages/model_system.py
EBB2675: removed

EBB2675:
FYI: I am done with your first round of feedback @JFRudzinski @ladinesa @ndaelman-hu

ladinesa:
> FYI: I am done with your first round of feedback @JFRudzinski @ladinesa @ndaelman-hu

there is one more unresolved comment but can also simply resolve it.

ladinesa (approved):

EBB2675:
> there is one more unresolved comment but can also simply resolve it.

done. Should I also modify the failing tests as well, or should it be left for another PR? @JFRudzinski @ladinesa

ladinesa:
> > there is one more unresolved comment but can also simply resolve it.
> 
> done. Should I also modify the failing tests as well, or should it be left for another PR? @JFRudzinski @ladinesa

Please fix them before merging. thanks

JFRudzinski:
@EBB2675   In addition to fixing the tests I would like to do one last review before we merge. I will try to get it done this afternoon :crossed_fingers:

EBB2675:
> @EBB2675 In addition to fixing the tests I would like to do one last review before we merge. I will try to get it done this afternoon 🤞

alright, i am updating the tests this afternoon, feel free to merge anytime after your review

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -4,7 +4,9 @@
 import numpy as np
 import pint
 from nomad.datamodel.data import ArchiveSection
-from nomad.datamodel.metainfo.basesections import Entity
+
+# from nomad.datamodel.metainfo.basesections import Entity
```
JFRudzinski: I think you can delete this

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -4,7 +4,9 @@
...
+from nomad.datamodel.metainfo.basesections.v2 import Element, Entity
```
JFRudzinski: remove Element if not used

src/nomad_simulations/schema_packages/general.py
```diff
@@ -249,21 +249,17 @@ def get_composition_recurs(system: ModelSystem, atom_labels: list[str]) -> None:
...
                 system=system, subsystems=subsystems, atom_labels=atom_labels
             )
             if subsystems:
                 for subsystem in subsystems:
                     get_composition_recurs(system=subsystem, atom_labels=atom_labels)
 
-        atoms_state = (
-            system_parent.cell[0].atoms_state if system_parent.cell is not None else []
-        )
         atom_labels = (
```
JFRudzinski, line 262: I think there might be 2 issues here:

1. I think there is a misunderstanding of what composition formula is supposed to be. What is implemented here is to get the "normal" chemical formula at each level. However, the composition formula is supposed to be a sort of generalized formula of each parent with respect to its children.
2. This code (which is the appropriate one for the lowest level of the system tree) is only valid for atoms not particles in general.

We don't need to address either of these in this MR, but let's make a couple TODOs and maybe open a or 2 corresponding issues.

JFRudzinski: For the second point, let's also make a note that we should generalize the naming from atom to particle.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -36,7 +36,9 @@
 from nomad.atomutils import Formula, get_normalized_wyckoff, search_aflow_prototype
 from nomad.config import config
 from nomad.datamodel.data import ArchiveSection
-from nomad.datamodel.metainfo.basesections import Entity, System
+
+# from nomad.datamodel.metainfo.basesections import Entity, System
```
JFRudzinski: remove

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -193,7 +199,7 @@ def get_geometric_space_for_atomic_cell(self, logger: 'BoundLogger') -> None:
         Args:
             logger (BoundLogger): The logger to log messages.
         """
-        atoms = self.to_ase_atoms(logger=logger)  # function defined in AtomicCell
+        atoms = self.to_ase_atoms(logger=logger)  # FIX: not in AtomicCell anymore
```
JFRudzinski: switch to "TODO" for searcheability, and be a little more verbose about what needs to be fixed so that it would be at least somewhat clear for someone else

src/nomad_simulations/schema_packages/model_system.py
JFRudzinski: Just add a "#?" comment that asks the use of this

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -358,55 +345,16 @@ class Cell(GeometricSpace):
...
-        description="""
-        Number of atoms in the atomic cell.
-        """,
-    )
-
     equivalent_atoms = Quantity(
         type=np.int32,
-        shape=['n_atoms'],
+        # shape=['n_atoms'],
+        shape=['*'],  # temporarily
```
JFRudzinski: Why? Please add "#TODO" "#?" or "#!" comment with more context

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -418,121 +366,13 @@ class AtomicCell(Cell):
     # ! improve description and clarify whether this belongs to `Symmetry` with @lauri-codes
     wyckoff_letters = Quantity(
         type=str,
-        shape=['n_atoms'],
+        # shape=['n_atoms'],
+        shape=['*'],  # temporarily
```
JFRudzinski: same here

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -667,33 +507,45 @@ def resolve_analyzed_atomic_cell(
...
+        # Create the cell (or atomic cell) for geometry only
+        atomic_cell = Cell(type=cell_type)
         atomic_cell.lattice_vectors = cell * ureg.angstrom
-        for label, atomic_number in zip(labels, atomic_numbers):
-            atom_state = AtomsState(chemical_symbol=label, atomic_number=atomic_number)
-            atomic_cell.m_add_sub_section(AtomicCell.atoms_state, atom_state)
-        atomic_cell.positions = (
-            positions * ureg.angstrom
-        )  # ? why do we need to pass units
+        # Note: positions are no longer stored in the cell in the new design.
```
JFRudzinski: I like the the extra comments in general, some suggestions: 
1. use a "#!" for more visibility
2. clarify where the positions are "# ! Positions are now stored directly under `model_system`"
3. clarify what new design means (you can use the new release planned): "# ! Positions are stored directly under `model_system` in nomad-simulations>=0.4"

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -914,6 +766,10 @@ class ModelSystem(System):
     """
     Model system used as an input for simulating the material.
 
+    This version stores the atomic positions at the top level in the quantity “positions”
```
JFRudzinski: I don't think you need to say "this version", you can just be more direct "Atom positions are stored at the top...", etc.

JFRudzinski (commented):
@EBB2675 Great work! I have just added some minor comments, mostly dealing with adding a few additional comments or possibly opening issues to log some of the TODOs.

I didn't review the tests. Let me know tomorrow once you are done with your adjustments and I will do a quick review :pray:

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: Yes, in solid state this would amount to the Wyckoff positions.
It think this should go under the cell symmetry.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -358,55 +343,18 @@ class Cell(GeometricSpace):
...
-        Number of atoms in the atomic cell.
-        """,
-    )
-
     equivalent_atoms = Quantity(
         type=np.int32,
-        shape=['n_atoms'],
+        # shape=['n_atoms'],
+        # TODO : the shape check is removed due to removing n_atoms from AtomicCell.
+        # @ndaelman: Please adjust according to AtomicCell's needs
```
ndaelman-hu: This is fine. I would see to enforce shape via class decorators anyhow.
You can remove this comment, it shouldn't go into production code.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -924,19 +782,18 @@ class ModelSystem(System):
...
     It is composed of the sub-sections:
-        - `AtomicCell` containing the information of the atomic structure,
         - `Symmetry` containing the information of the (conventional) atomic cell symmetry
         in bulk ModelSystem,
         - `ChemicalFormula` containing the information of the chemical formulas in different
         formats.
 
-    This class nest over itself (with the section proxy in `model_system`) to define different
-    parent-child system trees. The quantities `branch_label`, `branch_depth`, `atom_indices`,
+    This class nest over itself (with the section proxy in `sub_systems`) to define different
```
ndaelman-hu: typo: nest**s**

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1095,7 +969,126 @@ class ModelSystem(System):
...
+        """,
+    )
+
+    total_spin = Quantity(
+        type=np.int32,
+        description="""
+        Total spin state of the system.
+        Not to be confused with the spin multiplicity 2S + 1.
+        """,
+    )
```
ndaelman-hu: Good disclaimer. Just be sure, you mean the eigenvalue of the $S^2$ operator, correct?

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1157,39 +1150,50 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        return set(self._generate_comparer()) <= set(other._generate_comparer())
+
+    def is_gt_structure(self, other: 'ModelSystem') -> bool:
+        return set(self._generate_comparer()) > set(other._generate_comparer())
+
+    def is_ge_structure(self, other: 'ModelSystem') -> bool:
+        return set(self._generate_comparer()) >= set(other._generate_comparer())
+
+    def is_ne_structure(self, other: 'ModelSystem') -> bool:
+        return not self.is_equal_structure(other)
```
ndaelman-hu, line 1183: The reallocation to `ModelSystem` is appropriate.

Regarding changes to the interface, there are some subtle implications that you should consider. I'd also appreciate @kubanmar 's take here.

Python comes with a [predefined interface for comparison](https://docs.python.org/3/reference/datamodel.html#object.__lt__), which also powers symbols like `<`, etc.
Note that `other` in the argument specification, `object.__lt__(self, other)`, does not have to implement the comparison operator too. It's just the 2nd argument.
This behavior is comes from "There are no swapped-argument versions of these methods (to be used when the left argument does not support the operation but the right argument does);".
That is to say, `<` will first try `self.__lt__(other)`. If that raises `NotImplementedError` (i.e. `self` is not an object with a comparer class), the next action is to test `other.__gt__(self)`.
This also implies that `self` and `other` can come from different classes, each with its own comparer.

Note that your modification to `_generate_comparer` and the overall interface, deviates from the default Python requirements.

Now, there are 2 arguments to be made against the Python default (in our case).
At large, it works best for total orders, not partial orders, such as what we are finding here. Still, it also accommodates the latter, so hardly a strong argument.
Secondly, we are using our own functions, which do not come with a reflective catch system.
If we settle on a single way of comparing Systems(or previously Cells), then we could go with the default. Otherwise, its better to add multiple specialized comparisons in System, or extend them / the default via inheritance. 

Note that overwriting the default comparisons may break with the NOMAD implementation, though I think that the hash function resolves this issue.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1095,7 +969,126 @@ class ModelSystem(System):
...
+        for particle_state in self.particle_states:
+            if isinstance(particle_state, AtomsState):
+                # only collect symbols from particle entries that are instances of AtomsState
+                if (
+                    not particle_state.atom_definition_ref
+                    or not particle_state.atom_definition_ref.chemical_symbol
+                ):
+                    logger.warning(
+                        'Could not find `chemical_symbol` in an AtomsState entry.'
+                    )
```
ndaelman-hu: In line with the "Dependency Inversion Principle" (SOLI**D**), I'd put the burden on an `AtomsState` superclass to define the interface of providing the representation.
This is then can called for all particle states in a system branch.

```suggestion python
try:
    symbols.append(particle_state.fetch_symbol())
except AttributeError:
    logger.error('Particle cannot provide label information.')
    raise NotImplementedError
```

Note that I modified the name from `chemical_symbols` to `symbols`.
In your discussions with Areas A and C2, what system did you settle on for handling anything that goes outside of the atomic representation?

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -907,13 +761,17 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
-            self.m_cache['elemental_composition'] = formula.elemental_composition()
+            self.m_cache['elemental_composition'] = formula.elemental_composition
 
 
 class ModelSystem(System):
     """
     Model system used as an input for simulating the material.
 
+    Atom positions at the top level in the quantity “positions”
+    and the per‐atom state (including electronic state information) in a new subsection “particle_states”.
```
ndaelman-hu: you can remove the word "new" here.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -907,13 +761,17 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    Downstream subsystems refer to atoms via particle_indices.
```
ndaelman-hu, line 763: Please clarify the distinction between "atoms" and "particles", as the terms are now used interchangeably.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -667,33 +507,45 @@ def resolve_analyzed_atomic_cell(
...
-            positions * ureg.angstrom
-        )  # ? why do we need to pass units
+        # ! Positions are stored directly under model_system in nomad-simulations>=0.4.
         atomic_cell.wyckoff_letters = wyckoff
         atomic_cell.equivalent_atoms = equivalent_atoms
-        atomic_cell.get_geometric_space_for_atomic_cell(logger=logger)
+        try:
+            atomic_cell.get_geometric_space_for_atomic_cell(logger=logger)
+        except Exception as e:
+            logger.warning('Could not extract geometric space from cell', exc_info=e)
```
ndaelman-hu: Since `get_geometric_space_for_atomic_cell` accepts a `logger`, please move this message in there. Alternatively, remove the `logger` parameter.

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -545,9 +545,42 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
     """
-    A base section to define each atom state information.
+    Defines spin properties for a single atomic species.
+    TODO @ EBB : This definition will be enriched in a seperate PR.
+    """
+
+    is_collinear = Quantity(
+        type=bool,
+        default=True,
+        description='Specifies whether the spin is treated as a scalar (collinear) or vector (non-collinear).',
```
ndaelman-hu: Given your description, `is_scalar`, or `modeled_as_scalar` seems more appropriate.
You see, vector or spinors may still be collinear. That more so has to do with their local reference systems.

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -545,9 +545,42 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    atom_spin_vector = Quantity(
+        type=np.float64,
+        shape=['3'],
+        unit='bohr_magneton',
+        description="""
+        3D spin vector per atom.
+        For collinear calculations, only the z-component should be non-zero.
+        For non-collinear calculations, all components may be used.
+        """,
+    )
```
ndaelman-hu: So, in your approach, the user still ends up working with a vector.
In that case, you should provide functionalities that enforce this format.

Beter yet, leverages an abstract class, where different implementations exist, e.g. `ScalarSpin`, `VectorSpin`, etc.

I would keep `SpinDefn` abstract (like the name), defining only where the actual "spin" (be it scalar, or something else) is defined.

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -545,9 +545,42 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+
+class AtomDefn(Entity):
+    """
+    Defines a single atomic species or element, to be referenced
+    by `AtomsState` or other sections. It stores the essential properties of an
+    element (symbol, atomic number and charge), ensures consistency between them, and can
+    resolve one from the other if only one is provided.
+
+    Typical usage:
+      - Create one `AtomDefn` for each unique element (or pseudo-atom).
```
ndaelman-hu: what is a "pseudo-atom"?

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -633,3 +658,44 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
             self.chemical_symbol = self.resolve_chemical_symbol(logger=logger)
         if self.atomic_number is None:
             self.atomic_number = self.resolve_atomic_number(logger=logger)
+
+
+class ParticleState(Entity):
```
ndaelman-hu: I would add a "symbol" or "representation" attribute here. These would then coincide with the "chemical_symbol" in `AtomsState`.
The latter has the advantage of indicating which is the main way to define a specific instance of them.

#### My reasoning

In NOMAD we use sections (implemented via classes). A section is effectively a self-enforcing format, i.e. defined via its attributes and enforced via its functions.
`ParticleState` has no code indicating its semantics, only natural language. While there are use-cases for this (especially when treating sections as types), I would always seek to concretize a section.

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -545,9 +545,42 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    Defines a single atomic species or element, to be referenced
+    by `AtomsState` or other sections. It stores the essential properties of an
+    element (symbol, atomic number and charge), ensures consistency between them, and can
+    resolve one from the other if only one is provided.
+
+    Typical usage:
+      - Create one `AtomDefn` for each unique element (or pseudo-atom).
+      - In simulation data, multiple `AtomsState` sections can then reference the
+        same `AtomDefn`, thus avoiding repeated `chemical_symbol` or
+        `atomic_number` in every single atom entry.
```
ndaelman-hu: Would re-word this. Effectively, `AtomDefn` defines a simulated species (or element, as you described above). Most importantly, these properties do not change during the calculation!!

In `AtomsState.atom_definition_ref`, you can explain that different atoms states may share the same definition. I would do it there, as this is the way `AtomsState` uses this section, not some kind of inherent property.

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -633,3 +658,44 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        """,
+    )
+
+    spin_definition_ref = Quantity(
+        type=SpinDefn, description='Reference to the spin definition for this atom.'
+    )
+
+    orbitals_state = SubSection(sub_section=OrbitalsState.m_def, repeats=True)
+
+    core_hole = SubSection(sub_section=CoreHole.m_def, repeats=False)
```
ndaelman-hu: Maybe this can become a description of the core at large in the future?
Though, some of that also falls under `AtomsDefn`.

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -483,60 +483,81 @@ def resolve_orbital_references(
...
+                active_atom_state = model_system.particle_states[idx]
+
+                # If no orbitals_state => skip
+                if not active_atom_state.orbitals_state:
+                    logger.warning(
+                        f'No orbitals_state found in particle_states[{idx}].'
+                    )
+                    continue
+
+                # Append them
```
ndaelman-hu: A bit of a superfluous comment, no?
LLMs tend to generate them, but they can be removed from the production code.

ndaelman-hu: Btw, this applies to some of the other comments here too.
Use your own judgement where you find them useful, but commenting a single statement is a bit overkill.

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -483,60 +483,81 @@ def resolve_orbital_references(
...
             model_system = model_systems[model_index]
         except IndexError:
-            logger.warning(
-                f'The `ModelSystem` section with index {model_index} was not found.'
-            )
+            logger.warning(f'No ModelSystem at index {model_index}.')
             return None
 
-        # If `ModelSystem` is not representative, the normalization will not run
+        # 2) If the system is not representative, skip
```
ndaelman-hu: This isn't skipping, as done below...

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -483,60 +483,81 @@ def resolve_orbital_references(
...
         if is_not_representative(model_system=model_system, logger=logger):
             return None
 
-        # If `AtomicCell` is not found, the normalization will not run
-        if not model_system.cell:
-            logger.warning('`AtomicCell` section was not found.')
+        # # 3) NOT NEEDED ANYMORE: Check for at least one cell if that is your design’s requirement
+        # if not model_system.cell:
+        #     logger.warning("No AtomicCell found, skipping orbital references resolution.")
+        #     return None
```
ndaelman-hu: If it isn't needed anymore, pls delete this comment.

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -483,60 +483,81 @@ def resolve_orbital_references(
...
 
         Args:
             model_systems (list[ModelSystem]): The list of `ModelSystem` sections.
             logger (BoundLogger): The logger to log messages.
-            model_index (int, optional): The `ModelSystem` section index from which resolve the references. Defaults to -1.
+            model_index (int, optional): The ModelSystem index to use. Defaults to -1 (the last).
 
         Returns:
-            Optional[list[OrbitalsState]]: The resolved references to the `OrbitalsState` sections.
+            Optional[list[OrbitalsState]]: The resolved references to the OrbitalsState sections.
```
ndaelman-hu, line 499: Pls explain the return semantics, i.e. when should this function return `None` and when `[]`?
It seems that the distinction lies in the kind of exception it encounters.

Instead, think of it from the perspective of its usage: what kind of information is communicated to the code calling this function?

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -483,60 +483,81 @@ def resolve_orbital_references(
...
+                    logger.warning(
+                        f'No orbitals_state found in particle_states[{idx}].'
+                    )
+                    continue
+
+                # Append them
+                orbitals_ref.extend(active_atom_state.orbitals_state)
+
+        # 7) Return the collected orbitals
+        return orbitals_ref if orbitals_ref else []
```
ndaelman-hu:
```suggestion
        return orbitals_ref
```

src/nomad_simulations/schema_packages/properties/spectral_profile.py
```diff
@@ -422,37 +444,44 @@ def generate_from_projected_dos(
...
+        # Distinguish orbital and atom projected DOS
+        orbital_projected = self.extract_projected_dos('orbital', logger)
+        atom_projected = self.extract_projected_dos('atom', logger)
         if len(atom_projected) == 0:
+            # Group orbital PDOS by their parent AtomsState (from ModelSystem.particle_states)
             atom_data: dict[AtomsState, list[DOSProfile]] = {}
             for orb_pdos in orbital_projected:
-                # `entity_ref` is the `OrbitalsState` section, whose parent is `AtomsState`
-                entity_ref = orb_pdos.entity_ref.m_parent
+                # entity_ref is expected to be an OrbitalsState; get its parent AtomsState
```
ndaelman-hu: "parent" is a term that applies to "inheritance".
That's not the relationship with `AtomsState`, which is more so a "containing section"?

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -545,9 +545,42 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
                 )
 
 
-class AtomsState(Entity):
+class SpinDefn(Entity):
     """
-    A base section to define each atom state information.
+    Defines spin properties for a single atomic species.
+    TODO @ EBB : This definition will be enriched in a seperate PR.
+    """
```
ndaelman-hu: I'm confused: does this section "define" the spin in the same way the atoms state is defined (see my comment there)?

Or does it rather contain the representation of the spin, at the start of the calculation? I presume so, but collinearity is defined at a system-wide scale, not individual species. If so, I would call this section `SpinState`, and reserve `Defn` for immutable (under computation) characteristics.

P.s.: where (in `properties`) is the computed spin stored? Do we want to use the same definition?

src/nomad_simulations/schema_packages/properties/spectral_profile.py
```diff
@@ -101,12 +101,45 @@ def resolve_pdos_name(self, logger: 'BoundLogger') -> Optional[str]:
...
+            # Fallback: if the OrbitalsState itself has an atom_definition_ref (if defined), use it.
+            if parent_symbol is None and hasattr(
+                self.entity_ref, 'atom_definition_ref'
+            ):
+                parent_symbol = self.entity_ref.atom_definition_ref.chemical_symbol
+            if parent_symbol is None:
+                logger.warning(
+                    'Could not resolve chemical symbol for OrbitalsState reference.'
+                )
+                return None
```
ndaelman-hu: 1. You can split this more efficiently (from a performance pov) into an indented if-else block.
2. Note that deserialization may also fail and return a `str` instead. @TLCFEM is this anticipated behavior in that case?

src/nomad_simulations/schema_packages/properties/spectral_profile.py
```diff
@@ -101,12 +101,45 @@ def resolve_pdos_name(self, logger: 'BoundLogger') -> Optional[str]:
             )
             return None
 
-        # Resolve the `name` from the `entity_ref`
         name = None
+        # If the reference is to an AtomsState, use its atom_definition_ref
```
ndaelman-hu: While I'm fine with the logic at hand, it's kinda clumsy having all this logic in one place.
I can moreover see other sections using similar functionalities.
I think that this could be better handled with (a few) abstraction layer(s) that define the necessary behavior obtain representations.

ndaelman-hu (commented):
I only checked the code, not the tests.

There are some places where the new semantics becomes too vague. I also think that some abstraction and convention is needed in  a few places. I've marked the cases that I could find.

src/nomad_simulations/schema_packages/model_system.py
EBB2675: done

src/nomad_simulations/schema_packages/model_system.py
EBB2675: the eigenval of the S2 operator would be S(S+1)
here, the intention is the store the total spin quantum number, here is a reminder:


| Symbol | Meaning | Typical value | Stored as |
|--------|---------|---------------|-----------|
| **S** | Total spin quantum number | 0, ½, 1, 3⁄2 … | integer or half-integer |
| **2 S + 1** | Spin multiplicity | 1, 2, 3… | integer |
| **S (S + 1)** | Eigenvalue of **Ŝ²** (in units of ℏ²) | 0, ¾, 2… | real number |


I will update the docstring as follows, so that it is clear, thank you very much for the comment.

```
Total spin quantum number **S** of the system (so Ŝ² ψ = S(S+1) ħ² ψ).
Stored as an integer or half-integer represented in doubled form
(e.g. singlet → 0, doublet → 1, triplet → 2).
Not to be confused with the spin multiplicity 2S+1.
```

src/nomad_simulations/schema_packages/properties/spectral_profile.py
EBB2675: Good catch — I was sloppy with the wording. I will update it

src/nomad_simulations/schema_packages/atoms_state.py
EBB2675: Thanks for catching this. The section I called **`SpinDefn`** doesn’t define an immutable property the way, say, an element’s atomic number is defined. Instead, it records the **starting spin settings** we pass to the code for each species (initial magnetisation and similar knobs). Those values may change as the SCF cycle converges, so what we really have here is a *state*, not a definition.

With that in mind I’ll rename the section to **`SpinState`**, aligning it with our existing `AtomsState` and `OrbitalsState` sections. The global choice between collinear and non-collinear calculations belongs on the `System` object, so I’ll move that flag there. The physical property component of the Spin, we can discuss in another PR/thread.

src/nomad_simulations/schema_packages/general.py
JFRudzinski: I created an issue so I can check this later. This can be resolved.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -418,121 +362,13 @@ class AtomicCell(Cell):
     # ! improve description and clarify whether this belongs to `Symmetry` with @lauri-codes
     wyckoff_letters = Quantity(
         type=str,
-        shape=['n_atoms'],
+        # shape=['n_atoms'],
```
JFRudzinski: This can be deleted probably

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -418,121 +362,13 @@ class AtomicCell(Cell):
...
+        shape=['*'],  # TODO
```
JFRudzinski: Add TODO specifics

src/nomad_simulations/schema_packages/model_system.py
JFRudzinski: We probably need to go through later to make sure that we don't say atoms somewhere instead of particles. I think this could be it's own task. I create an issue for this. Can be resolved here.

src/nomad_simulations/schema_packages/model_system.py
JFRudzinski: @ndaelman-hu if there is something to discuss further here, let's open an issue about it and then resolve this comment so that we can push this MR through. I think it is beyond the scope here.

src/nomad_simulations/schema_packages/model_method.py
JFRudzinski: I agree with this comment, but I also think you can leave superfluous comments for now, if you are not sure whether they are helpful. I made an issue to comb through the code comments to clean them up. I can do this later.

However, if you have old pieces of code that you think is not necessary, let's verify and remove them before the merge

JFRudzinski (approved):
@EBB2675 This is nearly ready! I just made a couple new small comments.

There are also some open threads from Nathan's older review. Some are tiny things, others I have addressed by opening issues, so you can resolve those. 

Make sure all the threads are resolved one way or another and then go ahead and merge please :+1:

src/nomad_simulations/schema_packages/atoms_state.py
EBB2675: a dummy atom, or a single point charge

src/nomad_simulations/schema_packages/model_system.py
EBB2675: `chemical_symbol` is the element abbreviation and not semantically interchangeable with a generic `symbol` field.
I would rather keep all the logic about "what makes my symbol separate" at this stage. Maybe after `ParticleStates` is finalized.

JFRudzinski: Something like this if there are 2 opinions at the moment, just open an issue with the relevant info (link to this comment) and tag with milestone 1.0 and then you can resolve this without waiting for Nathan's response.

src/nomad_simulations/schema_packages/atoms_state.py
EBB2675: Spin will be tackled in another PR

src/nomad_simulations/schema_packages/atoms_state.py
EBB2675: this has not been yet implemented.

src/nomad_simulations/schema_packages/model_method.py
EBB2675: TB is not in the scope of this MR, please create an issue

src/nomad_simulations/schema_packages/model_system.py
EBB2675: I’ve moved the try/except (and its warning) into `get_geometric_space_for_atomic_cell` itself and removed the outer catch in `normalize()`. Now all logging of ASE‐related errors lives in the helper method that takes the logger.

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: did an issue get opened, or should I do it?

src/nomad_simulations/schema_packages/model_method.py
ndaelman-hu: I didn't understand your reply. What does TB stand for, tight-binding?
Anyhow, I'm fine adding this as a separate mini-issue.

EBB2675: Context is already in the code. The lines you commented on sit squarely in the tight‑binding helper. Please review what’s there before raising questions that the code itself answers

We already opened up an issue for improving the docstrings. I think this falls under that category.

If you have a legitimate tight‑binding concern, open a separate issue and tag me; otherwise, keep the reviews laser‑focused so we can ship faster from now on

ndaelman-hu: A simple "We already opened up an issue for improving the docstrings. I think this falls under that category." would've sufficed.

PR #172, opened 2025-04-03
Update Python version in CI/CD

ndaelman-hu:
The current `nomad-lab` version no longer supports Python 3.9, but Python 3.12 instead.
Bumping this is necessary for cross-compatiblity of dependencies like `pydantic`.


ladinesa (approved):

blueraft (approved):

JFRudzinski (approved):
:+1:

PR #176, opened 2025-04-09
Allow l=0 in OrbitalsState validation

### LINKED ISSUE #174: Allow l=0 in `OrbitalsState` validation
Currently, `OrbitalsState.validate_quantum_numbers` disallows l=0 because it enforces l > 0. 

But physically, the s orbital has l=0. 

> Expected: True for l=0, n=1, but got False with an error: "The l_quantum_number must be greater than 0."

EBB2675:
small fix for the validation of `l_quantum_number`


JFRudzinski (approved):
I am also getting these build error in #167 , these came suddenly without any reason from changes, so I think maybe something changed with nomad-lab. I already asked Alvin about it.

PR #186, opened 2025-05-09
documentation

EBB2675:


docs/model_system/model_system.md
```diff
@@ -1,4 +1,105 @@
...
+
+You can nest systems to represent subgroups:
+
+```python
+ModelSystem (root)
+├── sub_systems[0]: ModelSystem(branch_label="group_H2O", particle_indices=[0,1,2])
+│    └── sub_systems[0]: ModelSystem(branch_label="H2O", particle_indices=[0,1,2])
+└── sub_systems[1]: ModelSystem(branch_label="Cu")
+     └── …
+```
```
ndaelman-hu, line 85: This isn't Python syntax: you can just remove the qualifier at the top

docs/model_system/model_system.md
```diff
@@ -1,4 +1,105 @@
...
+
+After normalization, each branch gets a branch_depth, and its composition_formula is set:
+
+```python
+root:             composition_formula = "group_H2O(1)Cu(1)"
+group_H2O:        composition_formula = "H2O(3)"
+H2O (leaf):       composition_formula = "H(1)O(2)"
+Cu   (leaf):      composition_formula = "Cu(1)"
+```
```
ndaelman-hu, line 94: This isn't Python syntax: you can just remove the qualifier at the top

docs/model_system/model_system.md
ndaelman-hu, line 1: I like the chapters "Hierarchical ModelSystems" and "Normalization Workflow" best, since they give me an idea of what this tool models, how to use it, and what it means to normalize it. I'd actually prefer them more fleshed out.

Conversely, these tables should probably be listed in a `reference.md`. They can also be automatically generated, e.g. via 

```yaml
plugins:
  - mkdocstrings:
      handlers:
        python:
          options:
            show_docstring_attributes: true
            docstring_section_style: table
```

JFRudzinski:
I would like to merge as much work as possible before next week. I see no reason why this can't be simply merged. If these review notes are still relevant, I would suggest opening an issue that points to these and then merging this PR

docs/model_system/model_system.md
ndaelman-hu: I guess you just want the syntax highlight for the object. If so, just say so. Here it is sensible.

docs/model_system/model_system.md
ndaelman-hu: I guess the same reason, though the syntax highlighting does less here.

docs/model_system/model_system.md
ndaelman-hu: I thought it was clear here that these can be later on.
Wouldn't forget the references, though. Especially given the amount of information in the pydocs that @EBB2675 added.

JFRudzinski:
closed in favor of #348

PR #191, opened 2025-05-16
Restructure SCF

ndaelman-hu:
Specialize convergence reporting to dedicated properties (`ConvergenceMetrics`), instead of the general `PhysicalProperty`.
Define convergence plots based on `ConvergenceMetrics`. Fall back to inheritance for unit specification.

### DIFF
--- a/src/nomad_simulations/schema_packages/numerical_settings.py
+++ b/src/nomad_simulations/schema_packages/numerical_settings.py
@@ -855,7 +855,7 @@ class SelfConsistency(NumericalSettings):
         """,
     )
 
-    differential_threshold = Quantity(
+    convergence_tolerance = Quantity(
         type=np.float64,
         # guideline: specialize the unit
         description="""
@@ -867,10 +867,10 @@ class SelfConsistency(NumericalSettings):
 
 
 class EnergySelfConsistency(SelfConsistency):
-    differential_threshold = SelfConsistency.m_def.differential_threshold.m_copy()
-    differential_threshold.unit = 'J'
+    convergence_tolerance = SelfConsistency.convergence_tolerance.m_copy()
+    convergence_tolerance.unit = 'J'
 
 
 class ForceSelfConsistency(SelfConsistency):
-    differential_threshold = SelfConsistency.m_def.differential_threshold.m_copy()
-    differential_threshold.unit = 'N'
+    convergence_tolerance = SelfConsistency.convergence_tolerance.m_copy()
+    convergence_tolerance.unit = 'N'
--- a/src/nomad_simulations/schema_packages/numerical_settings.py
+++ b/src/nomad_simulations/schema_packages/numerical_settings.py
@@ -865,6 +865,7 @@ class SelfConsistency(NumericalSettings):
         """,
     )
 
+
 class EnergySelfConsistency(SelfConsistency):
     differential_threshold = SelfConsistency.m_def.differential_threshold.m_copy()
     differential_threshold.unit = 'J'
--- a/src/nomad_simulations/schema_packages/numerical_settings.py
+++ b/src/nomad_simulations/schema_packages/numerical_settings.py
@@ -837,12 +837,6 @@ class KSpace(NumericalSettings):
 class SelfConsistency(NumericalSettings):
     """
     A base section used to define the convergence settings of self-consistent field (SCF) calculation.
-    It determines the conditions for `is_scf_converged` in `SCFOutputs` (see outputs.py). The convergence
-    criteria covered are:
-
-        1. The number of iterations is smaller than or equal to `n_max_iterations`.
-        2. The total change between two subsequent self-consistent iterations for an output property is below
-        `threshold_change`.
     """
 
     # TODO add examples or MEnum?
@@ -863,9 +857,19 @@ class SelfConsistency(NumericalSettings):
 
     differential_threshold = Quantity(
         type=np.float64,
+        # guideline: specialize the unit
         description="""
         Specifies the threshold for the change between two subsequent self-consistent iterations on
         a given output property. The simulation `is_scf_converged` if this total change is below
         this threshold.
         """,
     )
+
+class EnergySelfConsistency(SelfConsistency):
+    differential_threshold = SelfConsistency.m_def.differential_threshold.m_copy()
+    differential_threshold.unit = 'J'
+
+
+class ForceSelfConsistency(SelfConsistency):
+    differential_threshold = SelfConsistency.m_def.differential_threshold.m_copy()
+    differential_threshold.unit = 'N'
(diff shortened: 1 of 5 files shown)


ndaelman-hu:
@ladinesa Do you think that connecting `SelfConsistency` to `ConvergenceMetrics` is appropriate via `after_normalize` (i.e. the parser), or is `normalize` better?

@JFRudzinski I know that we have this division between inputs (system and method) and outputs, but in the case of convergences, keeping them separate just needlessly requires logic to enforce connection. Moreover, there are many kinds of energy differences. 
@ladinesa How feasible would it be all quantities in the same section (both input and output running over the scf cycles) via the mapping annotation parser?

ladinesa:
> @ladinesa Do you think that connecting `SelfConsistency` to `ConvergenceMetrics` is appropriate via `after_normalize` (i.e. the parser), or is `normalize` better?
> 

I would do it at the root section normalize method.


> @JFRudzinski I know that we have this division between inputs (system and method) and outputs, but in the case of convergences, keeping them separate just needlessly requires logic to enforce connection. Moreover, there are many kinds of energy differences. @ladinesa How feasible would it be all quantities in the same section (both input and output running over the scf cycles) via the mapping annotation parser?

You can simply return a list of the scf cycles and mapping parser will run through the list of all quantities in the section with mapping annotation. what you cannot map are the references.

EBB2675:
I don't think I could follow it. 

Just importing `EnergySelfConsistency` already crashes, proving the new class still relies on a `differential_threshold` quantity that isn’t defined.

Do I understand correctly that when 

`differential_threshold = SelfConsistency.m_def.differential_threshold.m_copy()`

runs, Python is still building `SelfConsistency`; its metainfo (`SelfConsistency.m_def`) isn’t populated yet, hence `AttributeError`?

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -861,23 +855,22 @@ class SelfConsistency(NumericalSettings):
         """,
     )
 
-    threshold_change = Quantity(
+    differential_threshold = Quantity(
```
EBB2675: `differential_threshold` feels off because **“differential” ≈ derivative** in math, while here you just mean change between steps. Better options:

- `convergence_tolerance` :  common SCF term 

- `delta_threshold` or `change_threshold` : explicit that it’s the allowed Δ between iterations.

- `step_tolerance` : highlights it’s measured per step.

 If you need the unit-specific subclasses:

 `EnergySelfConsistency.energy_tolerance`, `ForceSelfConsistency.force_tolerance`.

We have a lot more of these thresholds & tolerances, it might be better if we agree on a common ground ( i do not have any preferences )

ndaelman-hu:
> I don't think I could follow it.
> 
> Just importing `EnergySelfConsistency` already crashes, proving the new class still relies on a `differential_threshold` quantity that isn’t defined.
> 
> Do I understand correctly that when
> 
> `differential_threshold = SelfConsistency.m_def.differential_threshold.m_copy()`
> 
> runs, Python is still building `SelfConsistency`; its metainfo (`SelfConsistency.m_def`) isn’t populated yet, hence `AttributeError`?

Okay, will take a look at the breaking attribute. Thx.

Overall, I need my question regarding design resolved first, before I fix testing and the like.
This was just an early stage review request. If you want, you can skip it, and I'll let you know once the tests pass.

EBB2675 (commented):
i will get back to this review once it is finished

ndaelman-hu:
> > @ladinesa Do you think that connecting `SelfConsistency` to `ConvergenceMetrics` is appropriate via `after_normalize` (i.e. the parser), or is `normalize` better?
> 
> I would do it at the root section normalize method.
> 
> > @JFRudzinski I know that we have this division between inputs (system and method) and outputs, but in the case of convergences, keeping them separate just needlessly requires logic to enforce connection. Moreover, there are many kinds of energy differences. @ladinesa How feasible would it be all quantities in the same section (both input and output running over the scf cycles) via the mapping annotation parser?
> 
> You can simply return a list of the scf cycles and mapping parser will run through the list of all quantities in the section with mapping annotation. what you cannot map are the references.

Indeed, mapping annotation cannot handle the references. I can make a workaround, that registers every setting for quick look-up.
Or, we avoid references by having inputs and outputs together, in a single section. The output convergence is a single, array-like quantity. These are typically dispersed over the output file. I was curious whether this could pose any trouble for mapping annotation.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: I'll retain the original `change_threshold` then, though `tolerance` has my personal preference.

ndaelman-hu: > We have a lot more of these thresholds & tolerances, it might be better if we agree on a common ground ( i do not have any preferences )

What do you mean here? _Different thresholds_, as in _different properties_ being tracked, or _different ways of tracking_?

EBB2675: both, hence I didn't distinguish

ndaelman-hu: Okay. The former (_different properties_) is already somewhat captured by the sections' polymorphism, e.g. `EnergySelfConsistency` and `ForceSelfConsistency`. It isn't meant to be an exhaustive list, though strain could be a standard addition.

This approach does not distinguish certain classes of energy measurement, e.g. total energy change, [residuum change](https://www.vasp.at/wiki/index.php/OSZICAR).
Those can either be captured with more inheritance, or a quantity like `label`. I prefer `label`, as I want to avoid extensions to the schema as much as possible.

I couldn't immediately come up with examples of the latter (_different ways of tracking_). @EBB2675 could you list some alternatives? Feel free to also weigh in on how to integrate them into the schema.

EBB2675: Sure—by “different ways of tracking” I just meant the various formulas codes use for the per-step change:

- **Absolute change** ‖Eₙ − Eₙ₋₁‖, ‖Fₙ − Fₙ₋₁‖

- **Relative change** ‖ΔE / Eₙ‖, ‖Δρ / ρₙ‖

- **Max component** max|Δρ(r)|, max|ΔF_i|

- **RMS / L² norm** √(Σ Δx² / N), DIIS error vector norm

 - **Residuum / gradient** ‖[F,P]‖, ‖∇E‖

 - **Compound criteria** e.g. VASP converges when both ΔE and Δρ are below their thresholds.

ndaelman-hu: Thx! Except for _compound criteria_, I'll capture this with an `Enum` quantity.

For compounds, if we assume that convergence may generally be written as `(requirement_1 && requirement_2 && ...) || (... && ...) || ...`, then we need tags for `&&` groups. This is very complex, as the mapping annotations work asynchronously from our POV.
Limiting the convergence requirements to a single `&&` would be much simpler, as a single quantity suffices.

JFRudzinski:
Closing in favor of #260 

@ndaelman-hu

PR #193, opened 2025-05-23
Map `model_system` to `outputs`

ndaelman-hu:
As shown here, `model_method` or `model_system` references in `outputs` are only populated when there is a single, clear section.
This is now extended to cover 1-1 mapping with `model_system`.

Screenshot from 2025-05-23 14-46-33

### DIFF
--- a/src/nomad_simulations/schema_packages/outputs.py
+++ b/src/nomad_simulations/schema_packages/outputs.py
@@ -168,6 +168,7 @@ class Outputs(Time):
                 isinstance(model_systems, list)
                 and isinstance(outputs, list)
                 and len(model_systems) == len(outputs)
+                and len(model_systems) > 0
             ):
                 return model_systems[self.m_parent_index]
         return None
--- a/src/nomad_simulations/schema_packages/outputs.py
+++ b/src/nomad_simulations/schema_packages/outputs.py
@@ -155,11 +155,10 @@ class Outputs(Time):
 
     def set_model_system_ref(self) -> Optional[ModelSystem]:
         """
-        Set the reference to the last `ModelSystem` if this is not set in the output. This is only
-        valid if there is only one `ModelSystem` in the parent section.
+        Provide a suggested `ModelSystem` that corresponds to the output collection.
 
         Returns:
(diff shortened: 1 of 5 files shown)


src/nomad_simulations/schema_packages/outputs.py
```diff
@@ -163,8 +163,13 @@ def set_model_system_ref(self) -> Optional[ModelSystem]:
...
         if self.m_parent is not None:
             model_systems = self.m_parent.model_system
-            if model_systems is not None and len(model_systems) == 1:
-                return model_systems[-1]
+            outputs = self.m_parent.outputs
+            if (
+                isinstance(model_systems, list)
+                and isinstance(outputs, list)
+                and len(model_systems) == len(outputs)
+            ):
```
ndaelman-hu: Note: this version is more restrictive (will fail with arrays, for example), but also more safe. Since `len` is a method of `list`, there's no reliance on duck typing here.

ndaelman-hu:
@JFRudzinski If you want, we can apply this logic to the workflows too.

EBB2675 (approved):
Thank you very much!

Will you be adding the unit tests (e.g. matching lengths vs. mismatched lengths) as well?

ndaelman-hu:
> Thank you very much!
> 
> Will you be adding the unit tests (e.g. matching lengths vs. mismatched lengths) as well?

Fair question. Let's see, are there any edge cases to cover? 3 cases come to mind:

1. other indexable types like `np.array` not being picked up on. Normally, `nomad` should handle the conversion of repeating subsections to `list`.
2. `m_parent_index` having a non-numerical or abnormal value. The stand-out case is `-1` for non-repeating subsections. This is also on `nomad` to enforce.
3. `IndexError` due to the lists being empty.

I patched point 3, thank you for making me re-evaluate it. Points 1 and 2 would be due to failures on `nomad`'s side.
I can check that these cases lead to failure, but recovery from them is beyond our purview.

ndaelman-hu:
@EBB2675 Have extended the testing. Will merge, but retain the branch for a day, in case I get further feedback.

PR #195, opened 2025-06-02
Positive data types

### LINKED ISSUE #165: Introduce custom types
Use custom types to enforce extra conditions, e.g. strictly positive values. This technique aims to reduce the burden on normalization.

ndaelman-hu:
Adds data types that enforce positive and bound data types.

Closes #165

### DIFF
--- a/src/nomad_simulations/schema_packages/data_types.py
+++ b/src/nomad_simulations/schema_packages/data_types.py
@@ -35,7 +35,6 @@ class PositiveInt(ExactNumber):
             raise ValueError(f'Invalid dtype for {self.__class__.__name__}.')
 
         super().__init__(dtype)
-        self._np_base = np.integer
 
     def convertible_from(self, other):
         if other in (int, np.int64, np.int32, np.int16, np.int8):
@@ -84,7 +83,6 @@ class PositiveFloat(InexactNumber):
             raise ValueError(f'Invalid dtype for {self.__class__.__name__}.')
 
         super().__init__(dtype)
-        self._np_base = np.inexact
 
     def convertible_from(self, other):
         if other in (float, np.float64, np.float32, np.float16):
@@ -145,7 +143,6 @@ class UnitFloat(InexactNumber):
             raise ValueError(f'Invalid dtype for {self.__class__.__name__}.')
 
         super().__init__(dtype)
-        self._np_base = np.inexact
         self._min_inclusive = True
         self._max_inclusive = True
 
--- a/src/nomad_simulations/schema_packages/data_types.py
+++ b/src/nomad_simulations/schema_packages/data_types.py
@@ -21,7 +21,7 @@ from collections.abc import Generator
 from typing import Any
 
 import numpy as np
-from nomad.metainfo.data_type import Datatype, normalize_type
+from nomad.metainfo.data_type import Datatype, Number, normalize_type
 
 # Match patterns like '[0,3)', '(0,5]', '[1,)', '(,10)', etc.
 bounds_patt = re.compile(r'^([\[\(])(-?\d*\.?\d*|),\s*(-?\d*\.?\d*|)([\]\)])$')
@@ -165,7 +165,7 @@ class Bound:
         return f'Bound({self.get_bounds_str()!r})'
 
 
-class BoundedNumber(Datatype):
+class BoundedNumber(Number):
     """
     Bounded numeric data type that coordinates between NOMAD's dtype system and bounds checking.
 
@@ -185,19 +185,22 @@ class BoundedNumber(Datatype):
             dtype: NOMAD data type (e.g., 'm_int32', 'm_float64', int, float) or Datatype instance
             bounds: Bound instance specifying the valid range
         """
-        super().__init__()
         # Support reconstruction from serialized data (called by normalize_type)
         if dtype is None:
-            # Will be populated by normalize_flags
+            # Will be populated by normalize_flags - use a placeholder dtype
+            super().__init__(float)
             self._base_datatype = None
             self._bounds = None
         else:
             # Normal initialization
             if isinstance(dtype, Datatype):
                 self._base_datatype = dtype
+                underlying_dtype = getattr(dtype, '_dtype', dtype)
             else:
                 self._base_datatype = normalize_type(dtype)
+                underlying_dtype = dtype
 
+            super().__init__(underlying_dtype)
             self._bounds = bounds if bounds is not None else Bound()
 
     def convertible_from(self, other: Any) -> bool:
@@ -263,13 +266,6 @@ class BoundedNumber(Datatype):
         if self._base_datatype is not None:
             self._base_datatype.attach_definition(definition)
         return self
-
-    @property
-    def _dtype(self):
-        """Get the underlying dtype."""
-        if self._base_datatype is None:
-            return None
-        return getattr(self._base_datatype, '_dtype', None)
 
     def __repr__(self):
         if self._base_datatype is None:
--- a/src/nomad_simulations/schema_packages/data_types.py
+++ b/src/nomad_simulations/schema_packages/data_types.py
@@ -20,12 +20,19 @@ import numpy as np
 from nomad.metainfo.data_type import ExactNumber, InexactNumber
 
 
-class PositiveInt(ExactNumber):
+class BoundedInt(ExactNumber):
     """
-    Integer data type that only allows positive values (> 0).
+    Integer data type with configurable bounds that can include +/- infinity.
+
+    Examples:
+        BoundedInt().min(1)                    # >= 1 (positive integers)
+        BoundedInt().min(0)                    # >= 0 (non-negative integers)
+        BoundedInt().min(1, inclusive=False)   # > 1
+        BoundedInt().max(10)                   # <= 10
+        BoundedInt().min(1).max(10)            # 1 <= value <= 10
     """
 
-    __slots__ = ()
+    __slots__ = ('_min_value', '_max_value', '_min_inclusive', '_max_inclusive')
 
     def __init__(self, *, dtype: type[int | np.int32] = int):
         if isinstance(dtype, np.dtype):
@@ -35,6 +42,22 @@ class PositiveInt(ExactNumber):
             raise ValueError(f'Invalid dtype for {self.__class__.__name__}.')
 
         super().__init__(dtype)
+        self._min_value = float('-inf')
+        self._max_value = float('inf')
+        self._min_inclusive = True
+        self._max_inclusive = True
+
+    def min(self, value, *, inclusive: bool = True):
+        """Set minimum bound. Supports +/- infinity."""
+        self._min_value = value
+        self._min_inclusive = inclusive
+        return self
+
+    def max(self, value, *, inclusive: bool = True):
+        """Set maximum bound. Supports +/- infinity."""
+        self._max_value = value
+        self._max_inclusive = inclusive
+        return self
 
     def convertible_from(self, other):
         if other in (int, np.int64, np.int32, np.int16, np.int8):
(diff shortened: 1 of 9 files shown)


ladinesa:
I would suggest to implement this in nomad, but either way it would be better to ask @TLCFEM

ndaelman-hu:
> I would suggest to implement this in nomad, but either way it would be better to ask @TLCFEM

I agree that it would be a better fit there, but I don't want to deal with all this red tape right now. First I implement and prove its utility, later on we potentially move it around. Theodore is aware of this PR. You are both invited to give feedback.

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,260 @@
...
+            if isinstance(item, list):
+                yield from _flatten_values(item)
+            else:
+                yield item
+    else:
+        yield data
+
+
+def _check_bounds(
+    value: Union[int, float],
```
TLCFEM: So obsolete, use `int | float`.
https://peps.python.org/pep-0604/

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,260 @@
...
+    def convertible_from(self, other: Any) -> bool:
+        """Check if this data type can convert from another type."""
+        return other in {int, np.int32, np.int16, np.int8}
+
+    def _filter_values_for_validation(self, flat_values: list[Any]) -> list[Any]:
+        """No filtering needed for integers - all values are valid for bounds checking."""
+        return flat_values
+
+    def normalize(self, value: Any, **kwargs: Any) -> Any:
+        normalized_value = ExactNumber.normalize(self, value, **kwargs)
```
TLCFEM: This `kwargs` contains the parent section that holds the data, in case it becomes dynamic somehow and relies on other quantities, you can get section from it can get the information accordingly, there are example usages in the codebase.

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,260 @@
...
+    def convertible_from(self, other: Any) -> bool:
+        """Check if this data type can convert from another type."""
+        return other in {float, np.float64, np.float32, np.float16}
+
+    def _filter_values_for_validation(self, flat_values: list[Any]) -> list[Any]:
+        """Filter out NaN values for float bounds checking."""
+        return [v for v in flat_values if not (isinstance(v, float) and np.isnan(v))]
+
+    def normalize(self, value: Any, **kwargs: Any) -> Any:
+        normalized_value = InexactNumber.normalize(self, value, **kwargs)
```
TLCFEM: I'd prefer `return self._validate_bounds(InexactNumber.normalize(self, value, **kwargs))`, if `_validate_bounds` returns valid values (happy path) and throws (bad path).

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,260 @@
...
+    left = '[' if min_inclusive else '('
+    right = ']' if max_inclusive else ')'
+
+    min_str = str(min_value) if np.isfinite(min_value) else '-∞'
+    max_str = str(max_value) if np.isfinite(max_value) else '∞'
+
+    return f'{left}{min_str}, {max_str}{right}'
+
+
+class BoundedDataType(ABC):
```
TLCFEM: This thing looks unnecessarily over-engineered.
One can directly use for example, `BoundedFloat(dtype=float,range='[0,3)')` to provide all necessary information.
I do not see the benefits of `BoundedFloat().min(0,inclusive=True).max(3,inclusive=False)`.
But I am okay with the current impl.

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,260 @@
...
+
+    Examples:
+        BoundedInt().min(1)                    # >= 1 (positive integers)
+        BoundedInt().min(0)                    # >= 0 (non-negative integers)
+        BoundedInt().min(1, inclusive=False)   # > 1
+        BoundedInt().max(10)                   # <= 10
+        BoundedInt().min(1).max(10)            # 1 <= value <= 10
+    """
+
+    __slots__ = ('_min_value', '_max_value', '_min_inclusive', '_max_inclusive')
```
TLCFEM: This line shall go to the parent?

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,260 @@
...
+    _error_message_prefix = 'All values'
```
TLCFEM: This line is not necessary?

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,260 @@
...
+
+    Examples:
+        BoundedFloat().min(0)                          # >= 0 (non-negative)
+        BoundedFloat().min(0, inclusive=False)         # > 0 (positive)
+        BoundedFloat().min(0).max(1)                   # [0, 1] (unit interval)
+        BoundedFloat().min(0).max(1, inclusive=False)  # [0, 1)
+        BoundedFloat().max(0)                          # <= 0 (non-positive)
+    """
+
+    __slots__ = ('_min_value', '_max_value', '_min_inclusive', '_max_inclusive')
```
TLCFEM: Same.

tests/test_data_types.py
```diff
@@ -0,0 +1,280 @@
...
+        bounded_float = setup_datatype_for_testing(
+            BoundedFloat().min(0).max(1), shape=None
+        )
+
+        test_values = [0.5, float('nan'), 0.8]
+        filtered = bounded_float._filter_values_for_validation(test_values)
+        assert filtered == [0.5, 0.8]  # NaN should be filtered out
+
+
+class TestBoundsStringRepresentation:
```
TLCFEM: Try add a test that `m_from_dict(section.m_to_dict())` to cover both deserialization and serialization.

TLCFEM:
> > I would suggest to implement this in nomad, but either way it would be better to ask @TLCFEM
> 
> I agree that it would be a better fit there, but I don't want to deal with all this red tape right now. First I implement and prove its utility, later on we potentially move it around. Theodore is aware of this PR. You are both invited to give feedback.

I'd prefer keeping it scoped within this repo only for the moment.

The current implementation handles data only.
You may want to validate if the schema can be serialized/deserialized, it may or may not work out-of-the-box.
Also the type needs to support conversion to mongo/elasticsearch types.

So the tasks:

- [x] bound checks
- [x] (de)serialization of data
- [ ] (de)serialization of schema/definition
- [ ] type detection/conversion (for elasticsearch mainly, if the quantities needs to be indexed)

src/nomad_simulations/schema_packages/data_types.py
ndaelman-hu: It fails to instantiate when moved to the parent (had this originally). I suspect that the reason is the double inheritance.

src/nomad_simulations/schema_packages/data_types.py
ndaelman-hu: Neat, I like this format way better! Is more in line with other type systems.
Since it is supported from 3.10 onward, I can apply it.

src/nomad_simulations/schema_packages/data_types.py
ndaelman-hu: That's good to know, will keep it in mind for the follow-up to this PR.

src/nomad_simulations/schema_packages/data_types.py
ndaelman-hu: I was trying to mimic the way of toggling defaults as with your data types. You're right, though, the bound handling should be more explicit here, rather than having infinite for defaults. We actually already have masks for some common defaults.

Do you have a specific suggestion for infinities? Else, I'd just leave it blank, e.g. `range='[0,)')`.

src/nomad_simulations/schema_packages/data_types.py
ndaelman-hu: If I rework the interface, I only need 1 class, so this is resolved then.

src/nomad_simulations/schema_packages/data_types.py
ndaelman-hu: Resolved with a single consistent interface.

TLCFEM (commented):
While the previous implementation is based on inheritance, here those types are function calls, they need to be consistent so you choose one approach.

Also those calls cannot customise dtype.

I think you really just need

```py
a_type = bounded_float(dtype=float, bounds=Bound('[1,inf)'))
```

and this bound obj implements the checker,

```py
self.bounds.check(value, ...)
```

No need for multiple inheritance, and this checker does not get mixed into inheritance.

src/nomad_simulations/schema_packages/data_types.py
ndaelman-hu: So, atm I handle the bound checking differently for NaN, depending on whether `dtype` is a `float` (filter out NaN) or `int` (supposes there are no NaN passed on). Can I delegate this to the NOMAD data types?

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,260 @@
...
+        return self
+
+    @abstractmethod
+    def _filter_values_for_validation(self, flat_values: list[Any]) -> list[Any]:
+        """Filter values for validation (e.g., remove NaN for floats)."""
+        pass
+
+    def _validate_bounds(self, normalized_value: Any) -> None:
+        """Common bounds validation logic."""
+        flat_values = list(_flatten_values(normalized_value))
```
TLCFEM: If it is converted to a list why bother having a generator interface?

You are using a slow mechanism to loop over the data and also make unnecessary copies of potentially large data.

~Overriding _normalize_impl is much simpler unless there is more complex logic.~

src/nomad_simulations/schema_packages/data_types.py
ndaelman-hu: Like this?

src/nomad_simulations/schema_packages/data_types.py
TLCFEM: We do not check nan and inf in the existing codebase.

TLCFEM:
Decoupling also makes it simpler for later integration, less boilerplate.

ndaelman-hu:
> While the previous implementation is based on inheritance, here those types are function calls, they need to be consistent so you choose one approach.
> 
> Also those calls cannot customise dtype.
> 
> I think you really just need
> 
> ```python
> a_type = bounded_float(dtype=float, bounds=Bound('[1,inf)'))
> ```
> 
> and this bound obj implements the checker,
> 
> ```python
> self.bounds.check(value, ...)
> ```
> 
> No need for multiple inheritance, and this checker does not get mixed into inheritance.

Okay, I decoupled range checking into `class Bound`, and use `class BoundNumber` to call the checks and delegate of the instantiation to NOMAD. Lemme know your thoughts, and if you approve, I move on to the tests.

Regarding de/serialization, I don't think that ES, Mongo, or JSON have proper counter versions of bound types. I would rely on NOMAD's serialization of the corresponding numerical type. This means, of course, that deserialization of an old schema may throw `ValueError`s.

TLCFEM:
No. Whatever new type added needs to support:
1. (de)serialization of data
2. (de)serialization of schema
3. make the indexing system aware of the new type otherwise it cannot be indexed.

Currently only 1. is implemented.

src/nomad_simulations/schema_packages/data_types.py
ndaelman-hu: Okay, I will allow NaN to pass the bound checks regardless. `Bound` isn't responsible for handling its semantics.

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,263 @@
...
+        # Check if this is a float type for NaN filtering
+        datatype_name = self._base_datatype.__class__.__name__.lower()
+        filter_nan = 'float' in datatype_name or 'inexact' in datatype_name
+
+        self._bounds.check(normalized_value, filter_nan=filter_nan)
+        return normalized_value
+
+    def serialize_self(self):
+        """Serialize the datatype configuration."""
+        return self._base_datatype.serialize_self()
```
TLCFEM: This is not correct.

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,263 @@
...
+        min_str = str(self._min_value) if np.isfinite(self._min_value) else '-∞'
+        max_str = str(self._max_value) if np.isfinite(self._max_value) else '∞'
+
+        return f'{left}{min_str}, {max_str}{right}'
+
+    def __repr__(self) -> str:
+        return f'Bound({self.get_bounds_str()!r})'
+
+
+class BoundedNumber:
```
TLCFEM: I think it is better to

```py
class BoundedNumber(Number):
    ....
```

or whatever parent class (ExactNumber or InexactNumber) appropriate.

1. override `serialize_self` so that serializing the data type itself yields a dict containing also the bounds information, see reference implementation `Primitive.serialize_self`.
2. override `normalize` to apply bound check logic.
3. revise function `normalize_type` so that it is aware of the bounds, and it can deserialize a dict into a type instance.

If you want to utilize the existing mechanism, you want to name the types as `m_int_bounded`, `m_float_bounded`.
The prefix must be `m_`, followed by `int` or `float`. Precision does not matter, whether `int32` or `int64`, or just `int`, extra information can be appended, so the pattern shall be `m_int*` or `m_float*`.

Otherwise override `standard_type` to return proper python compatible string representation.

`convertible_from` is not very important, let's assume it returns True for simplicity, and we are using widest types so we go 64-bit first.

TLCFEM: The current wrapper style makes it difficult to later be integrated into the existing system.
Also since it is a wrapper it is not a `DataType` so a lot of things will break.

tests/test_data_types.py
ndaelman-hu: I've added them under `TestBoundNumber.test_serialization_and_reconstruction` and `TestNOMADIntegration`

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,319 @@
...
+
+        Args:
+            dtype: NOMAD data type (e.g., 'm_int32', 'm_float64', int, float) or Datatype instance
+            bounds: Bound instance specifying the valid range
+        """
+        super().__init__()
+        # Support reconstruction from serialized data (called by normalize_type)
+        if dtype is None:
+            # Will be populated by normalize_flags
+            self._base_datatype = None
```
TLCFEM: Do proper OO please.
You do not need an internal `_base_datatype`, let alone wrapping around it.

The current implementation will fail the following checks.

```py
a_type = BoundedNumber(...)

isinstance(a_type, Number) # False
isinstance(a_type, ExactNumber) # False
isinstance(a_type, InexactNumber) # False
```

ndaelman-hu:
> No. Whatever new type added needs to support:
> 
>     1. (de)serialization of data
> 
>     2. (de)serialization of schema
> 
>     3. make the indexing system aware of the new type otherwise it cannot be indexed.
> 
> 
> Currently only 1. is implemented.

I added points 2 and 3 now.

TLCFEM:
This is probably all you need.

```py
from nomad.metainfo import ExactNumber


class Bound:
    def __init__(self): ...

    def check(self, value, **kwargs):
        # handles both scalar and array values
        # logic goes here
        return value

    def __repr__(self):
        return '[1,2)'


class m_int_bounded(ExactNumber):
    def __init__(self, dtype, bound=None):
        super().__init__(dtype=dtype)
        self.bound = bound or Bound()

    def convertible_from(self, other):
        # todo: need more work
        return True

    def serialize_self(self):
        return super().serialize_self() | {'type_bound': self.bound}

    def normalize(self, value, **kwargs):
        return self.bound.check(super().normalize(value, **kwargs))

    def standard_type(self):
        return 'int'
```

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,316 @@
...
+            'type_bound': str(self.bound),
+        } | self.flags
+
+    def normalize_flags(self, flags: dict):
+        """Reconstruct from serialized data."""
+        bounds_str = flags.get('type_bound', '')
+        self.bound = Bound(bounds_str)
+        # Apply any flags to base datatype
+        if hasattr(super(), 'normalize_flags'):
+            super().normalize_flags(flags)
```
TLCFEM:
```suggestion
        super().normalize_flags(flags)
```

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,316 @@
...
+        else:
+            return False
+
+    def serialize_self(self):
+        """Serialize the datatype configuration."""
+        return {
+            'type_kind': 'custom',
+            'type_data': f'{self.__class__.__module__}.{self.__class__.__name__}',
+            'type_bound': str(self.bound),
+        } | self.flags
```
TLCFEM:
```suggestion
        return super().serialize_self() | {'type_bound': self.bound}
```

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,316 @@
...
+    if isinstance(data, np.ndarray):
+        yield from data.flatten()
+    elif isinstance(data, list):
+        for item in data:
+            if isinstance(item, list):
+                yield from _flatten_values(item)
+            else:
+                yield item
+    else:
+        yield data
```
TLCFEM: Just return a list, this generator is consumed later as `list(_flatten_values(data))`, no point of using a generator.

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,316 @@
...
+        else:
+            return False
+
+    def serialize_self(self):
+        """Serialize the datatype configuration."""
+        return {
+            'type_kind': 'custom',
+            'type_data': f'{self.__class__.__module__}.{self.__class__.__name__}',
+            'type_bound': str(self.bound),
+        } | self.flags
```
TLCFEM:
```suggestion
        return return super().serialize_self() | {'type_bound': self.bound}
```

src/nomad_simulations/schema_packages/data_types.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def normalize_flags(self, flags: dict):
+        """Reconstruct from serialized data."""
+        bounds_str = flags.get('type_bound', '')
+        self.bound = Bound(bounds_str)
+        # Apply any flags to base datatype
+        if hasattr(super(), 'normalize_flags'):
+            super().normalize_flags(flags)
```
TLCFEM:
```suggestion
        super().normalize_flags(flags)
```

TLCFEM (approved):
After those minor changes, it looks good enough to be merged at any time.

src/nomad_simulations/schema_packages/data_types.py
ndaelman-hu: Sorry, this was a remnant of the previous implementations. Thx for capturing it.

src/nomad_simulations/schema_packages/data_types.py
ndaelman-hu: So the proper `type_kind` is `python` instead of `custom`? Will amend the tests to reflect this.

ndaelman-hu: Just note that deserialization then provides a python type, and we need the `m_def` to apply the bound constraint again. I guess this is as you intended (without the NOMAD schema, a data scientist can treat it as a regular python number), but just want to verify.

PR #196, opened 2025-06-17
H5md refactor

JFRudzinski:
- add generic label to atoms_state
- reintroduce CGBeadState (could be moved to another PR, no tests or usage yet)
- add some new system types for molecules
- add velocities to top-level system
- add some functions for working with bond_lists and molecules (and test for these)

### DIFF
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -1337,9 +1337,61 @@ class ModelSystem(System):
         if bonds.size == 0:
             return False
 
+        particle_indices = self.particle_indices
+        if particle_indices is None:
+            n_particles = (
+                len(self.positions) if self.positions is not None else self.n_particles
+            )
+            particle_indices = np.arange(n_particles, dtype=np.int32)
+
         # Build graph and check if connected
         graph = nx.Graph()
-        graph.add_nodes_from(self.particle_indices)
+        graph.add_nodes_from(particle_indices)
         graph.add_edges_from(bonds)
 
         return nx.is_connected(graph)
+
+    def is_molecule(self) -> bool:
+        """
+        Checks if the current subsystem forms a contiguous and isolated molecule:
+        - All particles are connected (single connected component).
+        - No bonds connect particles inside this subsystem to particles outside it.
+
+        Returns:
+            bool: True if the subsystem is an isolated molecule, False otherwise.
+        """
+        import networkx as nx
+
+        # Internal bonds for this subsystem
+        bonds = self.get_bond_list(set_local=False)
+
+        # Handle case: no bonds
+        if bonds.size == 0:
+            return False
+
+        # Determine particle indices (fallback to range if None)
+        particle_indices = self.particle_indices
+        if particle_indices is None:
+            n_particles = (
+                len(self.positions) if self.positions is not None else self.n_particles
+            )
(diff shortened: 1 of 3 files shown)


src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1028,6 +1042,55 @@ def get_chemical_symbols(self, logger: 'BoundLogger') -> list[str]:
...
+                symbol = particle_state.bead_symbol
+            if not symbol:
+                logger.warning('missing symbol in ParticleState.')
+            symbols.append(symbol)
+        return symbols
+
+    # ! Check for chemical elements
+    def are_valid_chemical_symbols(self, logger: 'BoundLogger') -> bool:
+        """
+        Checks if the chemical symbols in the particle_states are valid chemical symbols.
```
EBB2675: minor suggestion for docstring, since in this helper func we rely on `symbols2numbers`:

_Validate that all element symbols in the particle_states can be mapped to atomic numbers recognised by ASE._

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1028,6 +1042,55 @@ def get_chemical_symbols(self, logger: 'BoundLogger') -> list[str]:
                 chemical_symbols.append(particle_state.chemical_symbol)
         return chemical_symbols
 
+    # TODO: symbols should be a property right?
+    # ? To replace get_chemical_symbols
+    def get_symbols(self, logger: 'BoundLogger') -> list[str]:
+        """
```
EBB2675, line 1049: `get_symbols` looks good from my side

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1028,6 +1042,55 @@ def get_chemical_symbols(self, logger: 'BoundLogger') -> list[str]:
...
+    # ! Check for chemical elements
+    def are_valid_chemical_symbols(self, logger: 'BoundLogger') -> bool:
+        """
+        Checks if the chemical symbols in the particle_states are valid chemical symbols.
+        Args:
+            logger (BoundLogger): The logger to log messages.
+        Returns:
+            bool: True if all chemical symbols are valid, False otherwise.
+        """
+        symbols = self.get_chemical_symbols(logger)
```
EBB2675: I assumed the new `get_symbols()` was intended to supersede `get_chemical_symbols()`, but in this context we must pass pure element symbols to ASE, am I correct? So we either have two helpers, one generic `get_symbols` and one element-only `get_chemical_symbols`, or move to a single helper with a filter flag?

EBB2675 (approved):

src/nomad_simulations/schema_packages/model_system.py
JFRudzinski: Yes, the idea was to replace get_chemical_symbols(). When I started this branch I wasn't sure exactly the scope, but now I would like to address this issue in another PR, when I actually implement functional CGBeadStates. For now, I will just leave both functions. I did replace get_chemical_symbols() in 2-3 places where I think it is valid, and the tests still run, so I think I should not mess anything up for you.

JFRudzinski: If this seems fine with you, then I would opt for a quick merge of this PR, then I can move on to the next step. Please confirm.

EBB2675: sure :+1:

PR #199, opened 2025-07-09
Clarify subsystem referencing

### LINKED ISSUE #198: Clarification on SubSystem specification

```python
    particle_indices = Quantity(
        type=np.int32,
        shape=['*'],
        description="""
        """,
    )
```

ndaelman-hu:
Closes #198


ndaelman-hu:
@EBB2675 this was my first attempt, before posting the issue. We may use it as a jumping-off board.

PR #200, opened 2025-07-11
particle_indices and particle_states 

### LINKED ISSUE #198: Clarification on SubSystem specification

EBB2675:
## What this PR does

* **Clarifies indexing**  
  * `particle_indices` are now clearly documented as *global indices
    in the top-level `ModelSystem`* (not relative to the direct parent).
  * The `ModelSystem` and `particle_states` doc-strings were updated to repeat
    the rule and give a concrete example.

* **Makes `particle_states` stable**  
  * Added two small helpers:
    * `_walk_depth_first()` – iterate over all sub-systems.
    * `_consolidate_particle_states()` – if the root list is empty, pull the
      states up from the children and build a full, ordered list.
  * The consolidation runs as the very first step in `ModelSystem.normalize()`.

* **RElated tests**  
  * New test `test_consolidate_particle_states_from_children` checks that a
    root with no states gets them correctly from a child.

## Why 

* People weren’t sure whether `particle_indices` should be relative to the
  immediate parent or to the root – we now say “always the root”.
* Some parsers create all `particle_states` at the root, others only in the
  leaves.  After this patch both approaches work without extra glue code.

## Bottom linw

* No schema fields were renamed or removed.
* Existing archives that already store all states at the root behave exactly
  as before (consolidation is a no-op).  
* Child systems can still keep their own copies, nothing is deleted.

### DIFF
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -1130,8 +1130,77 @@ class ModelSystem(System):
 
         return system_type, dimensionality
 
+    def _walk_depth_first(self):
+        """Yield self and every descendant ModelSystem (depth-first)."""
+        yield self
+        for sub in self.sub_systems:
+            yield from sub._walk_depth_first()
+
+    def _consolidate_particle_states(self, logger):
+        """
+        Make sure the representative (top-level) ModelSystem owns a complete,
+        ordered ``particle_states`` list, even if only descendants populated it.
+        """
+        if not self.is_representative:
+            return  # run only on the root
+
+        # 1. Determine how many atoms we expect
+        n_atoms = self.n_particles
+        if n_atoms is None:
+            max_idx = -1
+            for ms in self._walk_depth_first():
+                if ms.particle_indices:
+                    max_idx = max(max_idx, *ms.particle_indices)
+            n_atoms = max_idx + 1
+
+        if n_atoms <= 0:
+            logger.debug('No particles detected; nothing to consolidate.')
+            return
+
+        # 2. Ensure self.particle_states has the right length
+        if not self.particle_states:
+            self.particle_states = [None] * n_atoms
+        elif len(self.particle_states) < n_atoms:
+            self.particle_states.extend([None] * (n_atoms - len(self.particle_states)))
+
+        # 3. Walk the tree and pull states up
+        for ms in self._walk_depth_first():
+            if not ms.particle_states:
+                continue
+            has_indices = (
+                ms.particle_indices is not None and len(ms.particle_indices) > 0
+            )
+            abs_indices = (
+                list(ms.particle_indices)
+                if has_indices
+                else list(range(len(ms.particle_states)))
+            )
+
+            for local_i, abs_i in enumerate(abs_indices):
+                state = ms.particle_states[local_i]
+                if state is None:
+                    continue
+                if self.particle_states[abs_i] is None:
+                    self.particle_states[abs_i] = state
+                elif self.particle_states[abs_i] is not state:
+                    logger.warning(
+                        'Conflicting ParticleState at absolute index %d; '
+                        'keeping the first encountered instance.',
+                        abs_i,
+                    )
+
+        # 4. Sanity check
+        if any(st is None for st in self.particle_states):
+            logger.error(
+                'Consolidation failed: some particle_states remain unresolved.'
+            )
+
     def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
         super().normalize(archive, logger)
+
+        # Check particle_states
+        if self.is_representative:
+            self._consolidate_particle_states(logger)
 
         # We don't need to normalize if the system is not representative
         if is_not_representative(model_system=self, logger=logger):
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -982,7 +982,27 @@ class ModelSystem(System):
     particle_states = SubSection(
         section_def=ParticleState.m_def,
         repeats=True,
-        description='Particle states',
+        description="""
+            Per-particle state information.
+
+            - Ordering: In the representative (top-level) `ModelSystem` this list
+            has exactly `n_particles` entries and the order matches
+            `positions[i]`.
+
+            - Normalisation behavior: If parsers fill the list only in child
+            subsystems, the normalizer pulls those states up so the root is always
+            complete. 
+
+            Example
+            -------
+            A water molecule (H₂O):
+
+                positions       : [[…], […], […]]      # 3 atoms
+                particle_states :
+                    [0] AtomsState(H)
+                    [1] AtomsState(H)
+                    [2] AtomsState(O)
+            """,
     )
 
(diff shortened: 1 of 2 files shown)


EBB2675:
Ruff trips on PLC0415 (import not top-level) in:

 -  tests/test_data_types.py

Ignore rule in config or add # noqa: PLC0415 to those lines? Let me know and I’ll patch whichever you prefer @ndaelman-hu

EBB2675:
this is an experimental implementation, for when you wish to populate the top-level particle states through the children, please briefly let me know what you think

JosePizarro3:
> Ruff trips on PLC0415 (import not top-level) in:

Is the ruff version pumped for >=0.11.4? If not, do you mind trying if the pipeline passes then? :-)

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -979,7 +982,27 @@ class ModelSystem(System):
     particle_states = SubSection(
         section_def=ParticleState.m_def,
         repeats=True,
-        description='Particle states',
+        description="""
+            Per-particle state information.
```
JosePizarro3, line 986: _Particle state of each of the particles conforming the `ModelSystem`. This is a list of `n_particles` elements and the order matches that of `positions`._

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -979,7 +982,27 @@ class ModelSystem(System):
...
+
+            - Ordering: In the representative (top-level) `ModelSystem` this list
```
JosePizarro3, line 988: this ordering could go in one line as I suggested above

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -979,7 +982,27 @@ class ModelSystem(System):
...
+            has exactly `n_particles` entries and the order matches
+            `positions[i]`.
+
+            - Normalisation behavior: If parsers fill the list only in child
```
JosePizarro3, line 992: you could explain the normalisation behavior in the normalizer function :)

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1127,9 +1150,78 @@ def resolve_system_type_and_dimensionality(
 
         return system_type, dimensionality
 
+    def _walk_depth_first(self):
```
JosePizarro3, line 1153: maybe @ladinesa should help you here regarding the ordering of the normalization.

for now, this looks good to me, but this is one of those things we never ended up fixing in the src code of nomad-lab

JFRudzinski (approved):
I did not look at the detailed code in depth, but the approach is looking nice to me :+1:

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1127,9 +1150,78 @@ def resolve_system_type_and_dimensionality(
...
+        """Yield self and every descendant ModelSystem (depth-first)."""
+        yield self
+        for sub in self.sub_systems:
+            yield from sub._walk_depth_first()
+
+    def _consolidate_particle_states(self, logger):
+        """
+        Make sure the representative (top-level) ModelSystem owns a complete,
```
JosePizarro3, line 1161: Mmm, as far as I see in the logic, this is copy-pasting the info of the `ParticleState` sections might exists in the children of the global ModelSystem parent (from L 1187 to 1210)

However, this might explode the archive with a lot of repetitive data

JosePizarro3 (commented):
Looking good! Just make sure of this archive size explosion I say about consolidating the particle states from the children. It might be that for systems with thousands of atoms, it really is too much

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: Would rephrase it without mentioning parsing necessarily:

```suggestion
            - Normalisation behavior: construct a common root from the subsystems.
```

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu, line 1: I would be clearer with the terminology, and avoid mixing _parent / child_ with _super- / subsystem_. The latter is more appropriate here.
The former mimics the nomenclature used for _class inheritance_, but that is decidedly not the mechanism used here (it's class composition).

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1127,9 +1150,78 @@ def resolve_system_type_and_dimensionality(
 
         return system_type, dimensionality
 
+    def _walk_depth_first(self):
+        """Yield self and every descendant ModelSystem (depth-first)."""
+        yield self
+        for sub in self.sub_systems:
+            yield from sub._walk_depth_first()
+
+    def _consolidate_particle_states(self, logger):
```
ndaelman-hu, line 1159: So, this method covers the very specific case where a _root_ system exists (since you need a reference for the `particle_indices`), it populated quantities (e.g. `positions`, `labels`) there, but somehow `particle_states` were written to the subsystems?

I think we're trying to overcompensate for lousy parsing during normalization time, which can get unruly very fast.
If you foresee an issue with parsing `particle_states`, we should look into providing a tool there (for during the conversion phase).

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: Given that we want the state definitions at the top, copying them is appropriate.
We could remove the originals, though. Overall, I'd focus more on the other point I raised above.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1127,9 +1150,78 @@ def resolve_system_type_and_dimensionality(
...
+
+        if n_atoms <= 0:
+            logger.debug('No particles detected; nothing to consolidate.')
+            return
+
+        # 2. Ensure self.particle_states has the right length
+        if not self.particle_states:
+            self.particle_states = [None] * n_atoms
+        elif len(self.particle_states) < n_atoms:
+            self.particle_states.extend([None] * (n_atoms - len(self.particle_states)))
```
ndaelman-hu, line 1184: I think these blocks can be removed. Their logic serves 2 minor purposes:

- ensuring the population of `n_atoms`. This can be done _independently_ during normalization.
- presetting the root's particle_states, which can be done by simply returning the constructed `list` and assigning it (if non-empty) in the normalization body.

ndaelman-hu (commented):
Thank you for aligning the definitions.
Regarding the normalization, I would instead opt for a tool that can help with the right setup during parsing.

ndaelman-hu:
@EBB2675 given that you use the same section definition for the root system as any subsystem, this means losing some control over quantities that shouldn't be populated downstream.

In regular OOP, one would use different class definitions, but I remember that this wasn't feasible / wishful here.
Still, NOMAD provides some alternatives. You might want to check out the `Filter` annotation in `nomad-FAIR/nomad/datamodel/metainfo/annotations.py`. Never used it myself (Area A might've), but maybe it's useful. The main question is whether it changes the root's `m_def` too?

PR #201, opened 2025-07-13
Physical property tweaking

ndaelman-hu:
As discussed, I broke off the modifications to `PhysicalProperty` in #197 (which was used to test plotting and mapping parser, as well as modifications to `outputs`) into its own PR.

This PR incrementally applies some modifications. If you disagree about any particular ones, mention the commit hash and we can see to cherry-pick it out. The rough order includes:

- Removal of superfluous logic, e.g. connected to `variables`.
- The integration of `PhysicalContribution` into `PhysicalProperty`. This aims to half the number of definitions for each property. There is a note for a follow-up PR, but we should consult Theodore here first.
- Addition of `sub_plots` to manage aggregation of plots from contributions.
- Cleaning of dependent sections + removal of unnecessary initializers.
- Update of tests.

@EBB2675 I would add the spin base section to a specialized version of `PhysicalProperty`. Would leave this as a short follow-up.
Similarly specializing the contributions for electronics requires the addition of `OrbitalsState` to `ModelSystem` itself, as I already raised.

@everyone Feel free to raise suggestions for the remaining in-code comments.


src/nomad_simulations/schema_packages/physical_property.py
```diff
@@ -90,15 +39,14 @@ class PhysicalProperty(ArchiveSection):
         Source of the physical property. This quantity is related with the `Activity` performed to obtain the physical
         property. Example: an `ElectronicBandGap` can be obtained from a `'simulation'` or in a `'measurement'`.
         """,
-    )
+    )  # ? useful
```
JFRudzinski: I think we can remove this. I guess this was intended for when PP would be pulled up to a higher level, like DataFrame now. But now the scope of PP is only `simulation`

src/nomad_simulations/schema_packages/physical_property.py
```diff
@@ -107,13 +55,8 @@ class PhysicalProperty(ArchiveSection):
         Label for additional classification of the physical property. Example: an `ElectronicBandGap`
         can be labeled as `'DFT'` or `'GW'` depending on the methodology used to calculate it.
         """,
-        # ! add more examples in the description to improve the understanding of this quantity
-    )
+    )  # TODO: specify use better
```
JFRudzinski, line 67: agree, maybe we can wait and see some natural desired usages of this in our examples and then decide how to clarify

src/nomad_simulations/schema_packages/physical_property.py
```diff
@@ -124,15 +67,7 @@ class PhysicalProperty(ArchiveSection):
...
-    )
-
-    physical_property_ref = Quantity(
-        type=Reference(SectionProxy('PhysicalProperty')),
-        description="""
-        Reference to the `PhysicalProperty` section from which the physical property was derived. If `physical_property_ref`
-        is populated, the quantity `is_derived` is set to True via normalization.
-        """,
-    )
+    )  # TODO: only used for electronic states, remove
```
JFRudzinski, line 79: I agree that this doesn't belong under PP. Are there real use cases of properties referencing individual atoms such that we need to generalize Outputs.modelsystem_ref?

I think if there is a per-atom PP, then Output just references the ModelSystem with the corresponding list of atoms in particles_indices

src/nomad_simulations/schema_packages/physical_property.py
```diff
@@ -146,109 +81,113 @@ class PhysicalProperty(ArchiveSection):
...
+    )
+
     is_scf_converged = Quantity(
         type=bool,
         description="""
         Flag indicating whether the physical property is converged or not after a SCF process. This quantity is connected
         with `SelfConsistency` defined in the `numerical_settings.py` module.
         """,
-    )
+    )  # ? tie to calculation, not individual property
```
JFRudzinski, line 107: I would say yes, bring up to the Output() level. I could see an argument for keeping it here, as there may be specific cases where it is more convenient, but I think we try to limit the contextualization within the PP, and instead keep those in Outputs, and if you need different contexts, you create different Output lists of PPs

src/nomad_simulations/schema_packages/physical_property.py
```diff
@@ -146,109 +81,113 @@ class PhysicalProperty(ArchiveSection):
...
 
     self_consistency_ref = Quantity(
         type=SelfConsistency,
         description="""
         Reference to the `SelfConsistency` section that defines the numerical settings to converge the
         physical property (see numerical_settings.py).
         """,
+    )  # ? remove
```
JFRudzinski, line 115: Not sure about removing, but at least move up to Outputs level? Similar to is_scf_converged

src/nomad_simulations/schema_packages/physical_property.py
```diff
@@ -146,109 +81,113 @@ class PhysicalProperty(ArchiveSection):
...
+
+    contributions = SubSection(
```
JFRudzinski, line 117: I think the reason we did not include contributions in PP before was 
1. To prevent deeper nested structures of sub-classifications for a property -- Can we restrict the usage of this to be 1-level deep?
2. We were a little afraid of enabling a mess of non-standardized or redundant contributions for some property sub-class. I think we have moved beyond this perspective now in favor of more flexibility, but just to clarify

src/nomad_simulations/schema_packages/properties/__init__.py
```diff
@@ -1,14 +1,12 @@
...
 from .energies import (
-    EnergyContribution,
-    FermiLevel,
     KineticEnergy,
     PotentialEnergy,
     TotalEnergy,
 )
 from .fermi_surface import FermiSurface
-from .forces import BaseForce, ForceContribution, TotalForce
+from .forces import TotalForce
```
JFRudzinski, line 9: Have you tested the Force or Energy contributions with the new implementation?

src/nomad_simulations/schema_packages/properties/energies.py
```diff
@@ -78,13 +37,7 @@ class TotalEnergy(BaseEnergy):
...
 
     # ? add a generic contributions quantity to PhysicalProperty
-    contributions = SubSection(sub_section=EnergyContribution.m_def, repeats=True)
-
-    def __init__(
-        self, m_def: 'Section' = None, m_context: 'Context' = None, **kwargs
-    ) -> None:
-        super().__init__(m_def, m_context, **kwargs)
-        self.name = self.m_def.name
+    contributions = SubSection(sub_section=SectionProxy('TotalEnergy'), repeats=True)
```
JFRudzinski: - Shouldn't the contributions be BaseEnergy's?
- How can you directly link the contribution to the Hamiltonian term with this implementation?

src/nomad_simulations/schema_packages/properties/band_gap.py
```diff
@@ -57,12 +57,6 @@ class ElectronicBandGap(PhysicalProperty):
         """,
     )
 
-    def __init__(
-        self, m_def: 'Section' = None, m_context: 'Context' = None, **kwargs
-    ) -> None:
-        super().__init__(m_def, m_context, **kwargs)
-        self.name = self.m_def.name
```
JFRudzinski, line 64: Can you explain your reasoning for removing these?

JFRudzinski (changes requested):
Thanks for your work on this @ndaelman-hu !

I think most of it is good as is. 
- My comments referring to the TODOs  could be addressed simply by coming to a consensus and adding to the TODO
- My main concern is the change in contributions implementation, I don't think it is functioning now in the way we intended. Let's discuss and either adjust it or move it to another PR

src/nomad_simulations/schema_packages/physical_property.py
EBB2675: I have some use cases of properties referencing individual atoms, and referencing through `particle_indices` sounds suitable from my end :thumbsup:

JosePizarro3: It might be pointing to specific orbital states too. 

There are indeed properties which are per atom or per orbital, while there might be a total one as well. Either way, I thought that the `model_system_ref` would be use for the global system, while this one for specific `AtomsState` or `OrbitalsState`.

src/nomad_simulations/schema_packages/properties/energies.py
```diff
@@ -33,42 +23,11 @@ class BaseEnergy(PhysicalProperty):
...
-    relevant atoms or electrons or as a function of them.
-    """
-
-
 ####################################
 # List of specific energy properties
 ####################################
 
 
-class FermiLevel(BaseEnergy):
```
JosePizarro3, line 58: Why `FermiLevel` is removed?

src/nomad_simulations/schema_packages/physical_property.py
ndaelman-hu: You mean "remove the entire quantity"?

src/nomad_simulations/schema_packages/physical_property.py
ndaelman-hu: so I leave this code line be?

src/nomad_simulations/schema_packages/properties/band_gap.py
ndaelman-hu: You mean the initialization? The name assignment has been moved to the parent section, i.e. `PhysicalProperty`.

src/nomad_simulations/schema_packages/properties/energies.py
ndaelman-hu: I would keep the properties for artifacts that are used in searching and data science. The Fermi level doesn't make much sense standalone. It will instead be added to the DOS, band structure, etc. where it manages the alignment.

In the other PR, there's also a derivation of the DOS based on the band structure, in which case the Fermi level may be substituted for the occupation level. I also prefer calling `highest_occupied_state`, as the definition of the Fermi level tends to vary on context.

JosePizarro3: what about chemical potential? That and highest_occupied_state (or perhaps filling?) would be enough

src/nomad_simulations/schema_packages/properties/energies.py
ndaelman-hu: You're right, I was a bit too fast here.
Overall, I'm trying to keep the the contributions the same to the containing class, to avoid complex section structures.

The eventual question, though, is whether we want to define this DSL at the schema-level, or somewhere else.
Since most energy / force decompositions are defined by the program (+ version), we might prefer the parsing level.
We can then mount some kind of standardization on top at a later point.

src/nomad_simulations/schema_packages/physical_property.py
JosePizarro3: If at the end someone has an `Output` section, composed of multiple properties, some might be used during scf, while others are simply extracted. The only way of knowing which is which, would be to read all the `Output` sections, and get the info of the ones that were changing.

If you remove this, then it becomes a bit annoying to get the info. Or this logic could be moved as a method in the class.

EBB2675:
> @EBB2675 I would add the spin base section to a specialized version of `PhysicalProperty`. Would leave this as a short follow-up. 

@ndaelman-hu  Approximately when will you be tackling this part? We will plan some remaining parts of the QC schema accordingly with Denis

src/nomad_simulations/schema_packages/physical_property.py
ndaelman-hu: Isn't this already covered by `PysicalProperty.is_derived`?
The main reason for moving `is_scf_converged` up, is because it will generically have the same value for all output properties under a calculation. Any unexpected edge cases can always be managed by a splitting of the calculation into its subroutines.

JosePizarro3: Well, that's a bit too much: I will simplify it by having two different booleans, but being more specific on the description.

I was defining:
- `is_scf_converged` to easily track which property is changing in each scf step
- `is_derived` to get properties that are derived during normalization
- none defined to get a property directly parsed

src/nomad_simulations/schema_packages/physical_property.py
JFRudzinski: yes, isn't that what you were suggesting with the "useful" comment?

src/nomad_simulations/schema_packages/properties/band_gap.py
JFRudzinski: You mean it assigns the name as "PhysicalProperty" or as the child class name?

src/nomad_simulations/schema_packages/physical_property.py
JFRudzinski: Either leave the comment or address it now as you like

src/nomad_simulations/schema_packages/properties/energies.py
JFRudzinski: I'm sorry, I don't understand what you mean here, perhaps we can discuss in our meeting

src/nomad_simulations/schema_packages/physical_property.py
ndaelman-hu: >  properties which are per atom or per orbital, while there might be a total one as well

Indeed, one can define orbitals at a system-wide level, or projected onto a specific ion, species, etc. This is why `particle_indices` alone don't suffice.
We have a 

I would work towards a bi-directional reference, where electronic entities (like orbitals) are integrated as well.
I was testing this out in the other PR, already.

JosePizarro3: I would develop anything based on how easy is to use, and leverage out logic to methods or references.

From the user perspective, is it easy to get a property and need to code something to read the particle indices to then get which atom refers to? And a similar question for the orbitals?  In my opinion, the current `entity_ref` was the easiest solution. Equivalently, you can also leverage out the logic to resolve this kind of things to a method.

I think we are on the same page @ndaelman-hu, I just wanted to make sure we are all in the same one.

src/nomad_simulations/schema_packages/physical_property.py
ndaelman-hu: Looking at our schema at large, we mostly use labels to define quick catch alls.
That is to say, either you scan several pieces of metadata, or you unite them into a label.
The example above reads more so as `type`. What do you think?

src/nomad_simulations/schema_packages/physical_property.py
ndaelman-hu: The discussion regarding (SCF) convergence has been deferred to its own issue (#207).

ndaelman-hu: > is_scf_converged to easily track which property is changing in each scf step

Is it common for properties not to be updated after a SCF cycle?
I guess it's theoretically possible, but what's the utility?

> is_derived to get properties that are derived during normalization
> none defined to get a property directly parsed

Wouldn't `False` be more appropriate for parsed properties, and `None` for cases were it wasn't tracked?
Your definition holds fine, under the assumption that the tracking system is perfect.

src/nomad_simulations/schema_packages/physical_property.py
ndaelman-hu:
```suggestion
   Short-hand for quickly distinguishing kinds of properties.
   Is typically mapped from some collection of metadata descriptors.
```

tests/properties/test_band_structure.py
```diff
@@ -139,7 +138,7 @@ def test_order_eigenvalues(
         )
         order_result = electronic_eigenvalues.order_eigenvalues()
         if not order_result:
-            assert order_result == result_validation
+            assert not result_validation  # Empty tuple means validation failed
```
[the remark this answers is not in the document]
ndaelman-hu: The type is `tuple`, not `bool`.

src/nomad_simulations/schema_packages/physical_property.py
ndaelman-hu: The discussion regarding (SCF) convergence has been deferred to its own issue (https://github.com/nomad-coe/nomad-simulations/issues/207).

src/nomad_simulations/schema_packages/properties/__init__.py
ndaelman-hu: Yes, though the tests have also been updated.

src/nomad_simulations/schema_packages/properties/energies.py
ndaelman-hu: Modified to `BaseEnergy`.

Semantically, I understand the distinctions in sections to be value judgements on our part. I would relax this to reflect that we just defer to the parsed code's definitions. This is its own discussion, though.

src/nomad_simulations/schema_packages/properties/energies.py
ndaelman-hu: `highest_occupied_state` this is present, yes.
For which kind of methods would you suggest chemical potential?

src/nomad_simulations/schema_packages/physical_property.py
JFRudzinski: I think both type and label are useful, and I am not sure of the precise usage. For me it would be easier to leave a TODO until we implement more examples and then update the description later based on our perspective now. But if you already have an idea, then change the description and also leave the/a TODO

src/nomad_simulations/schema_packages/physical_property.py
JosePizarro3: > Is it common for properties not to be updated after a SCF cycle?
> I guess it's theoretically possible, but what's the utility?

I am thinking that `Outputs` contains a list of physical properties, some were updated in each scf step (as they will appear in the list of `outputs[:]`), some are the output of the whole simulation (perhaps only appearing in the last output, `outputs[-1]`) and internal functions to get properties, and some could be resolved from the normalizers (also in the last `outputs[-1]` element).

> Wouldn't False be more appropriate for parsed properties, and None for cases were it wasn't tracked?
> Your definition holds fine, under the assumption that the tracking system is perfect.

I meant, none of `is_scf_converged` nor `is_derived` are set for those properties.

src/nomad_simulations/schema_packages/properties/energies.py
JosePizarro3: DMFT and, in general, any finite-T method

src/nomad_simulations/schema_packages/physical_property.py
ndaelman-hu: Well, this was my suggestion. There are already some examples in the new schema, but I'm fine building up more.

src/nomad_simulations/schema_packages/physical_property.py
ndaelman-hu: Regarding your point 2, the presence of a quantity or subsection does not require that it gets populated.

src/nomad_simulations/schema_packages/properties/band_gap.py
ndaelman-hu: Good point. `self` ensures that it also sets the children's names correctly.
Added a test to `test_physical_properties.py` to double-check.

src/nomad_simulations/schema_packages/physical_property.py
ndaelman-hu: For point 1, there's a trade-off. Before, you had to define a new `Contribution` section for each child instance of `PhysicalProperty`.
Working from the assumption that most contributions **may** contain similar (meta)data as their containing property section, we'd likely prefer to mirror it by default. This, of course, means that you also mirror `contributions` again. 

I consulted also with Theodore, and we agree that the schema-route for preventing nesting isn't the way to go. While semantically clearer, it would entail significant effort to provide this new _exclusion_ feature.
Instead, we suggest to use normalization for raising an "error" when `contributions` is nested.

ndaelman-hu: Btw, `type` or `label` could be repurposed to distinguish contributions. For now, I'll just introduce a new quantity.

ndaelman-hu: I've added the check at the normalization level, though in principle this should also work for `__setattr__`.
I held off on this approach, as I'm unsure about the overhead it'd create.

src/nomad_simulations/schema_packages/properties/energies.py
ndaelman-hu: Extended the existing `ChemicalPotential`

src/nomad_simulations/schema_packages/physical_property.py
ndaelman-hu: Discussion history logged at #207

src/nomad_simulations/schema_packages/physical_property.py
ndaelman-hu: Maybe just as a summary (also for myself), at the time of writing.

> Either way, I thought that the model_system_ref would be use for the global system, while this one for specific AtomsState or OrbitalsState.

 `ModelSystem` can point to a `ParticleState` or `AtomsState` at the top level alone (unsure whether programmatically enforced). The latter can point to an `OrbitalState`, so the connection is indirect at best.
There atm **no way to define the `OrbitalState`** (or `SpinState`) for a collection of particles / atoms. The latter is based on notes from a discussion with @EBB2675 

> From the user perspective, is it easy to get a property and need to code something to read the particle indices to then get which atom refers to? [...] In my opinion, the current entity_ref was the easiest solution.

I agree that direct references are generally indeed easier, but no comment on the use of `particle_indices`.
I think `entity_ref` is too limited, as it only allows me to reference an orbital / band, but it doesn't define the over which collection of particles / atoms.

I see 2 possible fixes. This tackles the same issue as Spin, so I suggest packing both together in a follow-up issue.

ndaelman-hu:
All tests pass, some points that have been brought up (SCF, referencing species orbitals, spin) are deferred to their own issues.
All other threads have been resolved, as far as I care, but feel free to double-check.

src/nomad_simulations/schema_packages/physical_property.py
```diff
@@ -1,73 +1,26 @@
...
+class PhysicalProperty(PlotSection):
     """
-    A base section used to define the physical properties obtained in a simulation, experiment, or in a post-processing
-    analysis. The main quantity of the `PhysicalProperty` is `value`, whose instantiation has to be overwritten in the derived classes
-    when inheriting from `PhysicalProperty`. It contains `variables`, to define the variables over which the physical property varies (see variables.py).
-    This class can also store several string identifiers and quantities for referencing and establishing the character of a physical property.
+    A base section for computational output properties,
+    containing all (meta)data relevant for visualization.
+    Define and use `value` when the data corresponds with the titular section property.
+    May also contain partitions as subsections.
```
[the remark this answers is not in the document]
ndaelman-hu: Changed the phrasing

```suggestion
    A base section for computational output properties, containing all relevant
    (meta)data. This includes support for visualization and plotting.

    - Supports the definition and use of `value` for the main property data.
    - Allows for the inclusion of contributions (e.g., via the `contribution_type` attribute and
      possible subsections), enabling representation of properties that are composed of multiple
      parts or sources.
    - Inherits from `PlotSection`, enabling direct integration with plotting and visualization tools.
```

src/nomad_simulations/schema_packages/physical_property.py
```diff
@@ -146,109 +86,190 @@ class PhysicalProperty(ArchiveSection):
...
+        self._validate_contributions_structure(logger_arg)
+        self._validate_contribution_type(logger_arg)
+
+        for contribution in self.contributions:
+            if hasattr(contribution, 'normalize'):
+                contribution.normalize(*args, **kwargs)
+
+        if plot_figures := self.plot(**kwargs):
+            self.figures.extend(plot_figures)
+        self.sub_plots(**kwargs)
```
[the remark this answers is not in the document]
ndaelman-hu: `PhysicalProperty.plot`by default returns an empty list, so there's minimal overhead.
Moreover, it is unclear which instance would issue the visualization request within the larger NOMAD framework.

src/nomad_simulations/schema_packages/physical_property.py
```diff
@@ -146,109 +86,190 @@ class PhysicalProperty(ArchiveSection):
...
-    def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
-        self.is_derived = self._is_derived()
+    def sub_plots(self, **kwargs) -> None:
+        """
+        Collects plots from `self.contributions` and overlays them onto the target figure.
+        """
+        if not self.contributions or not self.figures:
+            return
+
+        target_indices = kwargs.get('target_indices', -1)
```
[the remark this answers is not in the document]
ndaelman-hu: resolved

JFRudzinski (changes requested):
See #208

JFRudzinski (approved):

PR #202, opened 2025-07-15
Fix docstrings of ModelSystem

### LINKED ISSUE #198: Clarification on SubSystem specification

EBB2675:
this PR involves a quick fix on docstrings regarding subsystem specification, please see the discussion in #198 

I have also taken into account @JosePizarro3 's comments in #200

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,6 +106,7 @@ ignore = [
     "PLW2901", # redefined-loop-name
     "PLR1714", # consider-using-in
     "PLR5501", # else-if-used
+    "PLC0415", # Import should be at top-level (conflicts with lazy imports)
 ]
 
 fixable = ["ALL"]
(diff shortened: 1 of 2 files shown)


JFRudzinski (approved):
We should probably review these ruff ignore settings at some point to make sure they are consistent with internal standards

JosePizarro3:
Did you try `ruff>=0.11.4`?

EBB2675:
@JosePizarro3 @JFRudzinski 

> Did you try `ruff>=0.11.4`?


I tried this in [a5a93f9](https://github.com/nomad-coe/nomad-simulations/pull/202/commits/a5a93f9df59206d0ded2584ab686e7966158628d), but the lazy import issue persists. I thought its best to merge the ruff ignore settings, e.g. [391e41e](https://github.com/nomad-coe/nomad-simulations/pull/202/commits/391e41e830d1c7866cd217235b51f67af1025433)

please let me know :rocket:

PR #204, opened 2025-07-29
Fix: geometric‑space normalisation for systems without Cells

EBB2675:
### Fix 1
* `GeometricSpace.get_geometric_space_for_atomic_cell()` now obtains the `ase.Atoms` object **exclusively from its parent `ModelSystem`** (`self.m_parent.to_ase_atoms()`), because `Cell`/`AtomicCell` sections never carry atom data themselves.  
* This removes the **`AttributeError: to_ase_atoms`** that previously broke normalisation for molecular or cluster entries.

### Fix 2
* The log message **“No lattice_vectors found in cell[0].”** is **downgraded from `WARNING` → `INFO`** when the system is non‑periodic (`pbc == [False, False, False]`).  
* Periodic entries that truly miss lattice vectors still emit a warning so parser authors notice the problem.

### DIFF
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -192,12 +192,23 @@ class GeometricSpace(Entity):
     def get_geometric_space_for_atomic_cell(self, logger: 'BoundLogger') -> None:
         """
         Get the real space parameters for the atomic cell using ASE.
+        to_ase_atoms live under the parent ModelSystem.
 
         Args:
             logger (BoundLogger): The logger to log messages.
         """
+        parent = getattr(self, 'm_parent', None)
+        if parent is None or not hasattr(parent, 'to_ase_atoms'):
+            logger.warning(
+                'No parent ModelSystem with to_ase_atoms; skipping geometry.'
+            )
+            return
+
+        atoms = parent.to_ase_atoms(logger=logger)
+        if atoms is None:
+            return  # parent already logged the problem
+
         try:
-            atoms = self.to_ase_atoms(logger=logger)
             cell = atoms.get_cell()
             self.length_vector_a, self.length_vector_b, self.length_vector_c = (
                 cell.lengths() * ureg.angstrom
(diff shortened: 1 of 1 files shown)


src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -192,12 +192,23 @@ class GeometricSpace(Entity):
     def get_geometric_space_for_atomic_cell(self, logger: 'BoundLogger') -> None:
         """
         Get the real space parameters for the atomic cell using ASE.
+        to_ase_atoms live under the parent ModelSystem.
 
         Args:
             logger (BoundLogger): The logger to log messages.
         """
+        parent = getattr(self, 'm_parent', None)
```
ladinesa: why use `getattr`? doesn"t self.m_parent always work even for root sections?

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -192,12 +192,23 @@ class GeometricSpace(Entity):
...
+        if parent is None or not hasattr(parent, 'to_ase_atoms'):
```
ladinesa: I suggest not to use `hasattr` try to determine type of parent with `isinstance`

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -192,12 +192,23 @@ class GeometricSpace(Entity):
...
+            logger.warning(
+                'No parent ModelSystem with to_ase_atoms; skipping geometry.'
```
ladinesa: skipping geometry normalization?

ladinesa (approved):

src/nomad_simulations/schema_packages/model_system.py
EBB2675: I used it mostly out of habit. now switching to `parent = self.m_parent` in the next commit

src/nomad_simulations/schema_packages/model_system.py
EBB2675: changed to `if not isinstance(parent, ModelSystem):`

src/nomad_simulations/schema_packages/model_system.py
EBB2675: modified the msg

JosePizarro3 (approved):
lgtm

---

btw, if you are part of the HU Berlin, you can ask for a Github copilot education license. Then, it can make summaries of your issues and even do an initial review on your pr

PR #206, opened 2025-08-05
Extend docs on workflows

ndaelman-hu:
Lay out the main vision for the workflow section within computational sections


JFRudzinski (approved):
Great, thanks Nathan! Just FYI, in the end, I will likely adjust things as I develop also the more general workflow docs on the main docs page, but it's nice to have a starting point, also so we don't forget to address this more in depth later.

JFRudzinski:
Is this done? Let's merge this before next week please

ndaelman-hu:
Extend some of the examples. We can revisit them later on and elaborate more.

Issue #212, opened 2025-08-12
Electronic structure navigation


ndaelman-hu:
Atm, electronic structures, e.g. orbital definitions, can only be accessed via a `PhysicalProperty` or `AtomsState`, which in turn is mainly accessed via the top `ModelSystem`. This doesn't address orbitals defined over the (sub)systems at large.

Given that `ModelSystem` already defines subsystems, I would re-route any electronic property definition via `ModelSystem`.
Electronic structures typically refer to any combination of the 4 principal quantum numbers + their hybridization.
They may thus have their own internal hierarchy that should be reflected. `ModelSystem` should always contain the top level, while properties may reference any level.

Support for `AtomsState` should meanwhile also remain, but rather for defining certain input states. Both can use the same hierarchy schema.

Finally, `Spin`, in particular, comes at various levels of sophistication. There should be a default base section, wherefrom we specialize the various child classes. The base section should be directly contained in the hierarchy at large. This allows for direct use of `Spin` in properties, without having to defer to the hierarchy at large. This last point deals with the reality that most _ab initio_ outputs are split along spin channels.

EBB2675:
Some clarification here might be appropriate:

1. **"Electronic structures = four quantum numbers + hybridization" is not correct.**  
   The four quantum numbers (n, l, ml, ms) describe **atomic** orbitals in a central field.  
   Molecular orbitals and Bloch states usually don’t have those as good quantum numbers. They’re labeled by **MO index/irrep** (molecules) or **(k-point, band index)** (solids).  
   Also, "hybridization" (sp, sp², …) is a basis description.

2. Per-atom vs. system-wide states are being mixed in your above comment.
   `AtomsState` is already appropriate for **atom-centered shells/core levels**.  
   **MOs/bands** are **system/subsystem** objects and should not live under `AtomsState`.

## A preliminary solution sketch (if this `electronic_state` is absolutely needed):

- **Containers**
  - `ModelSystem.electronic_state` (top):
    - `MolecularOrbitalState` (molecules): fields = `mo_index`, energy, occupancy, irrep, `is_spinor`, ⟨S⟩.
    - `BandState` (periodic): fields = `(k_index, band_index)`, energy, occupancy, irrep, `is_spinor`, appropriate spin texture.
  - `AtomsState.electronic_state`:
[...]

PR #215, opened 2025-08-12
Electronic structure navigation

ndaelman-hu:
Aims to structure polymorphic spin sections within a larger electronic hierarchy (laid out #212).


ndaelman-hu:
Thx to @EBB2675 for spotting that the previous branch had old commits in it. Has been resolved by cherry-picking based from `develop`.

Moreover, any failing tests or formatting is handled at the end.

JosePizarro3:
Hey, I always wanted to restructure this part of the schema regarding the quantum numbers. Please, make sure the Python logic works as you intend + the theory behind is sounded (feel free to contribute here: https://github.com/nomad-coe/nomad-simulations/discussions/22)

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -3,38 +3,127 @@
...
+        if self.j_quantum_number is None:
+            return 2  # Default spin degeneracy (spin-up and spin-down)
+            
+        # J-coupling degeneracy calculation
+        degeneracy = 0
+        for jj in self.j_quantum_number:
+            if self.mj_quantum_number is not None:
+                # Specific mj values are defined
+                mjs = RussellSaundersState.generate_MJs(
+                    J=self.j_quantum_number[0], rising=True
```
EBB2675: should this be `J=jj` instead of `J=self.j_quantum_number`

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -1,109 +1,371 @@
...
+        """
+    )  # ? need to define orientation
+
+    # ? particle_index
+
+
+# TODO: add `class CrystalFieldState(BaseSpinOrbitalState)`
+
+
+class ElectronicState(Entity):
```
EBB2675, line 542: Could we change the semantics here from `ElectronicState` to something like `SpinOrbitalState` (or `OneElectronState`)?

In many subfields, electronic state typically refers to a many-electron wavefunction, e.g., a configuration state function or a linear combination of Slater determinants

Here, as I understand it, the current `ElectronicState` section represents a one-electron spin-orbital definition under a given coupling regime (LS, jj, relativistic). Am i right?

To avoid confusion between single-electron and many-electron levels of description, would it be clearer to rename it to `SpinOrbitalState` or a similar term that explicitly signals the one-electron nature of this entity?

ndaelman-hu: > Here, as I understand it, the current ElectronicState section represents a one-electron spin-orbital definition under a given coupling regime (LS, jj, relativistic). Am i right?

Actually, `ElectronicState` is meant to represent the multi-electronic state corresponding to a particular `ModelSystem`. This includes also `AtomsState` and 1-electron states.

But you're right, its `__init__` is too specific top single-electronic systems in a spherical potential. I'll see to move that up, to SphericalSymmetryState`.

ndaelman-hu:
@EBB2675  the tests are not passing, since the old infrastructure still references `OrbitalsState`. I focused instead more on drafting the new layout.

I can see to hook it up with the rest if the schema, if that helps. If we want to change design a lot, then it might be better to hold off, though. Your call.

**Edit:** just to clarify, I've already merge `develop` into this branch, as you requested (commit hash: 1614ac5bb645d538a6e2ed2f167884b1fc976940, follow-up commits are me just cleaning up a bit).

EBB2675:
is the intention following?


ElectronicState               # generic (multi-electron system)
│
├── SphericalSymmetryState    # 1-electron states with V(r)
├── MolecularOrbitalState     # 1-electron states in point-group symmetry
└── BlochOrbitalState         # 1-electron states in periodic potentials

ndaelman-hu:
> is the intention following?
> 
> ElectronicState # generic (multi-electron system) 
> ├── SphericalSymmetryState # 1-electron states with V(r)
> ├── MolecularOrbitalState # 1-electron states in point-group symmetry
> └── BlochOrbitalState # 1-electron states in periodic potentials

Indeed, something along those lines. I'm guessing that `# 1-electron states in periodic potentials` means a list of single-electron states?

Atm, it only contains (the parent class to) `SphericalSymmetryState`, though. Similar to `ModelSystem`, it may have other nested electronic states. Lemme know if it's better to discuss in a meeting.

ndaelman-hu:
As a bit more context to changes up to commit 5e2d51e:

In this refactor, I’ve made `ElectronicState` the central hub for all electronic structure data. It now acts as the main entry point for navigating electronic properties from an input-perspective. From here, we can talk about any set of atoms having an associated quantum state and projections of their properties onto substates.

`ElectronicState` is not necessarily meant to capture the exact numerics of the linear combinations or the populations. While it contains the `basis_orbitals`, this just references the quantum state objects necessary to describe or understand the state. Further quantum numbers describing any level in `ElectronicState` can be added modularly, by inheriting from `BaseSpinOrbitalState`.

By consolidating electronic structure data in this way, we ensure a consistent and extensible interface for downstream analysis, property calculation, and workflow integration. The result is a much cleaner interface: electronic structure is always accessed via `ElectronicState`, and the details of each orbital are encapsulated in quantum state objects referenced within it.

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -1,157 +1,248 @@
...
         """,
     )
 
     l_quantum_symbol = Quantity(
-        type=str,
+        type=MEnum('s', 'p', 'd', 'f'),
+        description="""
+        Symbolic representation of the `l` quantum number, i.e., 's', 'p', 'd', 'f'.
+        """,
+    )  # @EBB2675: do we want to serialize symbolic representations? or maybe just provide a setter / getter mapping?
```
EBB2675: I’d avoid serializing symbolic shell labels and keep the canonical storage purely numeric:

- Store: l: int (0=s, 1=p, 2=d, 3=f, 4=g, …).

- Expose: a derived l_symbol property (getter) and a validating setter that maps symbols ↔ integers.

most codes and libraries reason about l numerically; symbols are pure presentation anyhow

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -1,157 +1,248 @@
...
+        raise NotImplementedError('Subclasses must implement this method.')
+
+    @property
+    def _degeneracy(self) -> int:
+        raise NotImplementedError('Subclasses must implement this method.')
+
+
+class SphericalSymmetryState(
+    BaseSpinOrbitalState
+):  # @EBB2675 we could also split this section into 3 mutually inheriting sections
```
EBB2675, line 53: makes sense for the future but I’d keep this as a single section for now. Once requirements are clearer, we can add symmetry-specific sections (e.g., `CrystalFieldState`) to avoid premature complexity

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -1,157 +1,248 @@
...
+    """Describes a quantum state under spherical symmetry.
+    Supports SOC and relativistic effects."""
+
+    # TODO: define when these quantities are populated and `None` semantics
+
+    kappa_quantum_number = Quantity(
+        type=np.int32, description='κ = ±(j + 1/2), encodes l and j'
+    )  # ? should this be mutually exclusive with j_quantum_number?
```
EBB2675: I’d suggest not making them mutually exclusive. κ and (j, l) encode the same information, so it’s reasonable to accept either; if both are provided, we can check consistency (as your helpers already do) and, if helpful, derive the missing value during normalization

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -235,87 +324,449 @@ def resolve_number_and_symbol(
...
-        Reference to the OrbitalsState section that is used as a basis to obtain the `CoreHole` section.
+        Reference to the orbital state where the core-hole is generated.
+        Can be a single-electron state (SphericalSymmetryState) or a multi-electron state,
+        depending on the level of description needed for the vacancy.
         """,
     )
 
     n_excited_electrons = Quantity(
-        type=np.float64,
+        type=unit_float,  # ? too restrictive
```
EBB2675: I'd keep unit_float here 👍

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -235,87 +324,449 @@ def resolve_number_and_symbol(
...
+    atoms_state_ref = SubSection(
+        section_def=SectionProxy('AtomsState').m_def,
+        description="""
+        Reference to the atomic species (`AtomsState`) to which this electronic state belongs.
+        Populated automatically during normalization to link electronic structure to atoms.
+        """,
+    )
+
+    basis_orbitals = SubSection(
+        section_def=BaseSpinOrbitalState.m_def, # @EBB2675: do you see numerical_settings.basis_set also fit here?
```
EBB2675: Looks good to me, just to confirm my understanding:

we have `basis_orbitals` as semantic labels (AOs/SALCs/Wannier). We can later link to the numerical AO layer via `numerical_basis_ref` and an `ao_indices` array that points into the expanded AO list in `basis_set.AtomicOrbitals`

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -235,87 +324,449 @@ def resolve_number_and_symbol(
...
+        """,
+    )
+
+    def resolve_degeneracy(self) -> int | None:
+        """
+        Resolves the degeneracy of the orbital state using the general formula:
+        total_degeneracy = orbital_degeneracy × spin_degeneracy
+
+        The orbital degeneracy is 2*l + 1 (if ml not specified) or 1 (if ml specified).
+        The spin degeneracy is delegated to the spin_state._degeneracy property, which
```
EBB2675:
```suggestion
        The spin degeneracy is delegated to the spin_orbit_state._degeneracy property, which
```

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -235,87 +324,449 @@ def resolve_number_and_symbol(
...
+        """
+        # Calculate orbital degeneracy
+        if self.l_quantum_number is None:
+            return None
+
+        orbital_degeneracy = (
+            2 * self.l_quantum_number + 1 if self.ml_quantum_number is None else 1
+        )
+        spin_degeneracy = (
+            self.spin_state._degeneracy if self.spin_state is not None else 2
```
EBB2675:
```suggestion
            self.spin_orbit_state._degeneracy if self.spin_orbit_state is not None else 2
```

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -1,157 +1,248 @@
...
+
+    def compute_l_from_kappa(self, kappa: int) -> int:
+        """
+        Compute l quantum number from κ.
+
+        Args:
+            kappa: Relativistic quantum number
+
+        Returns:
+            l = |κ| - 1/2 if κ < 0, or l = κ for κ > 0
```
EBB2675:
```suggestion
            l = |κ| - 1 if κ < 0, or l = κ for κ > 0 (with κ ≠ 0)
```

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -235,87 +324,449 @@ def resolve_number_and_symbol(
             )
         return quantity
 
-    def resolve_degeneracy(self) -> int | None:
+    def _russel_saunders_j_values(
```
EBB2675:
```suggestion
    def _russell_saunders_j_values(
```

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -235,87 +324,449 @@ def resolve_number_and_symbol(
...
+
+        # Remove duplicates and sort
+        return sorted(list(set(all_j_combinations)))
+
+    def __init__(self, m_def: 'Section' = None, m_context: 'Context' = None, **kwargs):
+        super().__init__(m_def, m_context, **kwargs)
+        self._orbitals = {
+            -1: dict(zip(range(4), ('s', 'p', 'd', 'f'))),
+            0: {0: ''},
+            1: dict(zip(range(-1, 2), ('x', 'z', 'y'))),
```
EBB2675: this is a bit unconventional, some packages ( Gaussian/PySCF) use  (x, y, z), some others (Molden/CP2K) use (z, x, y) 

and tying ml = −1,0,+1 directly to real labels (px, py, pz) is itself convention-dependent


do you think it is better not to hard-code the label orders?

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -1,157 +1,248 @@
...
+
+    # TODO: define when these quantities are populated and `None` semantics
+
+    kappa_quantum_number = Quantity(
+        type=np.int32, description='κ = ±(j + 1/2), encodes l and j'
+    )  # ? should this be mutually exclusive with j_quantum_number?
+
+    j_quantum_number = Quantity(
+        type=np.float64,
+        shape=['1..2'],
```
EBB2675: I think the intent behind `shape=['1..2']` is to capture a small manifold (e.g., j=l ± 1/2​). For an actual state, though, j should be a single value. Could we make `j_quantum_number` scalar and compute candidate j values via a helper when needed?

EBB2675 (approved):
Looks great from my side, I added super minor suggestions


I think this came up before, i will open an issue (#269)  for the following:

We’re hard-coding ℓ/mℓ labels and orders (px/py/pz, dxy…) in multiple places in the code. It'd be great if we can centralize this, either in our own utilities or via something external, e.g. `libcint`

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu: Okay, let's keep it for a follow-up then. I'll slightly modify the comment to reflect this and resolve this convo

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu: Sounds good. I'd also prefer only 1 form, but supporting the other via a mapping.
My main aim is querying more so than representation. I don't think we support mapping queries in NOMAD, but would absolutely love that.

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu: I think this got addressed indirectly during other follow-up commits.

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu: okay. given that kappa is quite specific to a relativistic setting, where j and l are still proper quantum numbers. So, populating j and l from kappa makes contextual sense. The other way around, I'm less sure.

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu: Agreed, this would be a good a use case of `ElectronicState` actually.

The question is: how do we deal with a system that defines only the `l_quantum_number`? It would be a system with `self._degeneracy == 2`.
So, in this case, we can just leave it be (my preference), or generate 2 subsections with the individual `j_quantum_numbers`. Any preferences?

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu: @TLCFEM @ladinesa Any ideas how do to a query over a mapping well?
In the old GUI, I guess one would write the mapping in JavaScript. Are there different strategies now?

src/nomad_simulations/schema_packages/atoms_state.py
EBB2675: i was thinking of storing one `ElectronicState` with `j_quantum_number=None` when only `l` is known.

tbh i do not have specific preferences, we can indeed leave it be and see at the heavy parsing stage 👍

src/nomad_simulations/schema_packages/atoms_state.py
EBB2675: i have checked again, the most used convention is z,x,y

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu: > i was thinking of storing one ElectronicState with j_quantum_number=None when only l is known.

That's indeed better, and more aligned with the proposal here. Lemme see to make this extra clear.

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu: will fall back to getters and remove these maps

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu: we'll keep this open for when we write any queries

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu: That's the plan! Since we already have abstract numerical classes, I could also try to add a connection now. Do you have a suggested structure?

ndaelman-hu:
@EBB2675 @JosePizarro3 I further polished the new interface and went thoroughly over its roll-out.
@ladinesa I will soon also push any downstream modifications on the parsers' side.

Feel free to review on your own initiative. Given the time pressure, and that @EBB2675 seemed content with the PR thusfar, I _would like to aim for merging by tomorrow evening_. Pls write me, in case you feel a review is urgent, but does not fall within that timeframe.

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -14,145 +13,301 @@
...
     )
 
-    l_quantum_number = Quantity(
-        type=np.int32,
-        description="""
-        Azimuthal quantum number of the orbital state. This quantity is equivalent to `l_quantum_symbol`.
-        """,
-    )
+    @property
+    def _name(self) -> str:
```
ladinesa, line 38: you can add `@abstractmethod`

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -14,145 +13,301 @@
...
 
-    l_quantum_symbol = Quantity(
-        type=str,
-        description="""
-        Azimuthal quantum symbol of the orbital state, "s", "p", "d", "f", etc. This
-        quantity is equivalent to `l_quantum_number`.
-        """,
-    )
+    @property
+    def _degeneracy(self) -> int:
```
ladinesa, line 43: you normally use property decorator to cache the variable but it appears you did not implement it.

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -14,145 +13,301 @@
...
-        Azimuthal projection of the `j` vector. Necessary with strong L-S coupling or
-        non-collinear spin systems.
+        Angular quantum number of the orbital state. Must be >= 0.
+        Can be set independently when j or s are not known or not applicable.
         """,
     )
 
-    ms_quantum_number = Quantity(
+    @property
+    def l_quantum_symbol(self) -> str:
```
ladinesa, line 90: i am really skeptical about using property decorator for MSections as it somehow confuses the `property` idea (quantities and sub-sections). I would rather have getter methods

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -14,145 +13,301 @@
...
+
+        Returns:
+            True if consistent or only one is defined, False if inconsistent
+        """
+        if self.kappa_quantum_number is not None and self.j_quantum_number is not None:
+            try:
+                expected_j = self.compute_j_from_kappa(self.kappa_quantum_number)
+
+                if abs(self.j_quantum_number - expected_j) >= 1e-10:
+                    logger.error(
```
ladinesa, line 301: Avoid formatted strings in logs

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu: I see what you mean. Could you elaborate a bit more on how you see `@property` fit into NOMAD?

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu: done, also applied to `BaseSpinOrbitalState._degeneracy`

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu: added caching via `@cached_property`
also rolled out to `_name`, in this class and its child classes

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu: applied throughout

ndaelman-hu:
@ladinesa are you fine with me merging this? I applied all but 1 of your suggestions.

ladinesa:
> @ladinesa are you fine with me merging this? I applied all but 1 of your suggestions.

go ahead please

PR #222, opened 2025-08-13
Reassignment of generic particle states

JFRudzinski:
- **add normalization functions for reassigning particle states** from generic to atom or CG: The idea is that we pass the standard determination of atomic vs CG to the normalizer instead of repetitively dealing with it in the parsers or opaquely in a utility
- add _is_atomic prop to ModelSystem, stop normalization early if not atomic
- remove get_chemical_symbols() for generic get_symbols() (previously implemented)
-  add corresponding tests
- comment out is_representative exit of modelsystem normalization, see #225 
- move geom-space atom norm to atomiccell, see #224 
- a few minor fixes or extensions to old code

Notes:
- Passes tests in nomad-parser-plugins-simulation
- Tested on AA and CG h5md data

### DIFF
--- a/src/nomad_simulations/schema_packages/atoms_state.py
+++ b/src/nomad_simulations/schema_packages/atoms_state.py
@@ -1,4 +1,3 @@
-from functools import lru_cache
 from typing import TYPE_CHECKING, Any, Optional, Union
 
 import ase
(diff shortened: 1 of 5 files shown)


EBB2675 (approved):
looks good from my side (also works without troubles in combination with my upcoming huge schema)

i came across a very tiny bug in model system that i wish to solve right after you merge this

JFRudzinski:
@EBB2675 ok great! Then, I will go ahead and merge this. Note I completely removed now the `get_chemical_symbols()` function

@Bernadette-Mohr I still would like you to look through the PR for your own knowledge, and if you have any feedback just put it here and I can address in a subsequent PR

PR #227, opened 2025-08-14
temporary fix: 'molecule / cluster' breaks ModelSystem.type enum

### LINKED ISSUE #226: Class0D → 'molecule / cluster' breaks `ModelSystem.type` enum
`resolve_system_type_and_dimensionality()` assigns 'molecule / cluster' for Class0D, but the current `ModelSystem.type` enums don’t include this value, so `ValueError: ... not a value of this enumeration.`

EBB2675:
here i fix #226, and extend the ModelSystem tests

### DIFF
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -843,11 +843,13 @@ class ModelSystem(System):
     )
 
     # TODO work on improving and extending this quantity and the description
+    # TODO distinguish between molecule and cluster
     type = Quantity(
         type=MEnum(
             'atom',
             'active_atom',
-            'molecule / cluster',
+            'molecule',
+            'molecule / cluster',  # this is kept due to MatID Class0D classification
             'monomer',
             '1D',
             'surface',
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -847,9 +847,8 @@ class ModelSystem(System):
         type=MEnum(
             'atom',
             'active_atom',
-            'molecule',
+            'molecule / cluster',
             'monomer',
-            'cluster',
             '1D',
             'surface',
             '2D',
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -849,6 +849,7 @@ class ModelSystem(System):
             'atom',
             'active_atom',
             'molecule',
+            'cluster',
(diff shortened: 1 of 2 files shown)


src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -847,9 +847,8 @@ class ModelSystem(System):
         type=MEnum(
             'atom',
             'active_atom',
-            'molecule',
+            'molecule / cluster',
```
ladinesa: not sure about the correct terminology here but I feel either you choose molecule or cluster i personally do not like the `or`. Cluster maybe is more general. You can ask joe about this.

EBB2675: same here, but here i just temporarily fix the error that i receive

EBB2675: the terminology stems from MatID `Class0D`

ladinesa (approved):

src/nomad_simulations/schema_packages/model_system.py
JFRudzinski: @EBB2675 actually I changed this in one of my PRs, sorry I did not know this came from MatID or I would have talked to you about it. 

For me a molecule and a cluster are 2 very different things, are there also options to separated these in MatID? Otherwise I think we need to consider expanding the characterization, e.g., having `molecule/cluster`, `molecular`, and `cluster`, or some other solution idk

EBB2675:
```
class Class0D(Classification):
    """Structures that have a structure that is isolated in all directions by a
    vacuum gap.
    """
```

no further classification between molecule and cluster as far as i can see from the source code.

of course this requires a patch that distinguishes between a molecule and cluster, i was just providing a hotfix because my local setup started complaining with this enum errors. anyhow maybe i will just use this branch locally.

JFRudzinski: Can we just allow all 3 options for now? Or do you see that as problematic?

In any case, it's also fine to leave your changing as is, in that case just please add a TODO which is to extend for distinguishing between these

JFRudzinski (approved):

src/nomad_simulations/schema_packages/model_system.py
EBB2675: now all 3 options are allowed under `ModelSystem.type`

i added a TODO

In a seperate PR, i will deal with cluster vs molecule issue 

does that sound good?

Issue #86, opened 2024-06-12
Implement HDF5datasets for potentially large quantities


JFRudzinski:
go through the current schema and propose a set of quantities to be HDF5datasets. I already have in mind pretty much any array that scales with n_atoms, but there are probably other cases as well.

see also #110  -- Green's functions

JFRudzinski:
@ladinesa could you perhaps implement a representative example, e.g., for ModelSystem.positions, which I am sure we should store ad HDF5dataset, so that we could then easily follow for other quantities?

JFRudzinski:
@ladinesa could you perhaps implement a representative example, e.g., for ModelSystem.positions, which I am sure we should store ad HDF5dataset, so that we could then easily follow for other quantities?

ladinesa:
I will define something like a temporary `ModelSystem.positions_h5` and use it in one of the parsers?

JFRudzinski:
sure, perfect, thank you!

ladinesa:
I provided the parsing of charge density in [vasp](https://github.com/FAIRmat-NFDI/nomad-parser-plugins-simulation/pull/35) using the [schema](https://github.com/nomad-coe/nomad-simulations/pull/234)

JFRudzinski:
@EBB2675 This is something you could take: 1. Identify any potentially "large" quantities, 2. redefine these quantities as hdf5datasets in nomad-simulations (see Alvin's example), 3. check corresponding parser tests to make sure no adjustments are necessary (I think the setting of the quantities should not change though if I remember correctly)

PR #234, opened 2025-08-15
Initial charge density def

ladinesa:
Part of #86


ladinesa:
Hi Joe, I need this def for https://github.com/FAIRmat-NFDI/nomad-parser-plugins-simulation/pull/35

PR #238, opened 2025-08-18
Logger decorator

### LINKED ISSUE #217: Apply log decorator util
see https://github.com/FAIRmat-NFDI/nomad-parser-plugins-simulation/blob/develop/src/nomad_simulation_parsers/parsers/utils/general.py#L104 

In lieu of #141

ladinesa:
Closes #217

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ maintainers = [
 ]
 license = { file = "LICENSE" }
 dependencies = [
-    "nomad-lab>=1.3.15",
+    "nomad-lab[infrastructure]>=1.3.15",
     "matid>=2.0.0.dev2",
 ]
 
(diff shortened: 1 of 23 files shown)


ladinesa:
@JFRudzinski this is my implementation for the logger decorator, This can be attached to any function and that should get rid of logger from the function arg. I simply added it to the function annotations instead. Essentially you can define the logger in the function declaration like
```python
@log(logger=LOGGER)
def func(arg):
    pass
```
or when calling the function

```python
@log
def func(arg):
    pass

func(arg, logger=LOGGER)
```
I changed the `to_ase_atoms` method of `ModelSystem` to illustrate this. If you are happy I can do the changes to the other functions.

JFRudzinski (approved):
Sorry for the delay @ladinesa ! I already got very behind on the first day :facepalm: 

Looks great, please go ahead for the other functions :pray:

ladinesa:
This is ready for merging @JFRudzinski  if you can please just give it a quick spin and then we can merge.

Issue #218, opened 2025-08-13
Rename `AtomsState` and `OrbitalsState` for consistency


EBB2675:
- [ ] Rename `AtomsState` → `AtomState` and `OrbitalsState` → `OrbitalState` to align with existing singular sections like `ParticleState` and `CGBeadState`, since these represent a single site/state 
- [ ] Remove conditional imports with type_checking

PR #239, opened 2025-08-18
180 check creation of composition formula

### LINKED ISSUE #180: Check creation of composition formula
The below may no longer be an issue, but anyway the composition formula code needs to be generalized when we implement CG particles

I think there might be 2 issues here: (see [general.py](https://github.com/nomad-coe/nomad-simulations/blob/aeff0677f3f82fe393d40ac3fdf7ff7a57111b21/src/nomad_simulations/schema_packages/general.py#L272))

1. I think there is a misunderstanding of what composition formula is supposed to be. What is implemented here is to get the "normal" chemical formula at each level. However, the composition formula is supposed to be a sort of generalized formula of each parent with respect to its children.
2. This code (which is the appropriate one for the lowest level of the system tree) is only valid for atoms not particles in general.

JFRudzinski:
## Summary of Changes

* **Fix: `Optional` NameError**
  Don’t import `Optional` only inside `TYPE_CHECKING`. Use a runtime import (`from typing import Optional`) or string annotations (e.g., `'Optional[str]'`).

* **`resolve_composition_formula` Overhaul**

  * **Internal nodes:** build composition by counting child `branch_label`s (use `"Unknown"` if missing).
  * **Leaves:** derive composition from root `symbols` sliced by this node’s `particle_indices`; skip if indices/labels are missing.
  * **Honor custom formulas:** never overwrite if `composition_formula` is already set.
  * Implemented via a simple recursive `walk(system_parent)`.

* **`is_representative` Temporarily Ignored for Composition**
  Composition formulas are computed regardless of `is_representative` until scope is finalized. Tests updated to expect formulas even when not representative.

* **Stronger `ModelSystem.symbols`**

  * **Root:** collect labels from `AtomsState` / `CGBeadState` with sensible fallbacks.
  * **Child:** slice from the **ultimate root**; validate indices:

    * must be integers, **non-negative**, and **in range** → otherwise return `[]`.
  * Establishes explicit **“no negative indices”** policy (replacing previous xfail assumption).

* **Bond-List Utilities**
  `get_bond_list(set_local=True)` now **persists** filtered bonds on the child (`child.bond_list`), i.e., order is preserved, and returns an empty `(0, 2)` array when appropriate.

* **Hierarchy Helpers**
  `is_root_system()` / `get_root_system()` use **`m_def` identity checks** (robust to proxies) instead of `isinstance`.

* **ASE Conversions**
  `to_ase_atoms()` **rejects non-element symbols** for atomic systems with clear logging.
  `from_ase_atoms()` populates positions, cell/PBC, and recreates particle states from the ASE object.

## Review Roles

- @ladinesa please provide a technical coding review, I am especially unsure about the dealing with arrays vs. lists and associated typing in the functions where I am slicing root quantities with the subsystem particle_indices

- @EBB2675 please provide a review of the new/adjusted functionalities: see `ModelSystem.resolve_composition_formula()`, `ModelSystem.symbols`, `ModelSystem.is_root_system()`, `ModelSystem.get_root_system()`, `utils.get_composition()`

## Notes

- Passes nomad-parser-plugins-simulation tests


src/nomad_simulations/schema_packages/utils/utils.py
```diff
@@ -1,14 +1,16 @@
...
+from typing import TYPE_CHECKING, Callable, Optional
 
 import numpy as np
 from nomad.config import config
+from nomad.datamodel.data import ArchiveSection
+from structlog.stdlib import BoundLogger
 
-if TYPE_CHECKING:
-    from typing import Callable, Optional
+# if TYPE_CHECKING:
```
JFRudzinski: @ladinesa do we still need this conditional import here? I have the feeling like this was some previous issue that is now solved.

ladinesa: yes you can remove this, i think i dod the same for the parsers repo, unless there is a cyclic import error which i doubt since there are third party packages.

JFRudzinski: ok great, will address it for the other cases in #218

src/nomad_simulations/schema_packages/general.py
```diff
@@ -207,97 +207,57 @@ def _set_system_branch_depth(
             )
 
     def resolve_composition_formula(self, system_parent: ModelSystem) -> None:
-        """Determine and set the composition formula for `system_parent` and all of its
-        descendants.
-
-        Args:
```
ladinesa, line 213: insert rules but do not remove args

src/nomad_simulations/schema_packages/general.py
```diff
@@ -207,97 +207,57 @@ def _set_system_branch_depth(
...
-                        subsystem_labels = ['Unknown'] * len(idx)
-                else:
-                    subsystem_labels = []
-            else:
-                subsystem_labels = [
-                    subsystem.branch_label
-                    if subsystem.branch_label is not None
-                    else 'Unknown'
-                    for subsystem in subsystems
+        def set_for(system: ModelSystem) -> None:
```
ladinesa: i suggest keeping the old name, or more explicit `set_formula`

src/nomad_simulations/schema_packages/general.py
```diff
@@ -207,97 +207,57 @@ def _set_system_branch_depth(
...
+                # INTERNAL NODE: use child branch labels
+                children_names = [
+                    (child.branch_label or 'Unknown') for child in subsystems
                 ]
-            if system.composition_formula is None:
                 system.composition_formula = get_composition(
-                    children_names=subsystem_labels
+                    children_names=children_names
                 )
+                return
```
ladinesa: get rid these return l238, 242 and assign formula at the end, if they can be extracted from the various cases

src/nomad_simulations/schema_packages/general.py
```diff
@@ -207,97 +207,57 @@ def _set_system_branch_depth(
...
+                # No labels available → count Unknown tokens
+                n = len(system.particle_indices)
+                if n > 0:
+                    system.composition_formula = get_composition(
+                        children_names=['Unknown'] * n
+                    )
+
+        def walk(node: ModelSystem) -> None:
+            set_for(node)
+            for child in node.sub_systems or []:
```
ladinesa: sub-sections default to [], no need for `or []`

src/nomad_simulations/schema_packages/general.py
```diff
@@ -207,97 +207,57 @@ def _set_system_branch_depth(
...
+                system.composition_formula = get_composition(children_names=names)
+            else:
+                # No labels available → count Unknown tokens
+                n = len(system.particle_indices)
+                if n > 0:
+                    system.composition_formula = get_composition(
+                        children_names=['Unknown'] * n
+                    )
+
+        def walk(node: ModelSystem) -> None:
```
ladinesa: this is essentially already `self.resolve_composition_formula`, so no need to define walk just call `resolve_composition_formula` iteratively. this should actually be a static method as it appears it does not depend on the object i.e. no `self.` calls in the function.

src/nomad_simulations/schema_packages/general.py
```diff
@@ -320,8 +280,9 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
                 continue
             self._set_system_branch_depth(system_parent=system_parent)
 
-            if is_not_representative(model_system=system_parent, logger=logger):
-                continue
+            # ! Address when we know the role of is_representative
```
ladinesa: add TODO

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1054,28 +1055,64 @@ def _is_atomic(self, value: Optional[bool]) -> None:
     def is_atomic(self) -> bool:
         return self._is_atomic
 
-    # ? Is it better to get symbols with logging or make a property without?
-    def get_symbols(self, logger: 'BoundLogger') -> list[str]:
+    @property
```
ladinesa: for now, let it stay as is, do not use property in MSection as it  technically, all m-quantities are properties. i will add log decorator after

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1054,28 +1055,64 @@ def _is_atomic(self, value: Optional[bool]) -> None:
...
+    def symbols(self) -> list[str]:
```
ladinesa: rename this to what it does as a method, i.e. `get_symbols` to make it consistent with the others

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1054,28 +1055,64 @@ def _is_atomic(self, value: Optional[bool]) -> None:
...
-            if isinstance(particle_state, AtomsState):
-                symbol = particle_state.chemical_symbol
-            elif isinstance(particle_state, CGBeadState):
-                symbol = particle_state.bead_symbol
-            else:
-                symbol = particle_state.label
-            if not symbol:
-                logger.warning('missing symbol in ParticleState.')
+
+        def _root_symbols(ms: 'ModelSystem') -> list[str]:
```
ladinesa: you can remove this method and put everything under a static method get_symbols that take in a model_system. As much as possible vall this only once and any self.symbols ref should ve made to the msection quantity which ia derived from this method.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1054,28 +1055,64 @@ def _is_atomic(self, value: Optional[bool]) -> None:
...
+            if not ms.particle_states:
                 return []
-            symbols.append(symbol)
+            symbols: list[str] = []
+            for particle_state in ms.particle_states:
+                symbol = None
+                if isinstance(particle_state, AtomsState):
```
ladinesa: ia it possible to handle this logic in the PartcleState classes so you do not have the if statements for various types

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1054,28 +1055,64 @@ def _is_atomic(self, value: Optional[bool]) -> None:
...
+                elif isinstance(particle_state, CGBeadState):
+                    symbol = (
+                        particle_state.bead_symbol
+                        if particle_state.bead_symbol
+                        else particle_state.label
+                    )
+                    if not symbol:
+                        alts = particle_state.alt_labels
+                        if alts:
+                            try:
```
ladinesa: the try is a bit unnecessary  symbol = alt_labels(0) if al_labels else None

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1054,28 +1055,64 @@ def _is_atomic(self, value: Optional[bool]) -> None:
...
+                                symbol = alts[0]
+                            except Exception:
+                                symbol = None
+                else:
+                    symbol = particle_state.label
+                if not symbol:
+                    return []
```
ladinesa: maybe put a assign symbols to empty list then break rather than return

ladinesa: maybe put a assign symbols to empty list then break rather than return, several returns are harder to maintain.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1369,17 +1411,41 @@ def is_ge_structure(self, other: 'ModelSystem') -> bool:
     def is_ne_structure(self, other: 'ModelSystem') -> bool:
         return not self.is_equal_structure(other)
 
-    # functions for traversing the ModelSystem hierarchy
-    def get_root_system(self) -> 'ModelSystem':
+    def is_root_system(self) -> bool:
```
ladinesa, line 1398: maybe a one-liner if parent is None or parent
m_def is not ModelSystem.m_def?

ladinesa, line 1398: is the old method not used anymore? ok it follows

src/nomad_simulations/schema_packages/utils/utils.py
```diff
@@ -123,14 +119,48 @@ def is_not_representative(model_system, logger: 'BoundLogger'):
...
-def get_composition(children_names: 'list[str]') -> str:
+# def get_composition(children_names: 'list[str]') -> str:
+#     """
+#     Generates a generalized "chemical formula" based on the provided list `children_names`,
+#     with the format X(m)Y(n) for children_names X and Y of quantities m and n, respectively.
+#     """
+#     children_count_tup = np.unique(children_names, return_counts=True)
+#     formula = ''.join([f'{name}({count})' for name, count in zip(*children_count_tup)])
+#     return formula if formula else None
+def get_composition(children_names: list[str]) -> Optional[str]:
```
ladinesa: is there not an ase method that does this? or maybe try to base it on some ase functions if the children names are themselves formulas

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1097,7 +1134,7 @@ def are_valid_chemical_symbols(self, logger: 'BoundLogger') -> bool:
         Returns:
             bool: True if all chemical symbols are valid, False otherwise.
         """
-        symbols = self.get_symbols(logger)
+        symbols = self.symbols
```
ladinesa: please see my previous if possible do not call the method more than once and rely on the msection quantity derived from the get_symbols method.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1369,17 +1411,41 @@ def is_ge_structure(self, other: 'ModelSystem') -> bool:
...
+        return parent.m_def is not ModelSystem.m_def
+
+    def get_root_system(self) -> 'ModelSystem':
+        """
+        Walk up through parents while the parent behaves like a ModelSystem.
         """
         system = self
-        while isinstance(system.m_parent, ModelSystem):
-            system = system.m_parent
+        while True:
```
ladinesa: use the `is_root_system` above instead to break loop

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1369,17 +1411,41 @@ def is_ge_structure(self, other: 'ModelSystem') -> bool:
...
+                    continue
+            except Exception:
+                pass
+            if parent.m_def is ModelSystem.m_def:
+                system = parent
+                continue
+            break
         return system
 
     # functions for working with molecules
```
ladinesa, line 1421: i think this can also be a static method, remove the set_local arg, setting quantities should be done elsewhere.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1395,27 +1461,29 @@ def get_bond_list(self, set_local: bool = False) -> np.ndarray:
...
-            [
-                (i, j)
-                for i, j in root.bond_list
-                if i in indices_set and j in indices_set
-            ],
-            dtype=np.int32,
+            return empty_return
+
+        idx: NDArray[np.int32]
+        idx = (
```
ladinesa: in line typing idx: np.ndarray, we did have some issues with NDArray? mayb use np.ndarray

src/nomad_simulations/schema_packages/utils/utils.py
```diff
@@ -123,14 +119,48 @@ def is_not_representative(model_system, logger: 'BoundLogger'):
 
 
 # TODO remove function in nomad.atomutils
-def get_composition(children_names: 'list[str]') -> str:
+# def get_composition(children_names: 'list[str]') -> str:
```
ladinesa: remove comments

src/nomad_simulations/schema_packages/utils/utils.py
```diff
@@ -123,14 +119,48 @@ def is_not_representative(model_system, logger: 'BoundLogger'):
...
-    children_count_tup = np.unique(children_names, return_counts=True)
-    formula = ''.join([f'{name}({count})' for name, count in zip(*children_count_tup)])
-    return formula if formula else None
+    if not children_names:
+        return None
+
+    # Ensure string dtype; np.unique returns sorted unique names and counts
+    arr = np.asarray(children_names, dtype=object).astype(str)
+    names, counts = np.unique(arr, return_counts=True)  # case-sensitive
+    # ? switch the above line for case insensitive sorting below?
```
ladinesa: put a TODO if unresolvd.

tests/test_general.py
```diff
@@ -98,15 +98,30 @@ def get_flat_depths(system_parent: ModelSystem, quantity_name: str, value=None):
                 ['group_H20(1)', 'H20(3)', 'H(1)O(2)', 'H(1)O(2)', 'H(1)O(2)'],
                 [None, None, None, None, None],
             ),  # pure system
-            (
+            # (
```
ladinesa, line 101: remove comments

src/nomad_simulations/schema_packages/utils/utils.py
ladinesa: not sure if optional is still a thing or str | None

ladinesa (changes requested):
The get_ functions should be made consistent  and I believe should be static methods. Thee are are simplifications that are necessary. The tests look reasonable although have not checked the coverage.

src/nomad_simulations/schema_packages/utils/utils.py
EBB2675: in `get_composition()`  without sorting the keys the string order may differ across runs. This can make outputs non-deterministic. I don't really know to what extent it might be problematic though. Please see my randomized test:

```
Results over 20 runs:
01: H3 Unknown h Na+2
02: H3 h Na+2 Unknown
03: Unknown Na+2 H3 h
04: Unknown h Na+2 H3
05: H3 h Na+2 Unknown
06: Unknown h H3 Na+2
07: Unknown Na+2 H3 h
08: Unknown Na+2 H3 h
09: H3 Na+2 Unknown h
10: Na+2 Unknown h H3
11: H3 Unknown Na+2 h
12: Na+2 Unknown h H3
13: H3 Unknown Na+2 h
14: H3 Unknown Na+2 h
15: Na+2 H3 Unknown h
16: Na+2 H3 h Unknown
17: Unknown h Na+2 H3
18: H3 h Unknown Na+2
19: h H3 Na+2 Unknown
20: H3 Na+2 h Unknown

```

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1369,17 +1411,41 @@ def is_ge_structure(self, other: 'ModelSystem') -> bool:
...
+            if isinstance(parent, ModelSystem):
+                return False
+        except Exception:
+            pass
 
-        Returns:
-            ModelSystem: The top-level (root) ModelSystem.
+        return parent.m_def is not ModelSystem.m_def
+
+    def get_root_system(self) -> 'ModelSystem':
```
EBB2675, line 1408: would it make sense to add a simple visited-set or max-depth guard for edge cases? 

```
a = ModelSystem(branch_label="A")
b = ModelSystem(branch_label="B", m_parent=a)
c = ModelSystem(branch_label="C", m_parent=b)

# i create a cycle: A points back to C
a.m_parent = c

# aaaand infinite loop
c.get_root_system()
```

src/nomad_simulations/schema_packages/general.py
```diff
@@ -207,97 +207,57 @@ def _set_system_branch_depth(
             )
 
     def resolve_composition_formula(self, system_parent: ModelSystem) -> None:
```
EBB2675: I wish to check my understanding

- **in the case of internal node:** if a node has `sub_systems`, the formula is built via children's `branch_label`.
- **in the case of a leaf node:** if a node has `particle_indices`, then the formula is built from the root

this is great if i have only `sub_systems` or only `particle_indices`

do we need to cover the cases that a node has both? Because right now the function would silently choose one path

src/nomad_simulations/schema_packages/general.py
JFRudzinski: There are not overlapping cases, Leaf nodes are set one way always, internal nodes set the other way always

src/nomad_simulations/schema_packages/general.py
EBB2675: maybe at least we can make a note of the behavior in the docstring:

_If both are present, `sub_systems` take precedence_ or sth like that

src/nomad_simulations/schema_packages/general.py
JFRudzinski: is it better to use @staticmethod or move outside of Simulation()?

ladinesa: these are specific to the class so they should belong to it i suppose.

src/nomad_simulations/schema_packages/general.py
JFRudzinski: There are "Rules" already in the docstring, can you tell me more precisely if these are not clear?

In case of internal nodes (with children) but no branch_labels, should we revert to using the particle labels instead of `Unknown(XXX)`?

src/nomad_simulations/schema_packages/model_system.py
JFRudzinski: Will address in #244

src/nomad_simulations/schema_packages/utils/utils.py
JFRudzinski: This is a generalization of what ase does, I looked at their code but I didn't take away anything useful for now, it depends a little if people like this convention I developed or not

src/nomad_simulations/schema_packages/model_system.py
JFRudzinski: I did improve this I think but I think get_bond_list() should be a class method, like get_symbols(), otherwise the implementation in terms of calling the function feels a bit strange, because I am always doing `ModelSystem.get_bond_list(self)`, please check again

src/nomad_simulations/schema_packages/utils/utils.py
JFRudzinski: I think I fixed this, could you please check again?

src/nomad_simulations/schema_packages/general.py
JFRudzinski: I changed the rules now, to also use particle_labels if all branch_labels are Unknown, please check the clarity / if you agree

JFRudzinski:
The merge with the logger changes was difficult, dealing with the errors now...

src/nomad_simulations/schema_packages/atoms_state.py
```diff
@@ -569,6 +569,12 @@ class ParticleState(Entity):
         """,
     )
 
+    def get_label(self) -> str:
```
ladinesa, line 572: why do we need this wrapper

ladinesa: oh sorry now i get it, but I suggest you make it as an `@abstractmethod` so the children will have to provide this

JFRudzinski: actually, I think it could be needed at the ParticleStates level, perhaps we can discuss this in the system meeting today

src/nomad_simulations/schema_packages/general.py
```diff
@@ -206,98 +206,65 @@ def _set_system_branch_depth(
...
-    def resolve_composition_formula(self, system_parent: ModelSystem) -> None:
-        """Determine and set the composition formula for `system_parent` and all of its
-        descendants.
+    @staticmethod
+    def get_composition_formula(system: ModelSystem) -> str | None:
+        # Honor custom formulas
+        if system.composition_formula is not None:
+            return system.composition_formula
+
+        def get_formula_from_particles(system: ModelSystem) -> str | None:
```
ladinesa: maybe get rid of this function as we simplified the logic already

src/nomad_simulations/schema_packages/model_system.py
ladinesa: yes i agree

EBB2675:
In the `composition_formula` docstring we say:

>     At the deepest branch in the hierarchy, the composition_formula is expressed in terms of the atomic labels.

Right now a root system that is also a leaf (no `sub_systems`) ends up with `composition_formula = None`

Do we populate `composition_formula` _only when_ the root has subsystems? If yes, we should adjust the docstring

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1033,55 +1031,58 @@ class ModelSystem(System):
 
     def __init__(self, m_def: 'Section' = None, m_context: 'Context' = None, **kwargs):
         super().__init__(m_def, m_context, **kwargs)
-        self.__is_atomic_flag = None
+        self._cache: dict[str, Any] = {}
 
-    @property
-    def _is_atomic(self) -> bool:
+    def is_atomic(self) -> bool:
```
ladinesa, line 1041: we probably do not need this wrapper just access thr cache and return False by default

ladinesa (approved):

ladinesa:
From my side you can merge once you resolve remaining minor suggestions.

EBB2675 (approved):

JFRudzinski:
> In the `composition_formula` docstring we say:
> 
> > ```
> > At the deepest branch in the hierarchy, the composition_formula is expressed in terms of the atomic labels.
> > ```
> 
> Right now a root system that is also a leaf (no `sub_systems`) ends up with `composition_formula = None`
> 
> Do we populate `composition_formula` _only when_ the root has subsystems? If yes, we should adjust the docstring

Fixed so that empty root has composition formula, and added tests

src/nomad_simulations/schema_packages/model_system.py
JFRudzinski: sorry, I do not get it, can you send some pseudo-code?

ladinesa: perhaps you can  put l1040-43 directly under normalize? and then you can simply access the cache so you do not worry about this being filled in before particle_states are fliled.

JFRudzinski: hmm, interesting option. I have now changed the logic in is_atomic() so that it is consistent with the reassignment logic. Also, the reassignment criterion itself is simplified:
- always reassign particle_states list that are 1. Contain generic ParticleState instances, 2. Are mixed or inconsistent with our is_atomic() logic

I think in this way I have also solved the issue. However, we should still reassess whether what I have done now is good practice. I leave that to #248 as I would like to move on to some other tasks now

PR #245, opened 2025-08-20
Revert "Extend workflows"

ladinesa:
Reverts nomad-coe/nomad-simulations#243


JFRudzinski (approved):
what's the situation here, why reverting?

ladinesa:
I rebased it to the logger decorator just to pass the test but what I should be doing is first merge the logger decorator. I can simply also rebase the logger decorator to the merged develop but this will make the commit history messed up. So purely technical. sorry

Issue #242, opened 2025-08-19
Move Wyckoff to positions


ndaelman-hu:
So, the Wyckoff positions are high-symmetry points in the _conventional unit cell_. This should be clearly stated in the descriptions. This is important, as not every crystal structure is reported in this structure. In the NOMAD front-end, we do, in fact, report the conventional unit cell in the topology hierarchy whenever possible.

Any Wyckoff analysis provides 2 kinds of information:

1. Generic tabulation of the positions of the high-symmetry points. This is fully determined by the space group (type of symmetry for crystals), and the setting.
2. Labeling of which atomic positions correspond to which symmetry positions.

I'm unsure whether we'll provide any search functionalities for Wyckoff, but it is super-useful for inferring the effects of a site on all kinds of observables.

ndaelman-hu:
> In the NOMAD front-end, we do, in fact, report the conventional unit cell in the topology hierarchy whenever possible.

We could already do this in the `data` section, adding a new `model_system` in conventional units. This may be useful for other kinds of analyses down the line.

I would have the Wyckoff sites mimic the atomic positions, and add further detail about each type of site to a repeating subsection.
MatID uses the Wyckoff sites and atomic number to assign unique locations too. Given that the information (and more, like atomic settings) is all at the same level, I would remove this quantity.

EBB2675:
> We could already do this in the `data` section, adding a new `model_system` in conventional units. 


what does "a new `model_system`" mean?

EBB2675:
is there anything i can do to help solve this issue?

ndaelman-hu:
> is there anything i can do to help solve this issue?

No, thanks. Wyckoff has already been migrated. It's matching PR was closed in favor of another one, which has been merged for a few months now.

Merged solution: https://github.com/FAIRmat-NFDI/nomad-simulations/pull/304

Issue #240, opened 2025-08-19
Disambiguate crystal symmetry symbols


ndaelman-hu:
Depending on the choice of origin / setting, the H-M and Hall symbols will change. Hall records the changes that were applied, H-M doesn't. Technically, one could produce a list of equivalent symbols, especially for H-M.

There are standard (short) representations for H-M and Hall, which is what most people will use when trying to search for a crystal at large. We should also decide whether rotations and shifts should be applied to show the crystal in its standard setting.

PR #247, opened 2025-08-20
242 move wyckoff to positions

ndaelman-hu:
See issue #242


ndaelman-hu:
Main modifications:

- elaborate on Wyckoff description, explain what it represents (defined wrt conventional cell) and its limits (geometry specific).
- extend Wyckoff to include site multiplicity.
- relax unique site definition to include computational settings. Remove the matching MatID descriptor.
- move Wyckoff to `Modelsystem` to mimic positions (which annotates). Alternative location would be `Symmetry`, which makes semantically more sense. The issue is how to distinguish between multiple `Symmetry` subsections.
- start up `ModelSystem` docs to keep track of design choices.

Open questions:

- should we also provide the site symmetry operators with Wyckoff. These symmetry operators then need a definition.
- ratify `cell` and `symmetry`. `Symmetry` seems set up to further define `Cell`, but the connection is indirect (reference at `Symmetry.cell_ref`). It also isn't clear which `ModelSystem.cell` defines the current positions, or how to rapidly retrieve the original cell (atm requires scanning).

ndaelman-hu:
One option for `Cell` and `Symmetry` is have 1 contain the other. This only requires 1 repeating subsection then, where by definition index 0 refers to the `ModelSystem` as described.

ndaelman-hu:
@ladinesa Feel free to comment on the use of the `logs` decorator and whether I can be more aggressive in removing logging messages.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1477,3 +1484,33 @@ def is_molecule(self) -> bool:
                     return False
 
         return True
+
+    @log
+    def _compute_wyckoff_sites(self) -> list[str]:
```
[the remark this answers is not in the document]
ladinesa: yes this is correct

ladinesa:
> @ladinesa Feel free to comment on the use of the `logs` decorator and whether I can be more aggressive in removing logging messages.

maybe do it in a separate pr.

ndaelman-hu:
> Main modifications:
> 
>     * elaborate on Wyckoff description, explain what it represents (defined wrt conventional cell) and its limits (geometry specific).
> 
>     * extend Wyckoff to include site multiplicity.
> 
>     * relax unique site definition to include computational settings. Remove the matching MatID descriptor.
> 
>     * move Wyckoff to `Modelsystem` to mimic positions (which annotates). Alternative location would be `Symmetry`, which makes semantically more sense. The issue is how to distinguish between multiple `Symmetry` subsections.
> 
>     * start up `ModelSystem` docs to keep track of design choices.
> 
> 
> Open questions:
> 
>     * should we also provide the site symmetry operators with Wyckoff. These symmetry operators then need a definition.
> 
>     * ratify `cell` and `symmetry`. `Symmetry` seems set up to further define `Cell`, but the connection is indirect (reference at `Symmetry.cell_ref`). It also isn't clear which `ModelSystem.cell` defines the current positions, or how to rapidly retrieve the original cell (atm requires scanning).

From the sprint meeting, we decided to have a single `cell` define the `model_system` representation. Likewise, there will only be one `CyrstalGlobalSymmetry`, defined wrt to the standard conventional cell.
@lauri-codes is it fine that for the conventional cell representation we keep leaning on the topology normalizer and the viewer?

JFRudzinski (approved):
@ndaelman-hu Unfortunately I don't have time for an in-depth review, but at a glance it looks good to me, and we discussed quite a bit about it, so I feel like we are on the same page.

The docs I completely did not read, but I will do a review anyway at a later time. So, from my side this is fine to merge, I hope someone else can give you a more detailed review.

One question: should we make Cell() non-repeating, or was there still some debate about this?

ndaelman-hu:
> @ndaelman-hu Unfortunately I don't have time for an in-depth review, but at a glance it looks good to me, and we discussed quite a bit about it, so I feel like we are on the same page.
> 
> The docs I completely did not read, but I will do a review anyway at a later time. So, from my side this is fine to merge, I hope someone else can give you a more detailed review.

Thx, I'll give the other reviewers some more time.

> One question: should we make Cell() non-repeating, or was there still some debate about this?

Yes, it is addressed in issue #240. Will soon have a PR up there too.

docs/model_system/model_system.md
```diff
@@ -1,4 +1,209 @@
...
+
+2. **Primitive Cell** (`type='primitive'`)  
+   - Smallest possible unit cell that tiles the crystal
+   - Contains minimum number of lattice points
+   - Used for certain calculations (e.g., band structures)
+
+3. **Conventional Cell** (`type='conventional'`)
+   - Standardized cell following International Tables conventions
+   - Used for symmetry analysis and crystallographic databases
+   - **Typically used to define symmetry operators:** ["x,y,z", "-x,y,-z", "x+1/2,y+1/2,z", "-x+1/2,y+1/2,-z"] (conventional) vs ["x,y,z", "-x,-y,z", "-x,y,-z", "x,-y,-z"] (primitive).
```
lauri-codes, line 75: It should be maybe noted here that primitive and conventional systems may not directly correspond to the original cell: they will have different densities and "idealized" lattice parameters to match space group that they were matched with certain precision.

docs/model_system/model_system.md
```diff
@@ -1,4 +1,209 @@
...
+## Particle Positions: An Open Question
+
+Currently, `ModelSystem` stores particle positions at the top level (`ModelSystem.positions`), but these positions correspond to only **one** of the multiple cell representations. This creates ambiguity about which cell the positions refer to.
+
+```python
+model_system = ModelSystem()
+model_system.positions = [[0.0, 0.0, 0.0], [0.25, 0.25, 0.25]]  # Which cell?
+model_system.cell[0] = Cell(type='original', lattice_vectors=...)    # These positions?
+model_system.cell[1] = Cell(type='primitive', lattice_vectors=...)   # Or these positions?
+model_system.cell[2] = Cell(type='conventional', lattice_vectors=...) # Or these positions?
```
lauri-codes, line 97: From a practical perspective we need the atomic positions for each (original, primitive, conventional) cell in order to visualize them in the GUI. If you anyways run the symmetry analysis to fill the `symmetry` section, then I would say it probably makes sense to store also the atomic positions of the primitive cell/ conventional cell: otherwise we need to run the symmetry analysis again from scratch in the normalizer.

docs/model_system/model_system.md
```diff
@@ -1,4 +1,209 @@
...
+
+**Key features:**
+  - Format: <letter><multiplicity> (e.g., 'a1', 'b2')
+  - Crystallographic standard: Based on International Tables for Crystallography
+  - Geometric classification: Determined by atomic positions only (chemical species ignored)
+  - Per-atom assignment: Each entry corresponds to the position at the same index
+
+  Important: Wyckoff assignments are computed relative to the conventional cell representation, ensuring database compatibility and literature consistency.
+
+  For complete structural uniqueness, combine Wyckoff information with chemical composition data.
```
lauri-codes, line 203: I would suggest that wyckoff sites would have their own model that holds the letter, the multiplicity + any wyckoff  parameters.

Also worth noting that also the primitive and conventional cells might need their own Wyckoff sites if we e.g. want to visualize them. But maybe this is not necessary.

docs/model_system/model_system.md
ndaelman-hu: I will add this as a clarification to the description.

> they will have different densities

Not sure what you mean with densities here: mass, electronic? The former stays the same, the latter changes exactly like the geometry. However, k-samplings should change too, and this is not acknowledged (or addressed) anywhere.

docs/model_system/model_system.md
ndaelman-hu: Replied to in PR #253 . Let's continue any discussion there pls.

docs/model_system/model_system.md
ndaelman-hu: So you're suggesting splitting the format here? Is this for convention reasons, or maybe querying?

With "Wyckoff parameters", do you mean of the analysis itself, or the symmetry operations at each site?
The representation of these operations is indeed cell-dependent. The locations should simply match the geometric transformations.

ndaelman-hu:
@lauri-codes There was some confusion about whether the Wyckoff analysis only considers the positions, or also accounts for the different atomic species at these positions. Can you confirm that it is in fact the latter?

While including elements into the analysis is very sensible, it raises the question whether we should also consider species with _different computational settings_? There are cases where these settings still _represent the same element_, or those that were the settings model a _new physical reality_, e.g. core-hole excitations.

We could also run multiple Wyckoff analyses with each possible setting (positions only, chemical elements, computational settings).

lauri-codes:
Yes, Wyckoff analysis (and symmetry analysis in general) does take the species identity into account. You can yourself explicitly remove the species information of course before the analysis, but this completely changes the symmetries (e.g. rock salt becomes simple cubic).

I guess it is difficult to draw the line when a certain particle has been altered in it's behaviour so much that it has become a different "species". I would go with some pragmatic approach here that does not completely break the searchability of data based on symmetry.

docs/model_system/model_system.md
lauri-codes: With wyckoff parameters, I mean that some Wyckoff positions are parametrized with up to three parameters (x, y, z). Now whether  it is important to know these parameters or not is another question. MatID can give out this information [if you explicitly ask](https://nomad-coe.github.io/matid/docs/reference/symmetryanalyzer/#get_wyckoff_sets_conventional), see and example here: https://nomad-coe.github.io/matid/docs/learn/symmetry-analysis#wyckoff-sets.

If you think that just having the letter+multiplicity is enough, then having a list of strings should be fine. But this structure cannot be easily extended later, unlike a list of objects (list of objects does have other problems, like the fact that in ES you require nested queries to combine information from several fields).

ndaelman-hu:
1. Either we make a shortlist of physically relevant distinctions. Should just see whether we can pass non-existing element labels to MatID.
2. We produce all 3 versions: geometry, elements, computation. This is the most comprehensive approach, but also 3x as expensive.
3. We keep the original approach of just elements.

If feasible, I would go with option 1.

docs/model_system/model_system.md
ndaelman-hu: I think the main application of Wyckoff would be in visualization, yes.
I can only imagine somebody looking for an element (or with a slab, an adsorbate) being present at a specific site. This would be a very sophisticated query, though. I think users would have to be taught about it.

docs/model_system/model_system.md
ndaelman-hu: Ow, upon re-read, i think I understand it now. You're referring to deviations in the cell volume due to the symmetrization accuracy constraint.

docs/model_system/model_system.md
ndaelman-hu: Regarding the queries, I'll follow your suggestion up: I think we would want to support queries like

- find me materials where the representative position `['1/4', '0', '1/2']` that has multiplicity 6.
- find me materials (of symmetry ...) where `Fe` occupies a `b8` site. This might be more complex under our current setup.

ndaelman-hu: > find me materials (of symmetry ...) where `Fe` occupies a `b8` site. This might be more complex under our current setup.

To my understanding, this kind of query would have to be nested in ES. That is to say, we need an object like `[{element: Xx, wyckoff_letter: x, wyckoff_number: x, ...}, ...]`, similar to the proposal above. A flat array mirroring `positions` as I had originally, is likely nicer for visualization: no need to scan and flatten the object. Considering compatibility with the current schema: 

- **high:** the flat object and global symmetry descriptors.
- **low**: Wyckoff objects less so. They'd need particle data for the nested query, but they are only appropriate for crystalline system. An intermediate solution would be to have the `particle_indices` twice, but we've been trying to avoid double data, as it tends to confuse users.

If we want allow for nested queries in `data` (and that's a big IF), it might be nicer to have a specialized section, e.g. `data/query`, where we bundle any relevant data to power nested queries. This would clearly denote that this section solves a purely technical problem, as well as suggest a potential (more sophisticated) queries. The core schema can meanwhile grow in a more intuitive way for scientists and domain experts.

This is were my discussion with @JFRudzinski arrived at. @lauri-codes We would really appreciate your opinion here! The suggestion here could act as a template for future cases.

JFRudzinski: Thanks for the summary @ndaelman-hu 

I will just clarify that `data/query` or `results/query` are equally viable, or perhaps even a different structure, this is a matter of design to be dictated by you @lauri-codes

ndaelman-hu:
Closed in favor of #304

PR #250, opened 2025-08-21
removed is_not_representative() util, set default representative system

### LINKED ISSUE #225: Decide exact usage of is_representative
modelsystem.is_representative was previously being used to skip normalization, but this is not necessarily desired for all normalization, and it remains rather unclear what is_representative is or should be used for

JFRudzinski:


src/nomad_simulations/schema_packages/general.py
```diff
@@ -265,16 +262,13 @@ def set_composition_formula(system_parent: ModelSystem) -> None:
...
-        # define it as the last system reported (TODO CHECK THIS!).
         # TODO extend adding the proper representative system extraction using `normalizer.py`
         if not self.model_system:
             logger.error('No system information reported.')
             return
-        system_ref = self.model_system[-1]
-        # * We define is_representative in the parser
-        # system_ref.is_representative = True
-        self.m_cache['system_ref'] = system_ref
+        # Set the representative system to last in list if not already set
```
ndaelman-hu: Pls add this specification to the pydocs too. We can later add it to the docs.
I suggest adding an `representative_system_index` to facilitate retrieval of the representative system.

ndaelman-hu (approved):
This is just a simplification of the code. Perhaps this logic was originally centralized with the idea of further developing it? As is, it is definitely redundant.

Still, if possible, I'd use this as an opportunity to better define `is_representative`.
For example, I suppose there can only be 1 such system at any given point?
Moreover, it cannot be a subsystem, I assume?
Finally, in a serial calculation is should correspond with final update, regardless of convergence. What about parallel cases? Should it only exist in `data` or also in `workflow2` (I guess not).

EBB2675 (approved):
Looks good from my side.

One quick question, i guess Nathan also pointed this out: what happens if multiple systems are flagged as representative? Right now normalization allows/resolves this silently. Should we enforce uniqueness or at least warn?

JFRudzinski:
I really am not sure what the specification should be, this is exactly why I was asking about it yesterday. 

I think we can enforce a single representative system now and expand later if we need to. 

@ndaelman-hu @EBB2675 please have another quick look since I expanded quite a bit now

EBB2675 (approved):

ndaelman-hu:
> I really am not sure what the specification should be, this is exactly why I was asking about it yesterday.
> 
> I think we can enforce a single representative system now and expand later if we need to.
> 
> @ndaelman-hu @EBB2675 please have another quick look since I expanded quite a bit now

I think the exclusion of subsystems too?

JFRudzinski:
> > I really am not sure what the specification should be, this is exactly why I was asking about it yesterday.
> > I think we can enforce a single representative system now and expand later if we need to.
> > @ndaelman-hu @EBB2675 please have another quick look since I expanded quite a bit now
> 
> I think the exclusion of subsystems too?

yes, thank you, I forgot

src/nomad_simulations/schema_packages/general.py
```diff
@@ -262,19 +266,84 @@ def set_composition_formula(system_parent: ModelSystem) -> None:
...
+        None
+        """
+        logger = self._validate_and_set_representative_system.__annotations__['logger']
+
+        # indices of representative systems
+        rep_idx = [i for i, ms in enumerate(self.model_system) if ms.is_representative]
+
+        if len(rep_idx) == 0:
+            self.model_system[-1].is_representative = True
+            self.representative_system_index = -1
```
ndaelman-hu, line 318: This will break when `self.model_system` is `None`. In case of an empty list, it will point to a non-existing position. I think both can be handled without relying on `@log` here.

src/nomad_simulations/schema_packages/general.py
```diff
@@ -262,19 +266,84 @@ def set_composition_formula(system_parent: ModelSystem) -> None:
...
+        elif len(rep_idx) > 1:
+            logger.warning(
+                'Multiple representative systems found, one allowed.'
+                ' Will use the last one found.'
```
ndaelman-hu:
```suggestion
                ' Will use the last `ModelSystem`.'
```

The "last one found" sounds like it refers to `rep_idx`.

src/nomad_simulations/schema_packages/general.py
```diff
@@ -262,19 +266,84 @@ def set_composition_formula(system_parent: ModelSystem) -> None:
...
+            )
+            for idx in rep_idx:
+                self.model_system[idx].is_representative = False
+            self.model_system[rep_idx[-1]].is_representative = True
+            self.representative_system_index = rep_idx[-1]
+        else:
+            self.representative_system_index = rep_idx[0]
```
ndaelman-hu: maybe put a comment saying that is the `== 1` case, as it's easy to misread at first glance

src/nomad_simulations/schema_packages/general.py
```diff
@@ -262,19 +266,84 @@ def set_composition_formula(system_parent: ModelSystem) -> None:
...
+        and update `self.representative_system_index` to point to it.
+
+        Behavior
+        --------
+        - **Zero representatives:** Mark the *last* item in `self.model_system` as
+        representative and set `self.representative_system_index = -1`
+        (Python’s negative index for the last element).
+        - **Multiple representatives:** Log a warning, demote all, then keep the
+        *last* representative encountered. Store its (non-negative) index in
+        `self.representative_system_index`.
```
ndaelman-hu, line 282: This works under the presumption that the model_systems follow an order...
Maybe put it as a disclaimer? Either at the documentation level or the programmtic level, we should enforce this.

src/nomad_simulations/schema_packages/general.py
```diff
@@ -262,19 +266,84 @@ def set_composition_formula(system_parent: ModelSystem) -> None:
...
+            self.representative_system_index = rep_idx[-1]
+        else:
+            self.representative_system_index = rep_idx[0]
+
     def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
+        """
+
+        Normalize the `Simulation` section:
+        - Validate and set the representative system.
+            - Certain normalization steps are only applied to the representative system to avoid redundancy.
```
ndaelman-hu, line 337: TODO: enumerate later on

src/nomad_simulations/schema_packages/general.py
```diff
@@ -262,19 +266,84 @@ def set_composition_formula(system_parent: ModelSystem) -> None:
...
+            - The representative system is then used to power some downstream tools.
```
ndaelman-hu:
```suggestion
            - The representative system then denotes the jump-in point for the user.
```

JFRudzinski:
@ndaelman-hu final check?

ndaelman-hu (commented):
Still ongoing, just intermediate remarks

src/nomad_simulations/schema_packages/general.py
ndaelman-hu: Ow sorry, the check is handled below in the normalizer. Given the completeness of this method, may consider moving it up? I leave it up to you.

src/nomad_simulations/schema_packages/general.py
```diff
@@ -262,19 +266,84 @@ def set_composition_formula(system_parent: ModelSystem) -> None:
...
-        system_ref = self.model_system[-1]
-        # * We define is_representative in the parser
-        # system_ref.is_representative = True
-        self.m_cache['system_ref'] = system_ref
+        # Validate that there is exactly one representative system and set the index
+        self._validate_and_set_representative_system()
+
+        # Set the representative system to last in list if not already set
+        if all(not system.is_representative for system in self.model_system):
+            self.model_system[-1].is_representative = True
```
ndaelman-hu: Isn't this handled by `_validate_and_set_representative_system` already?

tests/test_general.py
ndaelman-hu, line 1: Minor point, but with the exception `test_idempotent`, I think that these can be rolled into 1 with `pytest.mark.parametrize`. You'd just need a setup function.

tests/utils/test_utils.py
ndaelman-hu, line 1: Again, consider `pytest.mark.parametrize` here.

ndaelman-hu (commented):
done

src/nomad_simulations/schema_packages/general.py
JFRudzinski: I think it needs to be in the normalizer or else the rest of the normalization will try to run after the error, and then you will have explicit run time errors, right?

src/nomad_simulations/schema_packages/general.py
JFRudzinski: it does, it's the last rep_idx found, but changed to: 

```python
            logger.warning(
                'Multiple representative systems found, one allowed.'
                ' Will use the last `ModelSystem` found.'
            )
```

src/nomad_simulations/schema_packages/general.py
JFRudzinski:
```python
        """

        Normalize the `Simulation` section:
        - Validate and set the representative system.
            - Certain normalization steps are only applied to the representative system to avoid redundancy.
            - The representative system is used to power some downstream tools.
            - The representative system should be the primary source of information for the user.

        Args:
            archive (EntryArchive): _description_
            logger (BoundLogger): _description_
        """
``` 


Better?

src/nomad_simulations/schema_packages/general.py
JFRudzinski: Simulation.model_systems is a list, so it has an order. I don't understand why it would be unclear, can you explain please?

tests/test_general.py
JFRudzinski: please check then I will remove old tests

tests/utils/test_utils.py
JFRudzinski: same as above

src/nomad_simulations/schema_packages/general.py
ndaelman-hu: Very much possible, hence "I leave it up to you".
Maybe just add to the description that the error handling is done up.

src/nomad_simulations/schema_packages/general.py
ndaelman-hu: > Simulation.model_systems is a list, so it has an order.

Lists aren't necessarily ordered, e.g. `[2, 3, 1]`. We presume that when a natural ordering (let's say iteration steps or time) is present, it is applied to `model_systems`, but this assumption is always implicit and not enforced. This is a good opportunity for making it explicit, showing why it's important.

tests/test_general.py
```diff
@@ -314,3 +314,126 @@ def test_root_leaf_custom_composition_preserved(self):
...
+        assert [ms.is_representative for ms in sim.model_system] == expected_flags
+
+        # Second pass: flags must be unchanged; index becomes canonical (non-negative)
+        sim._validate_and_set_representative_system()
+        assert [ms.is_representative for ms in sim.model_system] == expected_flags
+
+        true_idx = next(
+            i for i, ms in enumerate(sim.model_system) if ms.is_representative
+        )
+        assert sim.representative_system_index == true_idx
```
ndaelman-hu: So this also tests idem-potency. Looks good, just weird that it creates `true_index` for the 2nd test.
Couldn't it just run the same code block twice, or set up an explicit comparison?

tests/test_general.py
```diff
@@ -314,3 +314,126 @@ def test_root_leaf_custom_composition_preserved(self):
...
+            # 1) zero representatives → promote last, index -1 on first pass
+            ([False, False, False], -1, [False, False, True]),
+            # 2) multiple representatives → keep the last
+            ([True, False, True, False], 2, [False, False, True, False]),
+            # 3) exactly one representative → preserve it
+            ([False, True, False], 1, [False, True, False]),
+            # 4) singleton without representative → promote itself, index -1 on first pass
+            ([False], -1, [True]),
+        ],
+    )
```
ndaelman-hu: Much nicer. If you use `normalize` instead of `_validate_and_set_representative_system`, you can also check empty or unassigned `model_system`.

tests/utils/test_utils.py
```diff
@@ -76,69 +148,41 @@ def example_func2(a):
...
-    )
-    assert (
-        get_sibling_section(sibling_section, 'cell', logger, index_sibling=0).type
-        == section.type
-    )
-    assert (
-        get_sibling_section(sibling_section, 'cell', logger, index_sibling=1).type
-        == section2.type
-    )
+    # --- second call (idempotence) ---
```
ndaelman-hu: The logs won't be idempotent wrt execution. So, what are you enforcing idempotency on, the return value of the method? That fully depends on the method then.

ndaelman-hu (approved):
Looks much better. A few minor remarks, I leave them up to you to resolve, no need for another review.

tests/test_general.py
ndaelman-hu: See comments below

tests/utils/test_utils.py
ndaelman-hu: see comments below

PR #251, opened 2025-08-21
System Tree Integrity

### LINKED ISSUE #244: Enforce system tree integrity
would it make sense to add a simple visited-set or max-depth guard for edge cases? 

```
a = ModelSystem(branch_label="A")
b = ModelSystem(branch_label="B", m_parent=a)
c = ModelSystem(branch_label="C", m_parent=b)

# i create a cycle: A points back to C
a.m_parent = c

# aaaand infinite loop
c.get_root_system()
```

_Originally posted by @EBB2675 in https://github.com/nomad-coe/nomad-simulations/pull/239#discussion_r2284357284_

## Plan

During normalization check that each child.particle_indices is a subset of parent.particle_indices

JFRudzinski:
added validation of subsystems & check for cycles, improved get_root, system() functionality:

- previously get_root_system() was relying on no parent or non-ModelSystem parent, such that it could be in an infinite loop --- now check for cycles
- add get_root_system() to each system_parent in Simulation.model_systems to check for cycles
- added validation of subsystem particle indices in ModelSystem.normalize()

### DIFF
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -1340,7 +1340,7 @@ class ModelSystem(System):
 
         if self.particle_indices is None:
             logger.warning(
-                f'Cannot validate ModelSystem subsystem without particle_indices.'
+                'Cannot validate ModelSystem subsystem without particle_indices.'
             )
             return
 
@@ -1352,24 +1352,24 @@ class ModelSystem(System):
             n_particles = len(particle_array) if particle_array is not None else None
             if not n_particles:
                 logger.error(
-                    f'Cannot validate ModelSystem subsystem without root particle info.'
+                    'Cannot validate ModelSystem subsystem without root particle info.'
                 )
                 return
 
-            assert all(
-                i >= 0 and i < n_particles for i in self.particle_indices
-            ), 'Invalid particle_indices in ModelSystem subsystem.'
+            assert all(i >= 0 and i < n_particles for i in self.particle_indices), (
+                'Invalid particle_indices in ModelSystem subsystem.'
+            )
             return
 
         if parent.particle_indices is None:
             logger.error(
-                f'Cannot validate ModelSystem subsystem without parent particle_indices.'
+                'Cannot validate ModelSystem subsystem without parent particle_indices.'
             )
             return
 
-        assert all(
-            pi in parent.particle_indices for pi in self.particle_indices
-        ), 'Invalid particle_indices in ModelSystem subsystem.'
+        assert all(pi in parent.particle_indices for pi in self.particle_indices), (
+            'Invalid particle_indices in ModelSystem subsystem.'
+        )
 
         # TODO logger.warning or logger.error in each case?
 
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -1355,9 +1355,9 @@ class ModelSystem(System):
                 )
                 return
 
-            assert all(
-                0 <= i < n_particles for i in self.particle_indices
-            ), 'Invalid particle_indices in ModelSystem subsystem.'
+            assert all(0 <= i < n_particles for i in self.particle_indices), (
+                'Invalid particle_indices in ModelSystem subsystem.'
+            )
             return
 
         if parent.particle_indices is None:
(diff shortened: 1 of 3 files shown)


src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1332,11 +1332,54 @@ def _reassign_generic_particle_states(self, archive, logger) -> None:
...
+        if self.particle_indices is None:
+            logger.warning(
+                'Cannot validate ModelSystem subsystem without particle_indices.'
+            )
+            return
+
+        parent = self.m_parent
+        if parent.is_root_system():
+            particle_array = (
+                parent.positions or parent.particle_states or parent.velocities
```
EBB2675: `parent.positions` is an array. Wouldnt the truthiness will raise the `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`?

code to reproduce:
```
def test_validate_subsystem_handles_numpy_positions_without_truthiness_error():
    parent = ModelSystem(is_representative=True)
    parent.positions = np.zeros((4, 3))  
    child = ModelSystem()
    child.particle_indices = [0, 1]
    child.m_parent = parent 

    # incoming ValueError here
    child._validate_subsystem(logger)
```


 maybe we shoul use `is not None` (or explicit length) instead?

ladinesa: is this possible that you have positions or velocities but no particle states? should you just check particle_states?

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1332,11 +1332,54 @@ def _reassign_generic_particle_states(self, archive, logger) -> None:
...
+            )
+            n_particles = len(particle_array) if particle_array is not None else None
+            if not n_particles:
+                logger.error(
+                    'Cannot validate ModelSystem subsystem without root particle info.'
+                )
+                return
+
+            assert all(i >= 0 and i < n_particles for i in self.particle_indices), (
```
ladinesa: why is this check necessary and i think it is wrong since you are checking simply the length of the parent array, you can have an index which goes beyond that. the later check index in parent indices suffices i think. or am i misunderstanding it

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1332,11 +1332,54 @@ def _reassign_generic_particle_states(self, archive, logger) -> None:
...
+            )
+            return
+
+        if parent.particle_indices is None:
+            logger.error(
+                'Cannot validate ModelSystem subsystem without parent particle_indices.'
+            )
+            return
+
+        assert all(pi in parent.particle_indices for pi in self.particle_indices), (
```
ladinesa, line 1369: you can perhaps enforce a strictly positive particle_indices earlier

src/nomad_simulations/schema_packages/model_system.py
JFRudzinski: my first thought was that the positions should be the main quantity, one could want to store some configuration of positions without labeling them explicitly. Of course our parsers would not do that, and maybe it would break all the normalization anyway so it is moot.

I used particle_states for now

@EBB2675 what do you think?

src/nomad_simulations/schema_packages/model_system.py
JFRudzinski: This is the special case that parent is root, in which case there are no particle_indices, so we have to check that the indices are valid.

For children lower in the tree, we don't need to check if they are negative or too large, just if they are a subset of their parent, then by transitivity they are within the correct range relative to root

src/nomad_simulations/schema_packages/model_system.py
JFRudzinski: see explanation above, I guess if there are negative indices throughout it's slightly more efficient to move the check above, then multiple errors will be thrown instead of one, but I think it's unnecessary. Let me know though if you still think I am wrong...of course very possible

src/nomad_simulations/schema_packages/model_system.py
JFRudzinski: ok, I just simplify to checking the length of particle_states now

src/nomad_simulations/schema_packages/model_system.py
ladinesa: What i mean by earlier is before you do the validation, only for the root, check that particle indices are all positive. nevermind, I get it now.

src/nomad_simulations/schema_packages/model_system.py
ladinesa: more pythonic 0 <= i < n_particles.

ladinesa (approved):

PR #253, opened 2025-08-22
240 disambiguate crystal symmetry symbols

ndaelman-hu:
- Introduce base class `GlobalSymmetry` and rename `Symmetry` to `GlobalCrystalSymmetry`.
- Step away from defining standardized cells (_primitive_ and _conventional_) to just annotating the cell and symmetry for the system that is being represented. This removes complexity in the normalization and the navigation between the repeating subsections `ModelSystem.cell` and `ModelSystem.symmetry`.
- Make subsections `ModelSystem.cell` and `ModelSystem.symmetry` non-repeating.
- Heavily expand on the description of the crystal symmetry descriptors.
   - Clarify that all are defined in terms of the conventional unit cell.
   - Add lower-dimensional lattices too.

### DIFF
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -1267,9 +1267,9 @@ class ModelSystem(System):
 
         ase_atoms = ase.Atoms(symbols=symbols)
 
-        # Use cell data (from the first cell) for periodic boundary conditions and lattice
-        if self.cell and len(self.cell) > 0:
-            cell_section = self.cell[0]
+        # Use cell data for periodic boundary conditions and lattice
+        if self.cell:
+            cell_section = self.cell
             if cell_section.periodic_boundary_conditions is None:
                 logger.info(
                     'Cell periodic_boundary_conditions not found; using default [False, False, False].'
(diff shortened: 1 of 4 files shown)


ndaelman-hu:
@lauri-codes Do you see any issues with these changes? I understand that the previous implementation mimics the current `results` section, but I think that the topology normalizer already does enough constructing the conventional cell.

lauri-codes:
Having one section for the symmetry properties makes sense: this is also how the current results section deals with it, and this makes queries easier.

Having only one cell is not a big deal if one of the normalizers calculates the other representations. But like I mentioned in the other PR, if you anyways populate the `symmetry` section, you will basically automatically also have the primitive and conventional cells available, and the normalizers needs to re-run the symmetry analysis from scratch.

ndaelman-hu:
> Having one section for the symmetry properties makes sense: this is also how the current results section deals with it, and this makes queries easier.
> 
> Having only one cell is not a big deal if one of the normalizers calculates the other representations. But like I mentioned in the other PR, if you anyways populate the `symmetry` section, you will basically automatically also have the primitive and conventional cells available, and the normalizers needs to re-run the symmetry analysis from scratch.

Good point regarding double work. To my understanding, atm `topology` does indeed run the symmetry analysis again.
I know that the vision has been to move everything to the simulation schema, then copy over relevant parts to `results`.

If we want MatID to also cover molecules and 1D, as well as magnetic symmetry, then I guess we have some flexibility at which stage we perform the symmetry analysis. If not, we proceed with the "copying" vision.
Some of the recent changes to our schema, though, make it less obvious where new representation (and positions) would be stored. I guess this is something to be figured out with the rest of Area C ( @JFRudzinski ).

ndaelman-hu:
Closed in favor of #304

PR #255, opened 2025-09-01
re-implemented isinstance() preference check in is_root()

JFRudzinski:
fixes https://github.com/FAIRmat-NFDI/nomad-parser-plugins-simulation/actions/runs/17328941645/job/49200085013

### DIFF
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -1440,11 +1440,15 @@ class ModelSystem(System):
 
     def is_root_system(self) -> bool:
         """
-        A node is root if its parent is not a ModelSystem section.
-        Prefer isinstance for robustness; fall back to m_def identity.
+        True if this node has no parent or its parent is not a ModelSystem.
+        Prefer an ``isinstance`` check; fall back to comparing ``m_def`` to handle
+        proxy/wrapper parents that still expose a ModelSystem definition.
         """
 
-        return (self.m_parent is None) or (self.m_parent.m_def is not ModelSystem.m_def)
+        parent = self.m_parent
+        return (parent is None) or not (
+            isinstance(parent, ModelSystem) or parent.m_def is ModelSystem.m_def
+        )
 
     def get_root_system(self) -> 'ModelSystem':
         """


JFRudzinski:
@ladinesa I had removed the is_instance() check at your request, but I think the m_def check is insufficient when you have a system hierarchy, perhaps because of proxy instances. 

Let me know if you have some other preferred implementation :pray:

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1440,11 +1440,15 @@ def is_ne_structure(self, other: 'ModelSystem') -> bool:
...
-        Prefer isinstance for robustness; fall back to m_def identity.
+        True if this node has no parent or its parent is not a ModelSystem.
+        Prefer an ``isinstance`` check; fall back to comparing ``m_def`` to handle
+        proxy/wrapper parents that still expose a ModelSystem definition.
         """
 
-        return (self.m_parent is None) or (self.m_parent.m_def is not ModelSystem.m_def)
+        parent = self.m_parent
+        return (parent is None) or not (
+            isinstance(parent, ModelSystem) or parent.m_def is ModelSystem.m_def
```
ladinesa, line 1450: does the `is` work when comparing objects or `==` is preferred?

ladinesa (approved):

src/nomad_simulations/schema_packages/model_system.py
ladinesa: `instance(parent, ModelSystem)` is not anymore necessary?

JFRudzinski: ChatGPT says:

```
Short answer: prefer is, not ==, when comparing m_def.

m_def is effectively a singleton “definition object” for a section type. You want to know if two references point to the exact same definition, not whether they’re equal by some custom __eq__. Using == can:

invoke arbitrary equality semantics (or even fail) if __eq__ is defined,

be slower or ambiguous,

return True for distinct-but-similar definitions.
```

Do you think it's correct? Or should I change to `==`? 

In any case I cannot remove the `isinstance()` or else I get back is_root() true for all sub-systems

ladinesa: what is know is `is` if you really compare objects in memory, in general objects should be compared with `==`, anyway this is just a minor thing. indeed `is` is less error prone as we do not have to go through the `__eq__` check, but metainfo will not allow defining this so they are basically the same. For point 2, sorry yes indeed you are right, it will return True for everything other than ModelSystem

PR #259, opened 2025-09-04
atom centered basis set and related metadata definitions

EBB2675:
Here is the support for `AtomCenteredFunction`, `AtomCenteredBasisSet` and co.

As you know, i follow the TREXIO format:
https://trex-coe.github.io/trexio/trex.html

Meanwhile, Denis will also take a look at it from a scientific perspective.

### DIFF
--- a/src/nomad_simulations/schema_packages/basis_set.py
+++ b/src/nomad_simulations/schema_packages/basis_set.py
@@ -428,16 +428,18 @@ class AtomicOrbitals(ArchiveSection):
         population analysis, MO coefficients, or density matrices.
     """
 
-    num = Quantity(type=np.int32, description='Total number of AOs (contracted).')
+    n_atomic_orbitals = Quantity(
+        type=np.int32, description='Total number of AOs (contracted).'
+    )
     cartesian = Quantity(type=bool, description='True=Cartesian, False=Spherical.')
     shell_index = Quantity(
         type=np.int32,
-        shape=['num'],
+        shape=['n_atomic_orbitals'],
         description="For each AO: index of AtomCenteredFunction (the 'shell') it belongs to.",
     )
     normalization = Quantity(
         type=np.float64,
-        shape=['num'],
+        shape=['n_atomic_orbitals'],
         description='Per-AO factor to reconcile code conventions.',
     )
 
@@ -557,7 +559,7 @@ class AtomCenteredBasisSet(BasisSetComponent):
         """,
     )
 
-    total_number_of_basis_functions = Quantity(
+    n_total_basis_functions = Quantity(
         type=np.int32,
         description="""
         The **total** number of contracted basis functions in this entire set. 
@@ -588,7 +590,7 @@ class AtomCenteredBasisSet(BasisSetComponent):
 
         # Set AO count from AO view if present
         if self.atomic_orbitals is not None and self.atomic_orbitals.num is not None:
-            self.total_number_of_basis_functions = self.atomic_orbitals.num
+            self.n_total_basis_functions = self.atomic_orbitals.num
 
             if (
                 self.functional_compositions is not None
--- a/src/nomad_simulations/schema_packages/basis_set.py
+++ b/src/nomad_simulations/schema_packages/basis_set.py
@@ -452,10 +452,17 @@ class AtomicOrbitals(ArchiveSection):
         shape=['n_atomic_orbitals'],
         description="For each AO: index of AtomCenteredFunction (the 'shell') it belongs to.",
     )
+
     normalization = Quantity(
         type=np.float64,
         shape=['n_atomic_orbitals'],
-        description='Per-AO factor to reconcile code conventions.',
+        description="""
+            Unitless normalization factor for each atomic orbital after shell expansion.
+            It may include additional scaling applied during AO orthogonalization or
+            transformation steps reported by the code.
+            Distinct from shell_normalization, which normalizes contracted functions
+            before expansion into individual AOs.
+            """,
     )
 
 
--- a/src/nomad_simulations/schema_packages/basis_set.py
+++ b/src/nomad_simulations/schema_packages/basis_set.py
@@ -188,7 +188,7 @@ class AtomCenteredFunction(ArchiveSection):
 
     In many quantum-chemistry codes, an atom-centered basis set is composed of
     several "shells," each shell containing one or more basis functions of a certain
-    angular momentum. For instance, a shell of p-type orbitals (L=1) typically
+    angular momentum. For instance, a shell of p-type orbitals (ℓ=1) typically
     consists of 3 degenerate functions (p_x, p_y, p_z) if `harmonic_type='cartesian'`
     or 3 spherical harmonics if `harmonic_type='spherical'`.
 
@@ -197,10 +197,6 @@ class AtomCenteredFunction(ArchiveSection):
     In practice, these contract together to form the final basis function used by
     the SCF or post-SCF method. Often, each contraction is labeled by its
     angular momentum (e.g., s, p, d, f) and a set of exponents and coefficients.
-
-    This class also accepts *combined* shells ('sp', 'spd', 'spdf') as input.
-    On normalize(), it **expands** them into multiple single-ℓ shells and
-    replaces itself with those shells in the parent's `functional_compositions`.
 
     **References**:
       - T. Helgaker, P. Jørgensen, J. Olsen, *Molecular Electronic-Structure Theory*, Wiley (2000).
@@ -449,8 +445,7 @@ class AtomicOrbitals(ArchiveSection):
 class AtomCenteredBasisSet(BasisSetComponent):
     """
     Defines an **atom-centered basis set** for quantum chemistry calculations.
-    Unlike plane-wave methods, these expansions are typically built around each atom's
-    position, using either:
+    These expansions are typically built around each atom's position, using either:
       - Slater-type orbitals (STO)
       - Gaussian-type orbitals (GTO)
       - Numerical atomic orbitals (NAO)
--- a/src/nomad_simulations/schema_packages/basis_set.py
+++ b/src/nomad_simulations/schema_packages/basis_set.py
@@ -296,86 +296,79 @@ class AtomCenteredFunction(ArchiveSection):
     def _is_combined(self) -> bool:
         return self.function_type in self._COMBINED_ORDER
 
-    def _expand_combined_into_parent(self) -> None:
-        """
-        Expand a combined shell into multiple single-ℓ shells by rewriting this
-        instance in place to the first component and appending the rest to the
-        parent's functional_compositions. Quiet no-op if parent/data missing.
-        """
+    def _letters(self) -> list[str]:
+        return self._COMBINED_ORDER.get(self.function_type, [])
+
+    def _expected_combined_len(self, n_primitive: int, letters: list[str]) -> int:
+        return int(n_primitive) * len(letters)
+
+    def _expand_combined_into_parent(self, logger: 'BoundLogger') -> None:
         parent = self.m_parent
-        if not isinstance(parent, AtomCenteredFunction):
-            return
-        if (
-            self.n_primitive is None
-            or self.exponents is None
-            or self.contraction_coefficients is None
-        ):
+        if not isinstance(parent, AtomCenteredBasisSet):
             return
 
-        letters = self._COMBINED_ORDER[self.function_type]
-        step = int(self.n_primitive)
-        flat = np.asarray(self.contraction_coefficients, dtype=np.float32).ravel()
-        expected = step * len(letters)
+        letters = self._letters()
+        step = int(self.n_primitive) if self.n_primitive is not None else 0
+        flat = None
+        if self.contraction_coefficients is not None:
+            flat = np.asarray(self.contraction_coefficients, dtype=np.float32).ravel()
+            if self.n_primitive is not None:
+                expected = self._expected_combined_len(step, letters)
+                if flat.size != expected:
+                    logger.error(
+                        'Combined shell coefficients length mismatch: got %d, expected %d '
+                        '(function_type=%s, n_primitive=%s).',
+                        int(flat.size),
+                        int(expected),
+                        self.function_type,
+                        str(self.n_primitive),
+                    )
+                    # continue expansion, but coeff slices may be incomplete
 
-        # check expected length consistency, bail out if mismatch
-        if flat.size != expected:
-            logger.error('Combined shell coefficients length mismatch.')
+        for n, ltr in enumerate(letters):
+            coeff_slice = None
+            if flat is not None and step > 0 and flat.size >= (n + 1) * step:
+                coeff_slice = flat[n * step : (n + 1) * step]
 
-        # ---- rewrite THIS instance to the first component ----
-        first = letters[0]
-        first_slice = flat[0:step]
-        self.function_type = first
-        self.angular_momentum = self._L_MAP.get(first)
-        # keep n_primitive, exponents, primitive_factor, normalizations as-is
-        self.contraction_coefficients = np.array(first_slice, dtype=np.float32)
-        # n_components should now be single-ℓ
-        self.n_components = 1
+            if n == 0:
+                shell = self
+            else:
+                shell = AtomCenteredFunction()
+                parent.functional_compositions.append(shell)
 
-        # ---- append the remaining components as brand-new shells ----
-        for i, ltr in enumerate(letters[1:], start=1):
-            coeff_slice = flat[i * step : (i + 1) * step]
-            new_shell = AtomCenteredFunction(
-                angular_type=self.angular_type,
-                function_type=ltr,
-                angular_momentum=self._L_MAP.get(ltr),
-                r_power=self.r_power,
-                shell_normalization=self.shell_normalization,
-                n_primitive=self.n_primitive,
-                exponents=np.array(self.exponents, copy=True),
-                contraction_coefficients=np.array(coeff_slice, copy=True),
-                primitive_factor=(
-                    np.array(self.primitive_factor, copy=True)
-                    if self.primitive_factor is not None
-                    else None
-                ),
-                primitive_normalization=self.primitive_normalization,
-                point_charge=self.point_charge,
+            shell.angular_type = self.angular_type
+            shell.function_type = ltr
+            shell.angular_momentum = self._L_MAP.get(ltr)
+            shell.r_power = self.r_power
+            shell.shell_normalization = self.shell_normalization
+            shell.n_primitive = self.n_primitive
+            shell.exponents = np.array(self.exponents, copy=True)
+            shell.contraction_coefficients = (
+                np.array(coeff_slice, copy=True) if coeff_slice is not None else None
             )
-            parent.functional_compositions.append(new_shell)
+            shell.primitive_factor = (
+                np.array(self.primitive_factor, copy=True)
+                if self.primitive_factor is not None
+                else None
+            )
+            shell.primitive_normalization = self.primitive_normalization
+            shell.point_charge = self.point_charge
 
     @check_normalized
     def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
         super().normalize(archive, logger)
 
-        # components implied by function_type
-        self.n_components = (
-            len(self._COMBINED_ORDER[self.function_type]) if self._is_combined() else 1
-        )
-
-        # expand combined shells and exit
+        # expand combined shells and stop
         if self._is_combined():
-            self._expand_combined_into_parent()
+            self._expand_combined_into_parent(logger)
             return
 
-        # single-ℓ mapping: set angular_momentum from function_type if missing
         if self.function_type in self._L_MAP and self.angular_momentum is None:
             self.angular_momentum = self._L_MAP[self.function_type]
 
-        # infer n_primitive
         if self.exponents is not None and self.n_primitive is None:
             self.n_primitive = len(self.exponents)
 
-        # consistency check
         if self.n_primitive is not None:
             if self.exponents is not None and len(self.exponents) != self.n_primitive:
                 self.m_set(self.m_def.all_quantities['exponents'], None)
@@ -672,6 +665,11 @@ class AtomCenteredBasisSet(BasisSetComponent):
     @check_normalized
     def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
         super().normalize(archive, logger)
+
+        # # Ensure child shells are normalized first so combined shells split
+        # if self.functional_compositions:
+        #     for func in list(self.functional_compositions):
+        #         func.normalize(archive, logger)
 
         # Set AO count from AO view if present
         if self.atomic_orbitals is not None and self.atomic_orbitals.num is not None:
--- a/src/nomad_simulations/schema_packages/basis_set.py
+++ b/src/nomad_simulations/schema_packages/basis_set.py
@@ -428,10 +428,15 @@ class AtomicOrbitals(ArchiveSection):
         population analysis, MO coefficients, or density matrices.
     """
 
+    type = Quantity(
+        type=MEnum('spherical', 'cartesian'),
+        description='Angular representation of this atomic orbital.',
+    )
+
     n_atomic_orbitals = Quantity(
         type=np.int32, description='Total number of AOs (contracted).'
     )
-    cartesian = Quantity(type=bool, description='True=Cartesian, False=Spherical.')
+
     shell_index = Quantity(
         type=np.int32,
         shape=['n_atomic_orbitals'],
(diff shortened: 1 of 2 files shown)


src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,521 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    )
+
+    # optional embedded point charge
+    point_charge = Quantity(
+        type=np.float32, description='Optional embedded point charge.'
+    )
+
+    # helper (internal)
+    n_components = Quantity(
+        type=np.int32, description='Number of components implied by function_type.'
```
JFRudzinski: If this is only a helper then should it simply be a private class variable and not a proper quantity, since it seems that after normalization it will simply be = 1?

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,521 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        super().normalize(archive, logger)
+
+        # components implied by function_type
+        self.n_components = (
+            len(self._COMBINED_ORDER[self.function_type]) if self._is_combined() else 1
+        )
+
+        # expand combined shells and exit
+        if self._is_combined():
+            self._expand_combined_into_parent()
```
JFRudzinski: I don't fully understand what is going on here, but wondering if you need to guard against duplicate expansions in case of future refactors, e.g., is this covered in the tests?

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,521 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        | CP2K       | (pz, px, py)    | (p₀, p₊₁, p₋₁)              |
+
+        If you select `'Custom'`, specify the explicit order for each ℓ ≥ 1
+        in the optional field `ao_custom_order` below.  Otherwise parsers and
+        writers will assume the canonical order embedded in the named
+        convention.
+        """,
+    )
+
+    ao_custom_order = Quantity(
```
JFRudzinski, line 582: have you considered `type=JSON`?

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,521 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    replaces itself with those shells in the parent's `functional_compositions`.
+
+    **References**:
+      - T. Helgaker, P. Jørgensen, J. Olsen, *Molecular Electronic-Structure Theory*, Wiley (2000).
+      - F. Jensen, *Introduction to Computational Chemistry*, 2nd ed., Wiley (2007).
+      - J. B. Foresman, Æ. Frisch, *Exploring Chemistry with Electronic Structure Methods*, Gaussian Inc.
     """
 
-    pass
+    harmonic_type = Quantity(
```
JFRudzinski: is "cartesian harmonic" correct terminology? Maybe rename "angular_type" or "angular_representation". Enum types could be expanded to `spherical_harmonics` / `cartesian_polynomials` (or whatever is accurate terminology) if necessary to be clear

JFRudzinski (approved):
:1st_place_medal:

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,521 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    def _is_combined(self) -> bool:
+        return self.function_type in self._COMBINED_ORDER
+
+    def _expand_combined_into_parent(self) -> None:
+        """
+        Expand a combined shell into multiple single-ℓ shells by rewriting this
+        instance in place to the first component and appending the rest to the
+        parent's functional_compositions. Quiet no-op if parent/data missing.
+        """
+        parent = getattr(self, 'm_parent', None)
```
ladinesa: do not use getattr parent = self.m_parent

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,521 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        if parent is None or not hasattr(parent, 'functional_compositions'):
```
ladinesa:
```
if not isinstance(parent, AtomCenteredBasisFunctions):
    return
```

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,521 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+            return
+
+        letters = self._COMBINED_ORDER[self.function_type]
+        step = int(self.n_primitive)
+        flat = np.asarray(self.contraction_coefficients, dtype=np.float32).ravel()
+        expected = step * len(letters)
+
+        # check expected length consistency, bail out if mismatch
+        if flat.size != expected:
+            return
```
ladinesa: log error.

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,521 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+
+        # ---- rewrite THIS instance to the first component ----
```
ladinesa: maybe try to put this in the loop then say shell = AtomCenteredfunction() if n else self

ladinesa (approved):

src/nomad_simulations/schema_packages/basis_set.py
EBB2675: done

src/nomad_simulations/schema_packages/basis_set.py
EBB2675: i have removed expansion entirely

EBB2675:
Hi again @ladinesa @JFRudzinski 

I've pushed the final commits for the atom-centered basis. The PR is ready for final quick review.

- The main changes resolve the complexity around combined shells:  I removed the ambiguous combined shell types ('sp', 'spd', etc.) and all associated expansion/deletion logic. The schema now only accepts single-ℓ shells, shifting the expansion task to the parser (which is cleaner).

- I also eliminated all remaining unnecessary `getattr` calls.

- I anticipate applying more granular consistency checks and physical constraints in the future via normalization as I implement more parsers. But this shouldn’t stop us from merging this I think.

- Of course, I added a lot of tests, e.g. a mixed-atom centered setup where different basis sets were assigned to different atomic species and so on.

JFRudzinski (approved):
Looks good to me :+1:

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
 
 class AtomCenteredFunction(ArchiveSection):
     """
-    Specifies a single function (term) in an atom-centered basis set.
+    Specifies a single contracted basis function in an atom-centered basis set.
+
+    In many quantum-chemistry codes, an atom-centered basis set is composed of
+    several "shells," each shell containing one or more basis functions of a certain
+    angular momentum. For instance, a shell of p-type orbitals (L=1) typically
```
JosePizarro3: ...(ℓ=1)...

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    consists of 3 degenerate functions (p_x, p_y, p_z) if `harmonic_type='cartesian'`
+    or 3 spherical harmonics if `harmonic_type='spherical'`.
+
+    A single "atom-centered function" can be a linear combination of multiple
+    primitive Gaussians (or Slater-type orbitals, STOs).
+    In practice, these contract together to form the final basis function used by
+    the SCF or post-SCF method. Often, each contraction is labeled by its
```
JosePizarro3, line 198: Mm I am a bit confused about what the angular momentum ℓ then represents. I guess "contraction" means here "degeneracy", i.e., that for same values of ℓ there are multiple mℓ?

Just to clarify for me.

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+            'i',
+            'j',
+            'k',
+            'l',
+        ),
+        description='Angular-momentum label (s, p, d, f, etc.).',
+    )
+
+    # explicit single-ℓ metadata
+    angular_momentum = Quantity(
```
JosePizarro3, line 230: `function_type` and `angular_momentum` are redundant (right?). If so, can you mention this in the description?

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        description='Angular momentum quantum number ℓ.',
+    )
+
+    r_power = Quantity(
+        type=np.int32,
+        description="Radial power n_s for this shell's analytic form (typically 0 for GTOs).",
+    )
+
+    # per-shell normalization (contracted)
+    shell_normalization = Quantity(
```
JosePizarro3, line 241: is this without units? not sure from the description what is the normalization here (does it mean degeneracy-related normalization?)

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        type=np.float64, description='Per-shell normalization factor.'
+    )
+
+    # primitives
+    n_primitive = Quantity(
+        type=np.int32, description='Number of primitives in this shell.'
```
JosePizarro3: maybe mention again briefly what a primitive is. Either here or in `exponents`

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    )
+
+    exponents = Quantity(
```
JosePizarro3, line 260: I would keep naming here with `_primitive` to make clear this is related with the expansion in gaussians/whatever

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    )
+
+    contraction_coefficients = Quantity(
+        type=np.float32,
+        shape=['n_primitive'],
+        description='Contraction coefficients for the primitives in this single-ℓ shell.',
+    )
 
-    # TODO: design system for writing basis functions like gaussian or slater orbitals
+    primitive_factor = Quantity(
```
JosePizarro3, line 270: you can keep consistent naming with `..._primitive` (or `primitive_...`

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        shape=['n_primitive'],
+        description='Extra per-primitive multiplier (dimensionless).',
+    )
+
+    # optional embedded point charge
+    point_charge = Quantity(
+        type=np.float32, description='Optional embedded point charge.'
+    )
+
+    _L_MAP = {
```
JosePizarro3, line 281: I don't know how many times we wrote this mapping... 😄 I had: https://github.com/FAIRmat-NFDI/nomad-parser-wannier90/blob/develop/src/nomad_parser_wannier90/parsers/win_parser.py#L60

@ladinesa @ndaelman-hu would love if we could formalize this somewhere in the package, separated from here.

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+      - Recording whether the code expanded shells in **Cartesian** or
+        **spherical harmonic** form.
+      - Tracking the **per-AO normalization factors** (to reconcile
+        conventions across codes).
+      - Providing an explicit **AO→shell mapping** for downstream
+        population analysis, MO coefficients, or density matrices.
+    """
+
+    num = Quantity(type=np.int32, description='Total number of AOs (contracted).')
+    cartesian = Quantity(type=bool, description='True=Cartesian, False=Spherical.')
```
JosePizarro3: a bit inconsistent with the MEnum above in AtomCenteredFunction

I would keep the type as a MEnum

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    Typical use cases:
+      - Recording whether the code expanded shells in **Cartesian** or
+        **spherical harmonic** form.
+      - Tracking the **per-AO normalization factors** (to reconcile
+        conventions across codes).
+      - Providing an explicit **AO→shell mapping** for downstream
+        population analysis, MO coefficients, or density matrices.
+    """
+
+    num = Quantity(type=np.int32, description='Total number of AOs (contracted).')
```
JosePizarro3: `n_atomic_orbitals`

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    cartesian = Quantity(type=bool, description='True=Cartesian, False=Spherical.')
+    shell_index = Quantity(
+        type=np.int32,
+        shape=['num'],
+        description="For each AO: index of AtomCenteredFunction (the 'shell') it belongs to.",
+    )
+    normalization = Quantity(
```
JosePizarro3, line 456: yeah, still confused also when thinking about `shell_normalization` from above.

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        description="""
+        The role of this basis set in the calculation:
+          - 'orbital': main orbital basis for the SCF
+          - 'auxiliary_scf': used for RI-J or density fitting in SCF
+          - 'auxiliary_post_hf': used in MP2, CC, etc. 
+          - 'cabs': complementary auxiliary basis for explicitly correlated (F12) methods.
+        """,
+    )
+
+    ao_ordering_convention = Quantity(
```
JosePizarro3, line 551: Mmm not 100% happy with this, as the order should be decided by us (and parsing takes care of putting it in the correct format)

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        Example value::
+
+            {"1": ["pz", "px", "py"],
+            "2": ["d0", "d+1", "d-1", "d+2", "d-2"]}
+
+        Parsers/writers should `json.loads(...)` this string.
+        """,
+    )
+
+    total_number_of_basis_functions = Quantity(
```
JosePizarro3: `n_total_basis_functions`

JosePizarro3 (approved):
Sorry for my late review, the notification ended up deep in my inbox

I like the implementation. Also reminded me that eventually I need to go through this for Wannier and Slater-Koster models.

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+
+    A single "atom-centered function" can be a linear combination of multiple
+    primitive Gaussians (or Slater-type orbitals, STOs).
+    In practice, these contract together to form the final basis function used by
+    the SCF or post-SCF method. Often, each contraction is labeled by its
+    angular momentum (e.g., s, p, d, f) and a set of exponents and coefficients.
+
+    This class also accepts *combined* shells ('sp', 'spd', 'spdf') as input.
+    On normalize(), it **expands** them into multiple single-ℓ shells and
+    replaces itself with those shells in the parent's `functional_compositions`.
```
ndaelman-hu:
```suggestion
```

> The main changes resolve the complexity around combined shells: I removed the ambiguous combined shell types ('sp', 'spd', etc.) and all associated expansion/deletion logic.

I guess this part of the docs should also be removed?

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        shape=['num'],
+        description='Per-AO factor to reconcile code conventions.',
+    )
 
 
 class AtomCenteredBasisSet(BasisSetComponent):
     """
-    Defines an atom-centered basis set.
+    Defines an **atom-centered basis set** for quantum chemistry calculations.
+    Unlike plane-wave methods, these expansions are typically built around each atom's
```
ndaelman-hu:
```suggestion
    These expansions are typically built around each atom's
```

I wouldn't try to draw comparisons or distinctions between different techniques for mostly different communities. Each can stand on its own.
Btw, in practice PW methods have a dedicated approach for describing the core electrons, be it pseudopotentials or LAPW. PW is just used for describing the valence electrons at a system-level.

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
     """
 
-    functional_composition = SubSection(
+    basis_set = Quantity(
+        type=str,
+        description="""
+        **Name** or label of the basis set as recognized by the code or standard 
+        library. Examples: "6-31G*", "cc-pVTZ", "def2-SVP", "STO-3G", "LANL2DZ" (ECP).
+        """,
+    )
```
ndaelman-hu, line 514: I like the flexible approach of both using a standardized name as well as its building blocks.
**Please elaborate on whether the former suffices, or if the latter should always be present.**

ndaelman-hu, line 514: Suggestion for future: I'm sure you're already aware of [BSE](https://www.basissetexchange.org/).
This can help bridge the gap between both levels of description.

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -184,22 +184,438 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+                    max_idx = int(np.max(si)) if len(si) > 0 else -1
+                    if max_idx >= len(self.functional_compositions):
+                        logger.warning(
+                            'shell_index contains index values outside the bounds of functional_compositions.'
+                        )
+                except Exception:
+                    # Log an error for unexpected exception during max_idx check
+                    logger.error(
+                        'Failed to perform max_idx sanity check on shell_index.'
+                    )
```
ndaelman-hu, line 649: Consider Alvin's `log` decorator. You might have to move this part to its own method if you just want to target that.

tests/test_basis_set.py
```diff
@@ -419,3 +422,442 @@ def test_quick_step() -> None:
...
     }
     # TODO: generate a QuickStep generator in the CP2K plugin
+
+
+# -------------------------------
+# AtomCenteredFunction (combined shells, padding/truncation, inference)
+# -------------------------------
+
+
+def test_acf_infers_n_primitive_and_ell() -> None:
```
ndaelman-hu, line 432: `pytest.mark.parametrize` for testing various edge cases, e.g. `n_primitive == 0` (when the parser doesn't populate the section), etc.

tests/test_basis_set.py
```diff
@@ -419,3 +422,442 @@ def test_quick_step() -> None:
...
+
+    assert acf.n_primitive == 3
+    assert acf.angular_momentum == 2  # d → ℓ=2
+
+
+def test_acf_invalid_length_reset() -> None:
+    """
+    Tests that if n_primitive is set, but coefficient/exponent arrays mismatch length,
+    they are correctly reset to None (due to validation failure).
+    """
```
ndaelman-hu, line 453: If this generates any logs, you can check them too. Just provide a test logger object.

tests/test_basis_set.py
ndaelman-hu: The same comment may apply to some of the other tests below. I leave it up to your discretion which you extend.

tests/test_basis_set.py
```diff
@@ -419,3 +422,442 @@ def test_quick_step() -> None:
...
+                    contributions=[
+                        BaseModelMethod(name='kinetic'),
+                        BaseModelMethod(name='electron-ion'),
+                        BaseModelMethod(name='hartree'),
+                    ],
+                    numerical_settings=[],
+                )
+            ],
+        )
+    )
```
ndaelman-hu, line 741: This looks like a re-usable setup. I would produce it outside of the method, so here you focus in testing.
If you use a similar template often, consider moving it to the pytest `config`.

ndaelman-hu (approved):
Good work! Very minor feedback.

src/nomad_simulations/schema_packages/basis_set.py
EBB2675: ah yes, thank you!

src/nomad_simulations/schema_packages/basis_set.py
EBB2675: even if parsing takes care of the canonicalization, it is common practice to keep track of the custom order if given

src/nomad_simulations/schema_packages/basis_set.py
EBB2675: danke sehr, I ll update the docs to make it clear:

ℓ is the shell angular momentum as usual. A contracted AO is a linear combination of primitives with the same center & ℓ; the degeneracy comes from the 2ℓ+1 components

src/nomad_simulations/schema_packages/basis_set.py
EBB2675: thanks alot, very important point.

**Policy**

- **`basis_set` name alone is sufficient** when the parser can resolve that name to a **unique, versioned definition for all species (including any ECPs)** under the emitting code’s version (e.g., `def2-TZVP`, `cc-pVTZ`, `STO-3G`). In that case, the name is authoritative and the explicit shells are optional.

- **Explicit shell/ECP blocks are required** if any of the following holds:
  1. per-species **mixtures** (e.g., `def2-TZVP` on C/H, `def2-SVP` on N),
  2. **augmented, truncated, or edited** sets (added diffuse, “no-f”, re-contracted, custom exponents/coefficients),
  3. **ambiguous aliases** or code-specific variants that can’t be uniquely mapped,
  4. **partial coverage** (some elements rely on a fallback or custom block),
  5. the code emits AOs that **don’t match** the canonicalized order unless we materialize the explicit shells here for correctness/provenance.

- **Parsers** are responsible for the mapping **code order → NOMAD canonical order** and for deciding whether the name is uniquely resolvable. When in doubt, they should include the explicit blocks **in addition to** the name.

I am adding this policy to the quantity docstring 👍

EBB2675: yes, just like the Libxc canonicalization in DFT, I wish to do it with BSE for basis sets :+1:

src/nomad_simulations/schema_packages/basis_set.py
EBB2675: done 👍

src/nomad_simulations/schema_packages/basis_set.py
EBB2675: yes, this quantity is unitless.
It represents the normalization factor applied to each contracted AO (shell) so that the function satisfies the chosen normalization convention (e.g., each AO integrates to 1 over all space).

It’s not related to degeneracy, but to how the primitive Gaussians are combined and scaled. For example, if primitives are individually normalized and then combined, a further contraction normalization may be required to ensure the final AO remains normalized.

I’ve clarified the description to reflect this :+1:

src/nomad_simulations/schema_packages/basis_set.py
EBB2675: done

src/nomad_simulations/schema_packages/basis_set.py
EBB2675: elaborated in the docstring

src/nomad_simulations/schema_packages/basis_set.py
EBB2675: im adding an issue about this 👍

PR #261, opened 2025-10-02
Migrate md

### LINKED ISSUE #28: Add MD relevant outputs

JFRudzinski:
Closes #28 #47 
- [x] add MD tests
- [x] What to do with MDAnalysis dependency?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,6 @@ license = { file = "LICENSE" }
 dependencies = [
     "nomad-lab[infrastructure]>=1.3.15",
     "matid>=2.0.0.dev2",
-    "MDAnalysis>=2.0.0",
 ]
 
 [project.urls]
@@ -40,6 +39,9 @@ dependencies = [
 "Documentation" = "https://nomad-coe.github.io/nomad-simulations/"
 
 [project.optional-dependencies]
+md = [
+    "MDAnalysis>=2.4.0",
+]
 dev = [
     'mypy>=1.15',
     'pytest>= 5.3.0, <8',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ dev = [
     'pytest>= 5.3.0, <8',
     'pytest-timeout>=1.4.2',
     'pytest-cov>=2.7.1',
+    'pytest-asyncio>=0.21.0',
     'ruff>=0.6',
     'structlog>=1.0',
     'typing-extensions>=4.12',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ license = { file = "LICENSE" }
 dependencies = [
     "nomad-lab[infrastructure]>=1.3.15",
     "matid>=2.0.0.dev2",
+    "MDAnalysis>=2.0.0",
 ]
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,6 +41,7 @@ dependencies = [
 [project.optional-dependencies]
 md = [
     "MDAnalysis>=2.4.0",
+    "networkx"
 ]
 dev = [
     'mypy>=1.15',
--- a/src/nomad_simulations/schema_packages/common.py
+++ b/src/nomad_simulations/schema_packages/common.py
@@ -4,7 +4,7 @@ from nomad.datamodel.metainfo.annotations import ELNAnnotation
 from nomad.metainfo import Datetime, Quantity
 
 
-class SimulationOutputs(ArchiveSection):
+class SimulationTime(ArchiveSection):
     """
     Contains time-related quantities.
     """
@@ -12,7 +12,7 @@ class SimulationOutputs(ArchiveSection):
     datetime_end = Quantity(
         type=Datetime,
         description="""
-        The date and time when this computation ended.
+        The date and time when the computation ended.
         """,
         a_eln=ELNAnnotation(component='DateTimeEditQuantity'),
     )
--- a/src/nomad_simulations/schema_packages/common.py
+++ b/src/nomad_simulations/schema_packages/common.py
@@ -4,7 +4,7 @@ from nomad.datamodel.metainfo.annotations import ELNAnnotation
 from nomad.metainfo import Datetime, Quantity
 
 
-class Time(ArchiveSection):
+class SimulationOutputs(ArchiveSection):
     """
     Contains time-related quantities.
     """
--- a/src/nomad_simulations/schema_packages/numerical_settings.py
+++ b/src/nomad_simulations/schema_packages/numerical_settings.py
@@ -382,6 +382,7 @@ class KMesh(Mesh):
         elif self.center == 'Monkhorst-Pack':
             try:
                 points_array = monkhorst_pack(size=self.grid)
+                points = points_array
                 offset = get_monkhorst_pack_size_and_offset(kpts=points_array)[-1]
             except ValueError:
                 logger.warning(
@@ -389,7 +390,7 @@ class KMesh(Mesh):
                 )
                 # this is a quick workaround: k_mesh.grid should be symmetry reduced
                 return None, None
-        return points_array, offset
+        return points, offset
 
     def get_k_line_density(
         self, reciprocal_lattice_vectors: pint.Quantity | None, logger: 'BoundLogger'
--- a/src/nomad_simulations/schema_packages/numerical_settings.py
+++ b/src/nomad_simulations/schema_packages/numerical_settings.py
@@ -381,8 +381,9 @@ class KMesh(Mesh):
             offset = np.array([0, 0, 0])
         elif self.center == 'Monkhorst-Pack':
             try:
-                points = monkhorst_pack(size=self.grid)
-                offset = get_monkhorst_pack_size_and_offset(kpts=points)[-1]
+                points_array = monkhorst_pack(size=self.grid)
+                points = list(points_array) if points_array is not None else None
+                offset = get_monkhorst_pack_size_and_offset(kpts=points_array)[-1]
             except ValueError:
                 logger.warning(
                     'Could not resolve `KMesh.points` and `KMesh.offset` from `KMesh.grid`. ASE `monkhorst_pack` failed.'
--- a/src/nomad_simulations/schema_packages/numerical_settings.py
+++ b/src/nomad_simulations/schema_packages/numerical_settings.py
@@ -382,7 +382,6 @@ class KMesh(Mesh):
         elif self.center == 'Monkhorst-Pack':
             try:
                 points_array = monkhorst_pack(size=self.grid)
-                points = list(points_array) if points_array is not None else None
                 offset = get_monkhorst_pack_size_and_offset(kpts=points_array)[-1]
             except ValueError:
                 logger.warning(
@@ -390,7 +389,7 @@ class KMesh(Mesh):
                 )
                 # this is a quick workaround: k_mesh.grid should be symmetry reduced
                 return None, None
-        return points, offset
+        return points_array, offset
 
     def get_k_line_density(
         self, reciprocal_lattice_vectors: pint.Quantity | None, logger: 'BoundLogger'
--- a/src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
+++ b/src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
@@ -89,6 +89,12 @@ class MolecularRadiusOfGyrationResults(TypedDict):
     system_ref: Any  # reference to the molecule section
 
 
+# Module-level constants for molecular dynamics calculations
+MAX_MOLECULES_PERFORMANCE_WARNING = (
+    50000  # Threshold for performance warning in MSD calculations
+)
+
+
 def _log_missing_dependency(dependency: str, calculation: str) -> str:
     """
     Generates a standardized warning message for missing dependencies.
@@ -103,20 +109,30 @@ def _log_missing_dependency(dependency: str, calculation: str) -> str:
     return f'{dependency} required to calculate {calculation}.'
 
 
-def _check_mda_dependency(function_name: str) -> bool:
+def _check_package_dependency(
+    package_name: str, function_name: str, extras_name: str = 'md'
+) -> bool:
     """
-    Checks if MDAnalysis is available and logs an error if not.
+    Checks if a required package is available and logs an error if not.
 
     Args:
-        function_name: The name of the function requiring MDAnalysis
+        package_name: The name of the package (e.g., 'MDAnalysis', 'networkx')
+        function_name: The name of the function requiring the package
+        extras_name: The extras group to install (default: 'md')
 
     Returns:
-        True if MDAnalysis is available, False otherwise
+        True if the package is available, False otherwise
     """
-    if not _HAS_MDA:
+    is_available = False
+    if package_name.lower() == 'mdanalysis':
+        is_available = _HAS_MDA
+    elif package_name.lower() == 'networkx':
+        is_available = _HAS_NETWORKX
+
+    if not is_available:
         LOGGER.error(
-            f'{function_name} requires MDAnalysis. '
-            'Please install with `pip install nomad-simulations[md]`.'
+            f'{function_name} requires {package_name}. '
+            f'Please install with `pip install nomad-simulations[{extras_name}]`.'
         )
         return False
     return True
@@ -275,7 +291,7 @@ def create_empty_universe(
     .. versionchanged:: 1.0.0
         Universes can now be created with 0 atoms
     """
-    if not _check_mda_dependency('create_empty_universe'):
+    if not _check_package_dependency('MDAnalysis', 'create_empty_universe', 'md'):
         return None
 
     if not n_atoms:
@@ -381,7 +397,7 @@ def archive_to_universe(
 
         bonds (tuple, shape=([])): list of tuples with the atom indices of each bond
     """
-    if not _check_mda_dependency('archive_to_universe'):
+    if not _check_package_dependency('MDAnalysis', 'archive_to_universe', 'md'):
         return None
 
     # TODO explicit return until we convert to data
@@ -644,7 +660,7 @@ def _get_molecular_bead_groups(
     """
     Creates bead groups based on the molecular types as defined by the MDAnalysis universe.
     """
-    if not _check_mda_dependency('_get_molecular_bead_groups'):
+    if not _check_package_dependency('MDAnalysis', '_get_molecular_bead_groups', 'md'):
         return {}
 
     # Input validation
@@ -697,7 +713,7 @@ def calc_molecular_rdf(
     max_mols : int
         Maximum number of molecules per bead group for calculating the rdf, for efficiency purposes.
     """
-    if not _check_mda_dependency('calc_molecular_rdf'):
+    if not _check_package_dependency('MDAnalysis', 'calc_molecular_rdf', 'md'):
         return {}
 
     # TODO 5k default for max_mols was set after > 50k was giving problems. Should do further testing to see where the appropriate limit should be set.
@@ -1001,7 +1017,7 @@ def calc_molecular_msd(
     max_mols : int
         Maximum number of molecules per bead group for calculating the msd, for efficiency purposes.
     """
-    if not _check_mda_dependency('calc_molecular_msd'):
+    if not _check_package_dependency('MDAnalysis', 'calc_molecular_msd', 'md'):
         return {}
 
     def parse_jumps(universe, selection):  # TODO Add output declaration
@@ -1065,7 +1081,7 @@ def calc_molecular_msd(
 
         return np.array(nojump_positions)
 
-    def mean_squared_displacement(start: np.ndarray, current: np.ndarray):
+    def mean_squared_displacement(start: np.ndarray, current: np.ndarray) -> float:
         """
         Calculates mean square displacement between current and initial (start) coordinates.
         """
@@ -1110,9 +1126,9 @@ def calc_molecular_msd(
     has_large_groups = False
     has_errors = False
 
-    if max_mols > 50000:
+    if max_mols > MAX_MOLECULES_PERFORMANCE_WARNING:
         LOGGER.warning(
-            'Calculating mean squared displacements for more than 50k molecules.'
+            f'Calculating mean squared displacements for more than {MAX_MOLECULES_PERFORMANCE_WARNING} molecules.'
             ' Expect long processing times!',
         )
 
@@ -1198,7 +1214,7 @@ def calc_radius_of_gyration(
     molecule_particle_indices : np.ndarray
         The indices of the particles corresponding to a single molecule for which the Rg will be calculated.
     """
-    if not _check_mda_dependency('calc_radius_of_gyration'):
+    if not _check_package_dependency('MDAnalysis', 'calc_radius_of_gyration', 'md'):
         return {}
 
     if molecule_particle_indices is None or len(molecule_particle_indices) == 0:
@@ -1245,7 +1261,7 @@ def calc_molecular_rg(
     """
     Calculates the radius of gyration as a function of time for each polymer in the system.
     """
-    if not _check_mda_dependency('calc_molecular_rg'):
+    if not _check_package_dependency('MDAnalysis', 'calc_molecular_rg', 'md'):
         return []
 
     if universe is None:
@@ -1287,11 +1303,7 @@ def get_molecules_from_bond_list(
     """
     Returns a list of dictionaries with molecule info from each instance in the list of bonds.
     """
-    if not _HAS_NETWORKX:
-        LOGGER.error(
-            'get_molecules_from_bond_list requires networkx. '
-            'Please install with `pip install nomad-simulations[md]`.'
-        )
+    if not _check_package_dependency('networkx', 'get_molecules_from_bond_list', 'md'):
         return []
 
     system_graph = networkx.empty_graph(n_particles)
@@ -1361,7 +1373,7 @@ def model_system_to_universe(system: ModelSystem, logger=None) -> MDAUniverse | 
     Returns:
         A new mda.Universe created from the given data.
     """
-    if not _check_mda_dependency('model_system_to_universe'):
+    if not _check_package_dependency('MDAnalysis', 'model_system_to_universe', 'md'):
         return None
 
     n_atoms = len(system.positions)
--- a/src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
+++ b/src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
@@ -4,9 +4,8 @@ from array import array
 from collections import namedtuple
 from collections.abc import Callable
 from itertools import chain
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
-import networkx
 import numpy as np
 from scipy import sparse
 from scipy.stats import linregress
@@ -21,12 +20,20 @@ try:
     _HAS_MDA = True
 except ImportError:
     _HAS_MDA = False
+    MDAnalysis = None  # type: ignore
 
-if not _HAS_MDA:
-    raise ImportError(
-        'MDAnalysis is required for this functionality. '
-        'Please re-install the plugin with `pip install nomad-simulations[md]`.'
-    )
+try:
+    import networkx
+
+    _HAS_NETWORKX = True
+except ImportError:
+    _HAS_NETWORKX = False
+    networkx = None  # type: ignore
+
+# Import types for type checking only (not at runtime)
+if TYPE_CHECKING:
+    from MDAnalysis import Universe as MDAUniverse
+    from MDAnalysis.core.groups import AtomGroup
 
 from nomad import atomutils
 from nomad.metainfo import MEnum, MSection, Quantity, Reference, Section, SubSection
@@ -36,6 +43,39 @@ from nomad.utils import get_logger
 from nomad_simulations.schema_packages.model_system import ModelSystem
 
 LOGGER = get_logger(__name__)
+
+
+def _log_missing_dependency(dependency: str, calculation: str) -> str:
+    """
+    Generates a standardized warning message for missing dependencies.
+
+    Args:
+        dependency: The required dependency (e.g., 'universe', 'bead_groups')
+        calculation: The calculation being attempted (e.g., 'RDF', 'MSD', 'radius of gyration')
+
+    Returns:
+        A formatted warning message string
+    """
+    return f'{dependency} required to calculate {calculation}.'
+
+
+def _check_mda_dependency(function_name: str) -> bool:
+    """
+    Checks if MDAnalysis is available and logs an error if not.
+
+    Args:
+        function_name: The name of the function requiring MDAnalysis
+
+    Returns:
+        True if MDAnalysis is available, False otherwise
+    """
+    if not _HAS_MDA:
+        LOGGER.error(
+            f'{function_name} requires MDAnalysis. '
+            'Please install with `pip install nomad-simulations[md]`.'
+        )
+        return False
+    return True
 
 
 class BeadGroup:
@@ -130,7 +170,7 @@ def create_empty_universe(
     flag_velocities: bool = False,
     flag_forces: bool = False,
     timestep: float | None = None,
-) -> MDAnalysis.Universe:
+) -> MDAUniverse | None:
     """Create a blank Universe
 
     This function was adapted from the function empty() within the MDA class Universe().
@@ -187,6 +227,8 @@ def create_empty_universe(
     .. versionchanged:: 1.0.0
         Universes can now be created with 0 atoms
     """
+    if not _check_mda_dependency('create_empty_universe'):
+        return None
 
     if not n_atoms:
         n_residues = 0
@@ -238,7 +280,7 @@ def archive_to_universe(
     system_index: int = 0,
     method_index: int = -1,
     model_index: int = -1,
-) -> MDAnalysis.Universe:
+) -> MDAUniverse | None:
     """Extract the topology from a provided run section of an archive entry
 
     Input:
@@ -291,6 +333,8 @@ def archive_to_universe(
 
         bonds (tuple, shape=([])): list of tuples with the atom indices of each bond
     """
+    if not _check_mda_dependency('archive_to_universe'):
+        return None
 
     try:
         sec_run = archive.run[-1]
@@ -543,14 +587,17 @@ def archive_to_universe(
 
 
 def _get_molecular_bead_groups(
-    universe: MDAnalysis.Universe, moltypes: list[str] = []
+    universe: MDAUniverse | None, moltypes: list[str] = []
 ) -> dict[str, BeadGroup]:
     """
     Creates bead groups based on the molecular types as defined by the MDAnalysis universe.
     """
+    if not _check_mda_dependency('_get_molecular_bead_groups'):
+        return {}
+
     # Input validation
     if universe is None:
-        LOGGER.warning('Universe required to create beads.')
+        LOGGER.warning(_log_missing_dependency('universe', 'beads'))
         return {}
 
     if not moltypes:
@@ -572,7 +619,7 @@ def _get_molecular_bead_groups(
 
 
 def calc_molecular_rdf(
-    universe: MDAnalysis.Universe,
+    universe: MDAUniverse | None,
     bead_groups: dict[str, BeadGroup],
     n_traj_split: int = 10,
     n_prune: int = 1,
@@ -598,9 +645,12 @@ def calc_molecular_rdf(
     max_mols : int
         Maximum number of molecules per bead group for calculating the rdf, for efficiency purposes.
     """
+    if not _check_mda_dependency('calc_molecular_rdf'):
+        return {}
+
     # TODO 5k default for max_mols was set after > 50k was giving problems. Should do further testing to see where the appropriate limit should be set.
     if bead_groups is None or not bead_groups:
-        LOGGER.warning('bead_groups required to calculate RDF.')
+        LOGGER.warning(_log_missing_dependency('bead_groups', 'RDF'))
         return {}
 
     if (
@@ -608,7 +658,7 @@ def calc_molecular_rdf(
         or not universe.trajectory
         or universe.trajectory[0].dimensions is None
     ):
-        LOGGER.warning('universe required to calculate RDF.')
+        LOGGER.warning(_log_missing_dependency('universe', 'RDF'))
         return {}
 
     n_frames = universe.trajectory.n_frames
@@ -877,7 +927,7 @@ def shifted_correlation_average(
 
 
 def calc_molecular_mean_squared_displacements(
-    universe: MDAnalysis.Universe,
+    universe: MDAUniverse | None,
     bead_groups: dict[str, BeadGroup],
     max_mols: int = 5000,
 ) -> dict[str, Any]:
@@ -894,10 +944,10 @@ def calc_molecular_mean_squared_displacements(
     max_mols : int
         Maximum number of molecules per bead group for calculating the msd, for efficiency purposes.
     """
+    if not _check_mda_dependency('calc_molecular_mean_squared_displacements'):
+        return {}
 
-    def parse_jumps(
-        universe: MDAnalysis.Universe, selection: MDAnalysis.AtomGroup
-    ):  # TODO Add output declaration
+    def parse_jumps(universe, selection):  # TODO Add output declaration
         """
         See __get_nojump_positions().
         """
@@ -923,9 +973,7 @@ def calc_molecular_mean_squared_displacements(
 
         return jump_data
 
-    def generate_nojump_matrices(
-        universe: MDAnalysis.Universe, selection: MDAnalysis.AtomGroup
-    ):  # TODO Add output declaration
+    def generate_nojump_matrices(universe, selection):  # TODO Add output declaration
         """
         See __get_nojump_positions().
         """
@@ -941,9 +989,7 @@ def calc_molecular_mean_squared_displacements(
         )
         return nojump_matrices
 
-    def get_nojump_positions(
-        universe: MDAnalysis.Universe, selection: MDAnalysis.AtomGroup
-    ) -> np.ndarray:
+    def get_nojump_positions(universe, selection) -> np.ndarray:
         """
         Unwraps the positions to create a continuous trajectory without jumps across periodic boundaries.
         """
@@ -970,7 +1016,7 @@ def calc_molecular_mean_squared_displacements(
         return (vec**2).sum(axis=1).mean()
 
     if bead_groups is None or not bead_groups:
-        LOGGER.warning('bead_groups required to calculate MSD.')
+        LOGGER.warning(_log_missing_dependency('bead_groups', 'MSD'))
         return {}
 
     if (
@@ -978,7 +1024,7 @@ def calc_molecular_mean_squared_displacements(
         or not universe.trajectory
         or universe.trajectory[0].dimensions is None
     ):
-        LOGGER.warning('universe required to calculate MSD.')
+        LOGGER.warning(_log_missing_dependency('universe', 'MSD'))
         return {}
 
     n_frames = universe.trajectory.n_frames
@@ -1073,7 +1119,7 @@ def calc_molecular_mean_squared_displacements(
 
 
 def calc_radius_of_gyration(
-    universe: MDAnalysis.Universe, molecule_particle_indices: np.ndarray
+    universe: MDAUniverse | None, molecule_particle_indices: np.ndarray
 ) -> dict[str, Any]:
     """
     Calculates the radius of gyration as a function of time for the particles 'molecule_particle_indices'.
@@ -1081,9 +1127,12 @@ def calc_radius_of_gyration(
     molecule_particle_indices : np.ndarray
         The indices of the particles corresponding to a single molecule for which the Rg will be calculated.
     """
+    if not _check_mda_dependency('calc_radius_of_gyration'):
+        return {}
+
     if molecule_particle_indices is None or len(molecule_particle_indices) == 0:
         LOGGER.warning(
-            'molecule_particle_indices is required to calculate radius of gyration'
+            _log_missing_dependency('molecule_particle_indices', 'radius of gyration')
         )
         return {}
 
@@ -1092,7 +1141,7 @@ def calc_radius_of_gyration(
         or not universe.trajectory
         or universe.trajectory[0].dimensions is None
     ):
-        LOGGER.warning('universe is None. Cannot calculate radius of gyration.')
+        LOGGER.warning(_log_missing_dependency('universe', 'radius of gyration'))
         return {}
     selection = ' '.join([str(i) for i in molecule_particle_indices])
     selection = f'index {selection}'
@@ -1100,7 +1149,7 @@ def calc_radius_of_gyration(
     rg_results: dict[str, Any] = {}
     rg_results['times'] = []
     rg_results['value'] = []
-    time_unit = hasattr(universe.trajectory.time, 'units')
+    time_unit = getattr(universe.trajectory.time, 'units', None)
     for __ in universe.trajectory:
         rg_results['times'].append(
             universe.trajectory.time.magnitude
@@ -1120,17 +1169,22 @@ def calc_radius_of_gyration(
 
 
 def calc_molecular_radius_of_gyration(
-    universe: MDAnalysis.Universe, system_hierarchy: MSection
+    universe: MDAUniverse | None, system_hierarchy: MSection
 ) -> list[dict[str, Any]]:
     """
     Calculates the radius of gyration as a function of time for each polymer in the system.
     """
+    if not _check_mda_dependency('calc_molecular_radius_of_gyration'):
+        return []
+
     if universe is None:
-        LOGGER.warning('universe required to calculate molecular radius of gyration')
+        LOGGER.warning(
+            _log_missing_dependency('universe', 'molecular radius of gyration')
+        )
         return []
     if system_hierarchy is None or not system_hierarchy:
         LOGGER.warning(
-            'system_topology required to calculate molecular radius of gyration.'
+            _log_missing_dependency('system_topology', 'molecular radius of gyration')
         )
         return []
 
@@ -1162,6 +1216,13 @@ def get_molecules_from_bond_list(
     """
     Returns a list of dictionaries with molecule info from each instance in the list of bonds.
     """
+    if not _HAS_NETWORKX:
+        LOGGER.error(
+            'get_molecules_from_bond_list requires networkx. '
+            'Please install with `pip install nomad-simulations[md]`.'
+        )
+        return []
+
     system_graph = networkx.empty_graph(n_particles)
     system_graph.add_edges_from([(i[0], i[1]) for i in bond_list])
     molecules = [
@@ -1233,7 +1294,7 @@ def get_composition(children_names: list[str]) -> str:
     return formula
 
 
-def mda_universe_from_nomad_atoms(system, logger=None):
+def mda_universe_from_nomad_atoms(system, logger=None) -> MDAUniverse | None:
     """Returns an instance of mda.Universe from a NOMAD Atoms-section.
 
     Args:
@@ -1242,6 +1303,9 @@ def mda_universe_from_nomad_atoms(system, logger=None):
     Returns:
         A new mda.Universe created from the given data.
     """
+    if not _check_mda_dependency('mda_universe_from_nomad_atoms'):
+        return None
+
     n_atoms = len(system.positions)
     n_residues = 1
     atom_resindex = [0] * n_atoms
--- a/src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
+++ b/src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
@@ -1123,7 +1123,7 @@ def calc_molecular_radius_of_gyration(
 
     rg_results = []
     for molgroup in system_hierarchy:
-        for molecule, i_mol in enumerate(molgroup.subsystems):
+        for i_mol, molecule in enumerate(molgroup.subsystems):
             sec_monomer_groups = molecule.subsystems
             group_type = (
                 sec_monomer_groups[0].branch_label if sec_monomer_groups else None
--- a/src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
+++ b/src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
@@ -5,6 +5,7 @@ from collections import namedtuple
 from collections.abc import Callable
 from itertools import chain
 from typing import Any
+
 import networkx
 import numpy as np
 from scipy import sparse
@@ -31,8 +32,8 @@ from nomad import atomutils
 from nomad.metainfo import MEnum, MSection, Quantity, Reference, Section, SubSection
 from nomad.units import ureg
 from nomad.utils import get_logger
-from nomad_simulations.schema_packages.model_system import ModelSystem
 
+from nomad_simulations.schema_packages.model_system import ModelSystem
 
 LOGGER = get_logger(__name__)
 
--- a/src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
+++ b/src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
@@ -1,4 +1,35 @@
+from typing import Any, Callable, Optional
+from itertools import chain
+from collections import namedtuple
 import numpy as np
+from array import array
+from scipy import sparse
+from scipy.stats import linregress
+import networkx
+import numpy as np
+import MDAnalysis
+from MDAnalysis.core.topology import Topology
+from MDAnalysis.core.universe import Universe
+import MDAnalysis.analysis.rdf as MDA_RDF
+from MDAnalysis.core._get_readers import get_reader_for
+
+from nomad.utils import get_logger
+from nomad.units import ureg
+from nomad import atomutils
+
+from nomad.datamodel.data import ArchiveSection
+from nomad.metainfo import (
+    SubSection,
+    Section,
+    Quantity,
+    MEnum,
+    Reference,
+    MSection,
+)
+from nomad.datamodel.hdf5 import HDF5Dataset
+from nomad.datamodel.metainfo.workflow import Link
+
+from nomad_simulations.schema_packages.model_system import ModelSystem
 
 
 class BeadGroup(object):
@@ -1034,17 +1065,17 @@ def calc_molecular_mean_squared_displacements(
 
 
 def calc_radius_of_gyration(
-    universe: MDAnalysis.Universe, molecule_atom_indices: np.ndarray
+    universe: MDAnalysis.Universe, molecule_particle_indices: np.ndarray
 ) -> dict[str, Any]:
     """
-    Calculates the radius of gyration as a function of time for the atoms 'molecule_atom_indices'.
+    Calculates the radius of gyration as a function of time for the particles 'molecule_particle_indices'.
 
-    molecule_atom_indices : np.ndarray
-        The indices of the atoms corresponding to a single molecule for which the Rg will be calculated.
+    molecule_particle_indices : np.ndarray
+        The indices of the particles corresponding to a single molecule for which the Rg will be calculated.
     """
-    if molecule_atom_indices is None or len(molecule_atom_indices) == 0:
+    if molecule_particle_indices is None or len(molecule_particle_indices) == 0:
         LOGGER.warning(
-            'molecule_atom_indices is required to calculate radius of gyration'
+            'molecule_particle_indices is required to calculate radius of gyration'
         )
         return {}
 
@@ -1055,11 +1086,10 @@ def calc_radius_of_gyration(
     ):
         LOGGER.warning('universe is None. Cannot calculate radius of gyration.')
         return {}
-    selection = ' '.join([str(i) for i in molecule_atom_indices])
+    selection = ' '.join([str(i) for i in molecule_particle_indices])
     selection = f'index {selection}'
     molecule = universe.select_atoms(selection)
     rg_results: dict[str, Any] = {}
-    rg_results['type'] = 'molecular'
     rg_results['times'] = []
     rg_results['value'] = []
     time_unit = hasattr(universe.trajectory.time, 'units')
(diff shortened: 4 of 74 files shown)


src/nomad_simulations/schema_packages/common.py
```diff
@@ -4,7 +4,7 @@
 from nomad.metainfo import Datetime, Quantity
 
 
-class Time(ArchiveSection):
+class SimulationOutputs(ArchiveSection):
```
Bernadette-Mohr: SimulationResults

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +875,509 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+    value = Quantity(
+        type=np.float64,
+        shape=['n_bins'],
+        description="""
+        Values of the property.
+        """,
+    )
+
+
+class DiffusionConstant(PhysicalProperty):
```
Bernadette-Mohr: EnsembleProperty

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +875,509 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+    barostat_parameters = SubSection(sub_section=BarostatParameters.m_def, repeats=True)
+
+    shear_parameters = SubSection(sub_section=ShearParameters.m_def, repeats=True)
+
+    free_energy_calculation_parameters = SubSection(
+        sub_section=FreeEnergyCalculationParameters.m_def, repeats=True
+    )
+
+
+class EnsembleProperty(PhysicalProperty):
```
Bernadette-Mohr, line 892: Placeholder for bins?

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +875,509 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+        unit='m^2/s',
+        description="""
+        Values of the diffusion constants.
+        """,
+    )
+
+
+class CorrelationFunction(PhysicalProperty):
+    """
+    Base section for time correlation functions calculated from a trajectory.
```
Bernadette-Mohr: add abstract class specifications to description (also for ensembleaverage)

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +875,509 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+
+
+class MeanSquaredDisplacement(CorrelationFunction):
+    """
+    Section containing information about a calculation of any mean squared displacements (msds).
+    """
+
+    _msd_results = None
+
+    times = Quantity(
```
Bernadette-Mohr: needed?

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +875,509 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+
+    mean_squared_displacements = SubSection(
+        sub_section=MeanSquaredDisplacement.m_def, repeats=True
+    )
+
+    radial_distribution_functions = SubSection(
+        sub_section=RadialDistributionFunction.m_def, repeats=True
+    )
+
+    radius_of_gyration = SubSection(sub_section=RadiiOfGyration.m_def, repeats=True)
```
Bernadette-Mohr: duplicate

src/nomad_simulations/schema_packages/workflow/trajectory.py
```diff
@@ -0,0 +1,201 @@
...
+
+if TYPE_CHECKING:
+    from nomad_simulations.schema_packages.workflow.molecular_dynamics import (
+        FreeEnergyCalculationParameters,
+    )
+else:
+    FreeEnergyCalculationParameters = 'FreeEnergyCalculationParameters'
+
+
+class TrajectoryProperty(PhysicalProperty):
```
Bernadette-Mohr: ConfigurationalProperties?

tests/workflow/test_molecular_dynamics.py
```diff
@@ -13,7 +13,7 @@ def test_inputs_outputs(self, logger, archive):
         archive.data.outputs = [Outputs() for _ in range(self.n_outputs)]
```
Bernadette-Mohr, line 286: add tests

JFRudzinski:
@ladinesa @EBB2675 @Bernadette-Mohr There is no rush on a review due to my vacation, but it would be great to have some feedback by ~23.10

@ladinesa Note about the MDAnalysis dependency, I just added it explicitly for simplicity of passing the tests, but I guess we need to think about a conditional dependency or local imports...idk

src/nomad_simulations/schema_packages/workflow/general.py
```diff
@@ -181,11 +190,33 @@ def normalize(self, archive: EntryArchive, logger: BoundLogger):
         self.map_tasks(archive, logger=logger)
 
 
+class SerialWorkflowResults(SimulationWorkflowResults):
+    _label = 'Thermodynamics ouputs'
```
ladinesa: typo

src/nomad_simulations/schema_packages/workflow/general.py
```diff
@@ -50,26 +57,28 @@ def normalize(self, archive: EntryArchive, logger: BoundLogger) -> None:
             self.initial_method = archive.data.model_method[0]
 
 
-class SimulationWorkflowResults(ArchiveSection):
+class SimulationWorkflowResults(SimulationOutputs):
```
ladinesa: I am not sure I understand why this should inherit from simulation outputs. I think they should be independent and only contains reference to the resulting outputs section

src/nomad_simulations/schema_packages/workflow/general.py
```diff
@@ -181,11 +190,33 @@ def normalize(self, archive: EntryArchive, logger: BoundLogger):
...
+        sub_section=FreeEnergyCalculations.m_def, repeats=True
+    )
+
+
 class SerialWorkflow(SimulationWorkflow):
     """
     Base class for workflows where tasks are executed sequentially.
     """
 
+    # ? What is this mapping for exactly?
```
ladinesa: This is to initialize the wokflow results section if not defined, At the moment most workflow classes have this since there maybe some special functionalities that need to be implemented. We can also refactor this so it goes to the parent class and the the child class only needs to define a default results section.

src/nomad_simulations/schema_packages/workflow/general.py
JFRudzinski: `SimulationOutputs` previously `Time` just has these start/end datetimes for the cpu etc...I guess I thought we might want the same quantities to store these datetimes from workflow libraries/managers. But maybe I misunderstood the usage. I don't have a strong opinion here

ladinesa: I think we should separately define time quantities for workflow, as this will unnecessarily complicate the hierarchy. `SimulationResults` contains only a reference to the final output section but it is an Outputs child.

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +874,493 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+    _task_label = 'Step'
+
+    method = SubSection(sub_section=MolecularDynamicsMethod)
+
+    results = SubSection(sub_section=MolecularDynamicsResults)
+
+    def map_inputs(self, archive: EntryArchive, logger: BoundLogger = None) -> None:
+        if not self.method:
+            self.method = MolecularDynamicsMethod()
+
```
ladinesa, line 1356: maybe you call super().map_inputs() here so serial wokflow can add the method section to inputs and also to the first task automaticall.y

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +874,493 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+    def map_outputs(self, archive: EntryArchive, logger: BoundLogger = None) -> None:
+        if not self.results:
+            self.results = MolecularDynamicsResults()
+
```
ladinesa, line 1361: same here so results are added to the outputs.

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +874,493 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+        super().normalize(archive, logger)
+
+        self._universe = self.get_universe(archive)
+        if self._universe is None:
+            return
+
+        self._bead_groups = _get_molecular_bead_groups(self._universe)
+
+        # calculate molecular radial distribution functions
+        self.radial_distribution_functions = self.get_molecular_rdfs(archive)
```
ladinesa: I would recommend to use append or extend here but not sure if it works as esma pointed out, maybe `self.m_add_sub_section`? It is some minor thing but technically your function returning a list but repeating sub-sections are MSubSectionList so there may be some problems later.

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +874,493 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+            system_hierarchy = sec_system.subsystems
+            rg_results = calc_molecular_radius_of_gyration(
+                self._universe, system_hierarchy
+            )
+
+            # Create workflow RG sections
+            for rg in rg_results:
+                sec_rgs = RadiiOfGyration()  # Use plural name for trajectory workflow
+                sec_rgs._rg_results = rg
+                sec_rgs.normalize(archive, logger)
```
ladinesa: do you need to call normalize here right away? If yes, then maybe put a check in you RadiysOfGyration def to skip normalize if done.

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +874,493 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+                return []
+
+            system_hierarchy = sec_system.subsystems
+            rg_results = calc_molecular_radius_of_gyration(
+                self._universe, system_hierarchy
+            )
+
+            # Create workflow RG sections
+            for rg in rg_results:
+                sec_rgs = RadiiOfGyration()  # Use plural name for trajectory workflow
```
ladinesa, line 1315: is it really Radii?

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +874,493 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+    _bead_groups: dict[str, BeadGroup] | None
+
+    def __init__(self, m_def: Section = None, m_context: Context = None, **kwargs):
+        super().__init__(m_def, m_context, **kwargs)
+        self._cache: dict[str, Any] = {}
+        self._universe = None
+        self._bead_groups = None
+
+    @log
+    def get_universe(self, archive) -> MDAnalysis.Universe | None:
```
ladinesa: not sure how many uses this function but try to check if self._universe already exists.

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +874,493 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+        sub_section=RadialDistributionFunction.m_def, repeats=True
+    )
+
+    # Type annotations for private attributes
+    _universe: MDAnalysis.Universe | None
+    _bead_groups: dict[str, BeadGroup] | None
+
+    def __init__(self, m_def: Section = None, m_context: Context = None, **kwargs):
+        super().__init__(m_def, m_context, **kwargs)
+        self._cache: dict[str, Any] = {}
```
ladinesa: you can also simply put them at the top without redifining init

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +874,493 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
-    def map_inputs(self, archive: EntryArchive) -> None:
-        if not self.model:
-            self.model = MolecularDynamicsModel()
-        logger = self.map_inputs.__annotations__['logger']
-        super().map_inputs(archive, logger=logger)
+    def get_molecular_rdfs(
+        self, archive: EntryArchive
+    ) -> list[RadialDistributionFunction]:
+        # logger = self.get_molecular_rdfs.__annotations__['logger']
+        if self.radial_distribution_functions is not None:
```
ladinesa: this is a list so it should be `if not self.rad...`

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +874,493 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+            return self.radial_distribution_functions
+
+        if self._universe is None:
```
ladinesa: you already have this check in normalize. do you separately call it somewhere? If not maybe make them private

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +874,493 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+    trajectory = Quantity(
+        type=Reference(ModelSystem),
+        shape=['n_steps'],
+        description="""
+        Reference to the system of each step in the trajectory.
+        """,
+    )
+
+    # Properties
+    diffusion_constants = SubSection(sub_section=DiffusionConstant.m_def, repeats=True)
```
ladinesa, line 1072: are we not putting there under results?

pyproject.toml
```diff
@@ -31,6 +31,7 @@ license = { file = "LICENSE" }
 dependencies = [
     "nomad-lab[infrastructure]>=1.3.15",
     "matid>=2.0.0.dev2",
+    "MDAnalysis>=2.0.0",
```
EBB2675: does it make sense to add this dependency as a gated import, so that users who don’t use MD features can still import everything else?

```
[project.optional-dependencies]
md = [
    "MDAnalysis>=2.4.0",
]
```

also with some verbosity for users who will use MDAnalysis:


```
try:
    import MDAnalysis as mda
    _HAS_MDA = True
except ImportError:
    _HAS_MDA = False
```
anddd


```
if not _HAS_MDA:
    raise ImportError(
        "MDAnalysis is required for this functionality. "
        "Please install it via `pip install MDAnalysis`."
    )

```

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1258 @@
...
+    flag_trajectory: bool = False,
+    flag_velocities: bool = False,
+    flag_forces: bool = False,
+    timestep: float | None = None,
+) -> MDAnalysis.Universe:
+    """Create a blank Universe
+
+    This function was adapted from the function empty() within the MDA class Universe().
+    The only difference is that the Universe() class is imported directly here, whereas in the
+    original function is is passed as a function argument, since the function there is a classfunction.
```
EBB2675: is is

EBB2675: classfunction: class method

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1258 @@
...
+
+    if atom_resindex is None:
+        LOGGER.warning(
+            'Residues specified but no atom_resindex given.  '
+            'All atoms will be placed in first Residue.',
+        )
+
+    if residue_segindex is None:
+        LOGGER.warning(
+            'Segments specified but no segment_resindex given.  '
```
EBB2675: `residue_segindex` instead of segment_resindex

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1258 @@
...
+        if atom.get('mass'):
+            masses[atom_ind] = ureg.convert(
+                atom.mass.magnitude, atom.mass.units, ureg.amu
+            )
+        if atom.get('charge'):
+            charges[atom_ind] = ureg.convert(
+                atom.charge.magnitude, atom.charge.units, ureg.e
+            )
+
+    # get the atom positions, velocites, and box dimensions
```
EBB2675: velocities

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1258 @@
...
+                    which defaults to 0.
+        counter (bool, opt.):
+                    If True, returns length of frames (in general number of particles specified)
+        average (bool, opt.):
+                    If True,
+    Returns:
+        tuple:
+            A list of length N that contains the indices of the frames at which
+            the time series was calculated and a numpy array of shape (segments, N)
+            that holds the (non-avaraged) correlation data
```
EBB2675: averaged

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1258 @@
...
+
+            if has_counter == True: adds number of counts to output tupel.
```
EBB2675: tuple

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1258 @@
...
+) -> list[dict[str, Any]]:
+    """
+    Calculates the radius of gyration as a function of time for each polymer in the system.
+    """
+    if universe is None:
+        LOGGER.warning('universe required to calculate molecular radius of gyration')
+        return []
+    if system_hierarchy is None or not system_hierarchy:
+        LOGGER.warning(
+            'system_topology require to calculate molecular radius of gyration.'
```
EBB2675: The existing logger handles this case, which is great, but the message could be a bit more explicit

 e.g. :

   `"No hierarchy or topology found: per-molecule Rg will not be computed. Compute system-wide Rg or provide bonds/molnums to enable per-molecule grouping."`

so users and I will know that an empty result is expected when no hierarchy/bonds are available

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1258 @@
...
+            velocities=vels,
+            forces=forces,
+            dt=timestep,
+        )
+
+    return universe
+
+
+# TODO: update run to data
+def archive_to_universe(
```
EBB2675, line 342: I think there are multiple native H5MD variants, is the intention here to convert them all into a LAMMPS-H5MD-style layout internally (so the parsers will handle the conversion and the schema only operates on this unified layout)?

EBB2675 (approved):

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1258 @@
...
+
+    Returns:
+        bead_groups: returns  bead group object that links and subdivides the MDAnalysis.AtomsGroup, when attributes are requested.
+    """
+
+    def __init__(self, atoms, compound='fragments'):
+        """Initialize with an AtomGroup instance.
+        Will split based on keyword 'compounds' (residues or fragments).
+
+        self._atoms: AtomGroup (total number of atoms)
```
Bernadette-Mohr, line 156: Shouldn't this be "_particles"/"_beads", for semantic consistency?

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1258 @@
...
+
+    Args:
+        object (MDAnalysis.AtomsGroup): complete set of atoms for consideration.
+
+    Returns:
+        bead_groups: returns  bead group object that links and subdivides the MDAnalysis.AtomsGroup, when attributes are requested.
+    """
+
+    def __init__(self, atoms, compound='fragments'):
+        """Initialize with an AtomGroup instance.
```
Bernadette-Mohr: Is this the MDAnalysis.Atom**-s-**Group, or something else?

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1258 @@
...
+        Will split based on keyword 'compounds' (residues or fragments).
+
+        self._atoms: AtomGroup (total number of atoms)
+        self.compound: str (dictates type of grouping)
+        self._nbeads: int (total number of beads)
+        self.positions: list (total number of "compounds")
+        """
+        self._atoms = atoms
```
Bernadette-Mohr, line 161: Again, "_particles"/"_beads"?

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1258 @@
...
+
+    @property  # type: ignore
+    @MDAnalysis.lib.util.cached('universe')
+    def universe(self):
+        return self._atoms.universe
+
+
+def get_bond_list_from_model_contributions(
+    sec_run: MSection, method_index: int = -1, model_index: int = -1
+) -> list[tuple]:
```
Bernadette-Mohr: Uses `runschema`, needs to be updated?

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
EBB2675: doesnt the `BeadGroup` operate directly on an `MDAnalysis` `AtomGroup`?

So it’s meant to mirror the MDAnalysis API, not our “particles” abstraction

If this class were part of the schema layer (not the MDAnalysis utility layer), i.e., interacting with our `AtomsState` or Particles, then _particles would indeed be the right semantic name

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1258 @@
...
+        metainfo_universe.add_TopologyAttr('molnums', residue_molnums)
+    if len(residue_moltypes) > 0:
+        metainfo_universe.add_TopologyAttr('moltypes', residue_moltypes)
+
+    # add the box dimensions
+    for frame_ind, frame in enumerate(metainfo_universe.trajectory):
+        metainfo_universe.atoms.dimensions = dimensions[frame_ind]
+
+    # add the bonds
+    if hasattr(metainfo_universe, 'bonds'):
```
Bernadette-Mohr, line 649: In what scenario would this be the case?

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1258 @@
...
+
+LOGGER = get_logger(__name__)
+
+
+class BeadGroup:
+    """A helper class for calculating properties of groups of atoms ("beads") with the MDAnalysis package.
+    See https://github.com/MDAnalysis/mdanalysis/issues/1891#issuecomment-387138110 by @richardjgowers with performance improvements.
+
+    Args:
+        object (MDAnalysis.AtomsGroup): complete set of atoms for consideration.
```
Bernadette-Mohr: I think this should be `MDAnalysis.AtomGroup` (no "s").

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -1,14 +1,764 @@
...
+        type=int,
+        shape=[],
+        description='Index of each Lambdas.values for the current simulation step/state '
+        '(only valid if all targets share an aligned λ grid).',
+    )
+    # Otherwise, prefer per-target scalar λ values for the current state (one scalar per Lambdas entry):
+    current_lambdas = Quantity(
+        type=np.float64, shape=['*'], description='Scalar λ per Lambdas entry order.'
+    )
+    # TODO not really sure how to deal with this, cause it corresponds to a hierarchical workflow that may not be contained in the same upload
```
Bernadette-Mohr, line 657: If no topology information is part of a lambda upload, I think we can assume that nothing changed and link to the topology from the first lambda, or maybe the last previous where a topology was provided?

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +874,493 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+        description="""
+        Mean squared displacement values.
+        """,
+    )
+
+
+class MolecularDynamicsResults(SerialWorkflowResults):
     _label = 'MD results'
 
+    finished_normally = Quantity(
```
Bernadette-Mohr: How are we going to determine this?

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
JFRudzinski: I think there might be some confusion here:

- First, `archive_to_universe` is not about h5md, rather it is a backmapping from the nomad schema to mdanalysis (mdanalysis creates an object for efficient trajectory analysis from standard MD input/output files, which helps us to parse from simulation code to metainfo; archive_to_universe creates an (equivalent) MD analysis object directly from the archive)

- I moved the [nomad-h5md docs](https://fairmat-nfdi.github.io/nomad-parser-plugins-simulation/parsers/h5md/h5md_about.html) to the parser repo.

- [H5MD](https://h5md.nongnu.org/) was proposed as an exchange format for the MD community using hdf5 files.  The original specification is quite lean. The idea was that the community could expand upon it to develop a detailed standard (very much like the NEXUS perspective). There were only 1 or 2 small amendments though as far as I know.

- The idea with h5md-nomad is to suggest a detailed specification on top of the original schema, leaning on our internal nomad schema but restructuring for convenience in the hdf5 format. There is some work to do in terms of proper versioning and community input here...if we could manage this sometime next year though it would probably be of serious interest to the community (@Bernadette-Mohr something to think about; input from others is also very welcome here)

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
JFRudzinski: tbh I can't remember why I put it there. Upon inspection indeed it seems not possible to have bonds already in the universe but I also think I probably put it there for a reason so I am a bit hesitant to remove it...will try to do some tests

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
JFRudzinski: I'm actually thinking now if these log messages are necessary in this case, since it is a public utility function and this is just a check for the input of the function. 

The message you provided, while I agree useful for the user in the case of our pipeline, is really referencing a specific usage of this function. I don't want to overcomplicate the communication with the user, e.g., when this function is called via a parser, the user has no way to impact the presence of the system hierarchy.

Thoughts?

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
JFRudzinski: Can you make a note to bring this up in our next meeting. I am a bit confused about the relevance of topology and also a little about my own note as I forgot the context. In the end I think this is out of scope of this PR since we need to address the FEC workflows separately, but I would like to at least extend my current TODO note

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
JFRudzinski: Determined on a per parser basis from the logs if possible, maybe add a TODO for the Gromacs and Lammps parsers

Note: elevated this quantity to SimulationWorkflowResults()

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
EBB2675: at first glance i thought a bit more verbosity would be nice, but given that its utility function i would like to take my initial comment back

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
JFRudzinski: what do you mean? This is under `workflow2.results`

ladinesa: sorry my bad just saw original l136 class MolecularDynamics and I thought this is sub-section is under it

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
JFRudzinski: I use extend now, it seems to work. @EBB2675 what was the potential issue? Should I not do this?

JFRudzinski:
@ndaelman-hu @EBB2675 can you please advise on these failing mypy tests

EBB2675:
they all point to:
`
src/nomad_simulations/schema_packages/workflow/beyond_hf.py:7: error: Module "nomad_simulations.schema_packages.workflow.general" has no attribute "SimulationWorkflowModel"; maybe "SimulationWorkflowMethod", "SimulationWorkflow", or "SimulationWorkflowResults"?  [attr-defined]`

pyproject.toml
```diff
@@ -39,11 +39,15 @@ dependencies = [
 "Documentation" = "https://nomad-coe.github.io/nomad-simulations/"
 
 [project.optional-dependencies]
+md = [
```
ladinesa, line 42: not sure why this is an optional dep

EBB2675:
the part i have merged this morning has the old style of naming in terms of workflows, your branch changed the syntax for all workflows, except for what i have just merged

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+from itertools import chain
+from typing import Any
+
+import networkx
+import numpy as np
+from scipy import sparse
+from scipy.stats import linregress
+
+try:
+    import MDAnalysis
```
ladinesa, line 14: i suggest you make it as a core dep and this try should go away

ndaelman-hu:
> @ndaelman-hu @EBB2675 can you please advise on these failing mypy tests

```mypy
src/nomad_simulations/schema_packages/numerical_settings.py:385: error: Incompatible types in assignment (expression has type "list[Any]", variable has type "ndarray[tuple[int, ...], dtype[Any]]")  [assignment]
```

This one (under Python3.10) I solved and merged yesterday into `develop`. You can just merge `develop` into this branch to have it resolved.

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+#             contribution.n_interactions
+#         ):  # all bonds have been grouped into one contribution
+#             bond_list = [tuple(indices) for indices in atom_indices]
+#         else:
+#             bond_list.append(tuple(contribution.atom_indices))
+
+#     return bond_list
+
+
+def _create_empty_universe(
```
ladinesa: why is this private?

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+
+        velocities (float, shape=(n_frames,n_atoms,3)): atom velocities
+
+        dimensions (float, shape=(n_frames,6)): box dimensions (nb - currently assuming a cubic box!)
+
+        bonds (tuple, shape=([])): list of tuples with the atom indices of each bond
+    """
+
+    try:
+        sec_run = archive.run[-1]
```
ladinesa, line 408: why is this reading run? oh there is a TODO but then it should not be here.

src/nomad_simulations/schema_packages/workflow/general.py
```diff
@@ -19,7 +26,7 @@ class SimulationTask(Task):
     pass
 
 
-class SimulationWorkflowModel(ArchiveSection):
+class SimulationWorkflowMethod(ArchiveSection):
```
ndaelman-hu, line 30: > they all point to: ` src/nomad_simulations/schema_packages/workflow/beyond_hf.py:7: error: Module "nomad_simulations.schema_packages.workflow.general" has no attribute "SimulationWorkflowModel"; maybe "SimulationWorkflowMethod", "SimulationWorkflow", or "SimulationWorkflowResults"? [attr-defined]`

This is the offending line. You didn't update the terminology throughout. Do a dry run with 

```bash
sed -R 's/SimulationWorkflowModel/SimulationWorkflowMethod/' *.py
```

and add the option `-i` once you've confirmed that it correctly updates  the remaining references.

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+    return np.unique(np.int_(ls) - 1 + first)
+
+
+def _correlation(function, positions: list[float]):
+    iterator = iter(positions)
+    start_frame = next(iterator)
+    return map(lambda f: function(start_frame, f), chain([start_frame], iterator))
+
+
+def _calc_diffusion_constant(
```
ladinesa: the name is more specific than what the function really does

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+        return np.array(nojump_positions)
+
+    def mean_squared_displacement(start: np.ndarray, current: np.ndarray):
+        """
+        Calculates mean square displacement between current and initial (start) coordinates.
+        """
+        vec = start - current
+        return (vec**2).sum(axis=1).mean()
+
+    if bead_groups is None or not bead_groups:
```
ladinesa: `if not bead_groups` suffices

pyproject.toml
JFRudzinski: it is related to trying to enable easier stand-alone use of the parsers. If someone wants to just use vasp eg they need to install MDAnalysis (both for the parser dep and for nomad-simulations). So, I made the MDAnalysis optional in both packages...is it not a good approach? Do you have an alternative suggestion?

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+
+    n_frames = universe.trajectory.n_frames
+    if n_frames < 50:
+        LOGGER.warning(
+            'At least 50 frames required to calculate molecular'
+            ' mean squared displacements, skipping.',
+        )
+        return {}
+
+    dt = getattr(universe.trajectory, 'dt')
```
ladinesa, line 1112: does `universe.trajectory.dt` not work?

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+                )
+                selection = ' '.join([str(i) for i in atom_indices_rnd])
+                selection = f'index {selection}'
+                ags_moltype_rnd = universe.select_atoms(selection)
+                bead_groups[moltype] = BeadGroup(ags_moltype_rnd, compound='fragments')
+                LOGGER.warning(
+                    'Maximum number of molecules for calculating the msd has been reached.'
+                    ' Will make a random selection for calculation.'
+                )
+            except Exception:
```
ladinesa, line 1155: not sure how big moltypes is but in any case i suggest moving warning outside loop

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+                LOGGER.warning(
+                    'Error in selecting random molecules for large group when calculating msd. Skipping this molecule type.'
+                )
+                del_list.append(i_moltype)
+
+    for index in sorted(del_list, reverse=True):
+        del moltypes[index]
+
+    msd_results: dict[str, Any] = {}
+    msd_results['type'] = 'molecular'
```
ladinesa: put all intitializations in one line

pyproject.toml
ladinesa: Ah I see then ignore related comments

JFRudzinski:
> > @ndaelman-hu @EBB2675 can you please advise on these failing mypy tests
> 
> ```
> src/nomad_simulations/schema_packages/numerical_settings.py:385: error: Incompatible types in assignment (expression has type "list[Any]", variable has type "ndarray[tuple[int, ...], dtype[Any]]")  [assignment]
> ```
> 
> This one (under Python3.10) I solved and merged yesterday into `develop`. You can just merge `develop` into this branch to have it resolved.

@ndaelman-hu I already incorporated your changes into my branch, that is why I asked...unless I am missing something

pyproject.toml
ladinesa: But thinking about it again, the plugin will not load since you raise the error? Maybe put the try in workflow module init.

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+            'molecule_particle_indices is required to calculate radius of gyration'
+        )
+        return {}
+
+    if (
+        not universe
+        or not universe.trajectory
+        or universe.trajectory[0].dimensions is None
+    ):
+        LOGGER.warning('universe is None. Cannot calculate radius of gyration.')
```
ladinesa: Maybe better:  universe required to calculate radius of gyration

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+        return {}
+    selection = ' '.join([str(i) for i in molecule_particle_indices])
+    selection = f'index {selection}'
+    molecule = universe.select_atoms(selection)
+    rg_results: dict[str, Any] = {}
+    rg_results['times'] = []
+    rg_results['value'] = []
+    time_unit = hasattr(universe.trajectory.time, 'units')
```
ladinesa: this will return bool?  but you use it as unit in l1113?

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+
+
+def calc_molecular_radius_of_gyration(
+    universe: MDAnalysis.Universe, system_hierarchy: MSection
+) -> list[dict[str, Any]]:
+    """
+    Calculates the radius of gyration as a function of time for each polymer in the system.
+    """
+    if universe is None:
+        LOGGER.warning('universe required to calculate molecular radius of gyration')
```
ladinesa: you have this same warning maybe introduce a function to generate msg for diff properties

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+def get_molecules_from_bond_list(
+    n_particles: int,
+    bond_list: list[tuple],
+    particle_types: list[str] = [],
+    particles_typeid: array | None = None,
+) -> list[dict[str, Any]]:
+    """
+    Returns a list of dictionaries with molecule info from each instance in the list of bonds.
+    """
+    system_graph = networkx.empty_graph(n_particles)
```
ladinesa, line 1314: this is not included in the deps?

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+    system_graph.add_edges_from([(i[0], i[1]) for i in bond_list])
+    molecules = [
+        system_graph.subgraph(c).copy()
+        for c in networkx.connected_components(system_graph)
+    ]
+    molecule_info: list[dict[str, Any]] = []
+    molecule_dict: dict[str, Any] = {}
+    for mol in molecules:
+        molecule_dict = {}
```
ladinesa: put all inits in one line

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+            for i_bond, bond in enumerate(bond_list_names)
+        }
+
+    if sorted(mol_1['names']) != sorted(mol_2['names']):
+        return False
+
+    bond_list_dict_1 = get_bond_list_dict(mol_1)
+    bond_list_dict_2 = get_bond_list_dict(mol_2)
+
+    if bond_list_dict_1 == bond_list_dict_2:
```
ladinesa: `return  bond_list_dict_1 == bond_list_dict_2`

ndaelman-hu:
> > > @ndaelman-hu @EBB2675 can you please advise on these failing mypy tests
> > 
> > 
> > ```
> > src/nomad_simulations/schema_packages/numerical_settings.py:385: error: Incompatible types in assignment (expression has type "list[Any]", variable has type "ndarray[tuple[int, ...], dtype[Any]]")  [assignment]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This one (under Python3.10) I solved and merged yesterday into `develop`. You can just merge `develop` into this branch to have it resolved.
> 
> @ndaelman-hu I already incorporated your changes into my branch, that is why I asked...unless I am missing something

okay, lemme take another look

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+        return True
+
+    return False
+
+
+def get_composition(children_names: list[str]) -> str:
```
ladinesa: i have a feeling this is private but please look at which functions are private

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1273 @@
...
+    """
+    Generates a generalized "chemical formula" based on the provided list `children_names`,
+    with the format X(m)Y(n) for children_names X and Y of quantities m and n, respectively.
+    """
+    children_count_tup = np.unique(children_names, return_counts=True)
+    formula = ''.join([f'{name}({count})' for name, count in zip(*children_count_tup)])
+    return formula
+
+
+def mda_universe_from_nomad_atoms(system, logger=None):
```
ladinesa: put type annot

ladinesa: you have `archive_to_universe` so I suggest `system?atoms?_to_universe`

src/nomad_simulations/schema_packages/workflow/general.py
```diff
@@ -50,26 +57,92 @@ def normalize(self, archive: EntryArchive, logger: BoundLogger) -> None:
...
+    """
+    Contains time-related quantities.
+    """
+
+    datetime_end = Quantity(
+        type=Datetime,
+        description="""
+        The date and time when the workflow ended.
+        """,
+        a_eln=ELNAnnotation(component='DateTimeEditQuantity'),
```
ladinesa: do we enable editing?

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -381,8 +381,9 @@ def resolve_points_and_offset(
             offset = np.array([0, 0, 0])
         elif self.center == 'Monkhorst-Pack':
             try:
-                points = monkhorst_pack(size=self.grid)
-                offset = get_monkhorst_pack_size_and_offset(kpts=points)[-1]
+                points_array = monkhorst_pack(size=self.grid)
+                points = list(points_array) if points_array is not None else None
+                offset = get_monkhorst_pack_size_and_offset(kpts=points_array)[-1]
```
ndaelman-hu, line 386: This causes one of the `mypy` errors in Python 3.10 and 3.11.
I guess it originally fixed an old mypy error? Anyhow, this fix is no longer valid, as I updated the type signature. Line 385 should simply be removed.

src/nomad_simulations/schema_packages/workflow/general.py
ndaelman-hu: While you correctly updated `SimulationWorkflowMethod`, this should be extended to other, more specific names, as evidenced by the current  `mypy` error:

```mypy
 src/nomad_simulations/schema_packages/workflow/configuration_interaction.py:7: error: Module "nomad_simulations.schema_packages.workflow.beyond_hf" has no attribute "BeyondHFModel"; maybe "BeyondHFMethod"?  [attr-defined]
```

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -1,14 +1,777 @@
...
+        ]
+        if len(output_idxs) > 1:
+            logger.warning(
+                'Multiple "output" Lambda entries detected; only the first one will be kept.'
+            )
+            for i in sorted(output_idxs[1:], reverse=True):
+                del self.lambdas[i]
+
+        # Check uniqueness of interaction_type (non-"output")
+        seen = {}
```
ladinesa: maybe simpler
```
interaction_types = [lambda.interaction_type for lambda in self.lambdas]
if len(interaction_types) != len(list(set(interaction_types))):
    logger.warning...
```

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -1,14 +1,777 @@
...
+                    'Duplicate Lambda interaction types detected; duplicates may indicate inconsistent setup.'
+                )
+
+        # Range checks for λ grids (alchemical: expect [0,1] for common targets)
+        strict_targets = {'vdw', 'coulomb', 'bonded', 'mass'}
+        if self.calc_type == 'alchemical':
+            for lam in self.lambdas or []:
+                if lam.interaction_type in strict_targets and lam.values:
+                    vmin, vmax = min(lam.values), max(lam.values)
+                    if vmin < 0.0 or vmax > 1.0:
```
ladinesa: put warning outside of loop

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -1,14 +1,777 @@
...
+                        logger.warning(
+                            'One or more Lambda grids for alchemical targets are outside the expected [0,1] range.'
+                        )
+        else:
+            logger.info(
+                'Lambda range enforcement skipped for non-alchemical calculation type.'
```
ladinesa: is this info necessary?

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
ladinesa: simpler
```
values = np.array([lam.values for lam in self.lambas if lam.interaction_type in strict-targets]).flattten()
if np.any((values > 1)|(values<0)):
 ...
```

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -1,14 +1,777 @@
...
+        if len(output_idxs) > 1:
+            logger.warning(
+                'Multiple "output" Lambda entries detected; only the first one will be kept.'
+            )
+            for i in sorted(output_idxs[1:], reverse=True):
+                del self.lambdas[i]
+
+        # Check uniqueness of interaction_type (non-"output")
+        seen = {}
+        for lam in self.lambdas or []:
```
ladinesa: self.lambdas by default is `[]` so no need for on here and in the folllowing

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -1,14 +1,777 @@
...
+        if self.current_lambda_index is not None:
+            grids = [
+                tuple(getattr(lam, 'values', []) or []) for lam in (self.lambdas or [])
+            ]
+            non_empty = [g for g in grids if len(g) > 0]
+            if len(non_empty) > 1 and any(g != non_empty[0] for g in non_empty[1:]):
+                logger.warning(
+                    'A single current_lambda_index was provided but Lambda grids differ; index will be ignored.'
+                )
+                self.current_lambda_index = None
```
ladinesa, line 715: Not sure if I understand but I have the impression that this is not a section quantity but only a variable to handle normalization logic

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -1,14 +1,777 @@
...
+        # Validate per-target scalar λs if present
+        current_lambdas = self.current_lambdas
+        if current_lambdas is not None:
+            n_targets = len(self.lambdas or [])
+            if len(current_lambdas) != n_targets:
+                logger.warning(
+                    'The number of current Lambda values does not match the number of Lambda entries.'
+                )
+            if self.calc_type == 'alchemical':
+                for lam_val in current_lambdas:
```
ladinesa: make this one line so you get rid of multiple warnings

ndaelman-hu:
Regarding the `mypy` errors exclusive to `python3.12`, namely

```mypy
src/nomad_simulations/schema_packages/force_field.py:948: SyntaxWarning: invalid escape sequence '\d'
  Phase shift $\delta$.
src/nomad_simulations/schema_packages/force_field.py:1056: SyntaxWarning: invalid escape sequence '\p'
  $V = k_1(\theta - \theta_0)^2 + k_2(\phi - \phi_0)^2 + k_3(\theta - \theta_0)(\phi - \phi_0)$
src/nomad_simulations/schema_packages/force_field.py:1098: SyntaxWarning: invalid escape sequence '\o'
  Equilibrium improper angle $\omega_0$.
src/nomad_simulations/schema_packages/force_field.py:1138: SyntaxWarning: invalid escape sequence '\d'
  Phase shift $\delta$.
```

I traced them back to a [new specification](https://docs.python.org/3/whatsnew/3.12.html#other-language-changes):

> A backslash-character pair that is not a valid escape sequence now generates a [SyntaxWarning](https://docs.python.org/3/library/exceptions.html#SyntaxWarning), instead of [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning). For example, re.compile("\d+\.\d+") now emits a [SyntaxWarning](https://docs.python.org/3/library/exceptions.html#SyntaxWarning) ("\d" is an invalid escape sequence, use raw strings for regular expression: re.compile(r"\d+\.\d+")). In a future Python version, [SyntaxError](https://docs.python.org/3/library/exceptions.html#SyntaxError) will eventually be raised, instead of [SyntaxWarning](https://docs.python.org/3/library/exceptions.html#SyntaxWarning). (Contributed by Victor Stinner in [gh-98401](https://github.com/python/cpython/issues/98401).)

In other words, you have to escape the backslash, i.e. `$\delta$` becomes `$\\delta$`.
No clue why this triggers now, though, as you did not modify this file. Does our CI/CD build a new minor version each time perhaps?

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -1,14 +1,777 @@
...
+        output_grid = None
+        for lam in self.lambdas or []:
+            if lam.interaction_type == 'output' and lam.values:
+                output_grid = list(lam.values)
+                break
+        if output_grid is not None:
+            for lam in self.lambdas or []:
+                if lam.interaction_type != 'output':
+                    vals = lam.values
+                    if not vals or len(vals) == 0:
```
ladinesa: if lamba.values type is array this will be prolematic

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +887,477 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+
+    def _populate_output_rgs(
+        self, archive: EntryArchive, rg_results: list, logger
+    ) -> None:
+        """
+        Populate the radius of gyration data in the archive's output sections.
+        This is separated from the workflow calculation to maintain clean separation of concerns.
+        """
+        try:
+            data = archive.data[-1]
```
ladinesa: data is not a repeating sub-section

ladinesa: data.model_system

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +887,477 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+            n_frames = rg.get('n_frames')
+            if len(sec_systems) != n_frames:
+                logger.warning(
+                    'Mismatch in length of system references in calculation and calculated Rg values. '
+                    'Will not store Rg values under calculation section'
+                )
+                continue
+
+            for out in sec_outputs:
+                if not out.system_ref:  # TODO check relevance
```
ladinesa: out.model_system_ref

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +887,477 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+                    continue
+                sys_ind = out.system_ref.m_parent_index
```
ladinesa: here also

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +887,477 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+                # Use the correct property name from outputs.py (radii_of_gyration, plural)
+                if not hasattr(out, 'radii_of_gyration') or not out.radii_of_gyration:
+                    sec_rgs_out = out.m_create(RadiusOfGyration)
+                    sec_rgs_out.kind = rg.get('type')  # check quant names
+                else:
+                    sec_rgs_out = out.radii_of_gyration[0]
+
+                # TODO Fix this assignment fails with TypeError
+                try:
+                    sec_rgs_out.atomsgroup_ref = [rg.get('atomsgroup_ref')]
```
ladinesa: I do not see atomsgroup_ref in RadiusOfGyration

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +887,477 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+        Also populates the corresponding output sections in the archive.
+        """
+        logger = self.get_molecular_rgs.__annotations__['logger']
+
+        # Check if already calculated
+        if self.radii_of_gyration:
+            return self.radii_of_gyration
+
+        try:
+            data = archive.data[-1]
```
ladinesa: see previous comments

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -124,28 +887,477 @@ class MolecularDynamicsModel(SimulationWorkflowModel):
...
+                break  # TODO Should transfer Rg's to workflow results if they are already supplied in calculation
+
+        workflow_rgs = []
+
+        if not flag_rgs:  # Calculate RGs if not already present
+            if self._universe is None:
+                logger.warning('Universe is None, cannot calculate RGs')
+                return []
+
+            system_hierarchy = sec_system.subsystems
```
ladinesa: sub_systems

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
JFRudzinski: possibly, not sure tbh, some of this is quite preliminary until we continue with the MD dev that has been put on hold for quite a while, I added a TODO to consider demoting to class variable for normalization

pyproject.toml
JFRudzinski: I have adjusted the strategy by not throwing the import error, and instead using the flag to exit out of functions using the dependency. Let me know what you think about this

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
JFRudzinski: I didn't remove this because it has the try so it will not be an issue if run is missing, but I have now added an explicit exit if the run section is missing. 

tbh I didn't fully work through the entire utils module, this will need to be tested and improved in connection with the Gromacs and Lammps parsers...but the changes you suggested already help us in this direction, thanks!

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
JFRudzinski: I don't really understand, you want it more general or more specific? Like `_calc_diffusion_constant_from_msd()`?

ladinesa: more general like fit or something.

pyproject.toml
JFRudzinski: now also added something in __init__ to prevent crashing of entire MD import

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1351 @@
...
+            result = np.array(list(correlate(start_frame)))
+        else:
+            result += np.array(list(correlate(start_frame)))
+    result = np.array(result)
+    result = result / len(start_frames)
+
+    return correlation_times, result
+
+
+def calc_molecular_mean_squared_displacements(
```
ladinesa: nitpick: just for consistency you used calc_molecular_rdf so maybe calc_molecular_msd?

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1351 @@
...
+    universe: MDAUniverse | None,
+    bead_groups: dict[str, BeadGroup],
+    max_mols: int = 5000,
+) -> dict[str, Any]:
```
ladinesa: can be in a separate mr but point of improvement maybe use TypedDict

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1351 @@
...
+                np.array(
+                    np.vstack([m[:i_frame, :].sum(axis=0) for m in nojump_matrices]).T
+                )
+                * box
+            )
+            nojump_positions.append(selection.positions - delta)
+
+        return np.array(nojump_positions)
+
+    def mean_squared_displacement(start: np.ndarray, current: np.ndarray):
```
ladinesa: output type

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1351 @@
...
+
+    if bead_groups == {}:
+        return bead_groups
+
+    moltypes = [moltype for moltype in bead_groups.keys()]
+    del_list = []
+    has_large_groups = False
+    has_errors = False
+
+    if max_mols > 50000:
```
ladinesa: maybe define this as a module variable?

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1351 @@
...
+    msd_results['error_diffusion_constant'] = np.array(
+        msd_results['error_diffusion_constant']
+    )
+
+    return msd_results
+
+
+def calc_radius_of_gyration(
+    universe: MDAUniverse | None, molecule_particle_indices: np.ndarray
+) -> dict[str, Any]:
```
ladinesa: also here try to use Typeddict

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -0,0 +1,1351 @@
...
+        dependency: The required dependency (e.g., 'universe', 'bead_groups')
+        calculation: The calculation being attempted (e.g., 'RDF', 'MSD', 'radius of gyration')
+
+    Returns:
+        A formatted warning message string
+    """
+    return f'{dependency} required to calculate {calculation}.'
+
+
+def _check_mda_dependency(function_name: str) -> bool:
```
ladinesa: generalize this to include other packages like networkx see l1248

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -1,14 +1,773 @@
...
+    from MDAnalysis.core.topology import Topology
+    from MDAnalysis.core.universe import Universe
+
+    _HAS_MDA = True
+except ImportError:
+    _HAS_MDA = False
+
+if not _HAS_MDA:
+    raise ImportError(
+        'MDAnalysis is required for this functionality. '
```
ladinesa: use the same check md dependency formatter and utils

ladinesa: not sure if i understand but the md utils can be imported partially without the md deps but not the schema?

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
JFRudzinski: please check if my implementation is what you meant

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
JFRudzinski: yes, what you said WAS true, but after thinking about it, this was only an artifact of me not also updating my approach for the schema module.

I now took a similar approach to the utils: cond import of MDA with a flag if not avail, then in the schema module I simply make sure to exit out of the universe creation function if no MDA. Otherwise the only other links to MDA are from the functions in utils, which are already covered

ndaelman-hu:
@JFRudzinski I don't think I have to follow up further here. Will unsub from notifications for this PR.

JFRudzinski:
@ladinesa please let me know if I can merge this or we need more refinements

ladinesa (approved):

PR #263, opened 2025-10-07
Replace `GeometricSpace` and `Cell` by `Representation`

ndaelman-hu:
This PR consolidates the cell representation architecture by replacing the deprecated `GeometricSpace`, `Cell`, and `AtomicCell` classes with a unified `Representation` + `RelativeRepresentation` aprpoach. While almost all quantities are retained, the structure now:

- Puts forward `ModelSystem` (root) as the default representation, whereas alternative representations are stored under `ModelSystem.representations`. This avoids unnecessary clutter for when there is only one representation.
- `RealtiveRepresentation` now acts as a base class for defining various kinds of base sections, not just crystallographic ones.

It also updates the type signature of `from_ase_atoms` and makes it a standalone method. This simplifies the initiation of `ModelSystem`s from `ase.Atoms` by providing a default clean starting point. The method';s mapping is also extended. These extensions only make sense in light of `Representation`

**NOTE**: it is strongly recommended to review this PR using _semantic diff_.


ndaelman-hu:
@JFRudzinski I would also like to add GlobalSymmnetry here, from #253 
The reason is that it nicely shows how representations interact with symmetry.

ladinesa:
@ndaelman-hu sorry i probably will not be able to look at this this week.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -341,57 +252,72 @@ class AtomicCell(Cell):
...
-        if not isinstance(parent, ModelSystem):
-            logger.warning(
-                'Parent is not a ModelSystem → geometric-space normalisation skipped.'
-            )
-            return
 
-        atoms = parent.to_ase_atoms(logger=logger)
-        if atoms is None:
-            return  # parent already logged the problem
+class RelativeRepresentation(Representation):
```
JFRudzinski: Just a small comment (perhaps too nitpicky): Could this just be named `AlternateRepresentation` or `AlternativeRepresentation` or something like that? I feel that `Relative` puts too much focus on the transformation. More generally to symmetry transformations, it may not be of interest/scope/practicality to store the transformation function

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -256,83 +129,121 @@ def __eq__(self, other):
...
         """,
     )
 
     periodic_boundary_conditions = Quantity(
         type=bool,
         shape=[3],
         description="""
-        If periodic boundary conditions are applied to each direction of the crystal axes.
+        Whether periodic boundary conditions are applied.
+        Runs over each direction of the crystal axes.
```
JFRudzinski: remove crystal specificity of description

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -256,83 +129,121 @@ def __eq__(self, other):
...
+        The second index runs over the reference Cartesian coordinate system, in the order $x, y, z$.
         """,
     )
 
     periodic_boundary_conditions = Quantity(
         type=bool,
         shape=[3],
         description="""
-        If periodic boundary conditions are applied to each direction of the crystal axes.
+        Whether periodic boundary conditions are applied.
```
JFRudzinski: This is inherently linked to the lattice vectors, should we mention this? Should we add a warning if this is set yes without lattice vectors? (I guess only relevant if we have functionalities which apply the PBC)

JFRudzinski (approved):
Very nice! I like this new structure.

I see the back compatibility with cell, this is nice. Can you add an issue in the parser plugin branch to remove cell usage from all parsers, to make sure we tackle this at some point?

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -256,83 +129,121 @@ def __eq__(self, other):
         return np.allclose(self.representative_variable, other.representative_variable)
 
 
-class Cell(GeometricSpace):
+class Representation(ArchiveSection):
```
ladinesa, line 135: Not sure why the need to define this base class separate from `RelativeRepresentation`? This is not used anywhere aside from cell backward compatibility, which I think we should not be doing at the development stage.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1027,12 +955,50 @@ class ModelSystem(System):
...
+        sub_section=RelativeRepresentation.m_def,
+        repeats=True,
+        description="""
+        Alternative representations of the original model system, e.g., in a different simulation box or crystal cell.
+        """,
+    )
+
     sub_systems = SubSection(sub_section=SectionProxy('ModelSystem'), repeats=True)
 
+    # Backward compatibility: add cell as a subsection that references Representation
```
ladinesa: imo we do not need to provide this at the development stage it will just create a lot of clutter.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -766,7 +705,7 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
             self.m_cache['elemental_composition'] = formula.elemental_composition
 
 
-class ModelSystem(System):
+class ModelSystem(System, Representation):
```
ladinesa, line 698: for sure I do not understand the logic here but in general multiple inheritances is a symptom of not optimal design

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: `Representation` is also a base class to `ModelSystem`, which is the default representation. It contains the common quantities definitions between `ModelSystem` and `ModelSystem.representations`.

It's not particularly meant for backwards compatibility, but may provide it (see bottom of `model_system.py`).

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: Yes, I'd prefer not using it either. Let's see:

- `System` comes from NOMAD, so the sub-optimal design are how definitions are split along the projects. 
- `Representation` is just a way to avoid defining the same quantities twice. I worst case, it can be removed, though I would explore other options first.
- Any specific tests I should run to see whether double inheritance causes an issue?

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1027,12 +955,50 @@ class ModelSystem(System):
...
+        """
+        logger = self.get_geometric_space_for_cell.__annotations__['logger']
+
+        atoms = self.to_ase_atoms(
+            representation_index=0 if self.representations else None, logger=logger
+        )
+        if atoms is None:
+            return  # parent already logged the problem
+
+        try:
```
ladinesa: The log decorator already does this, you can provide a custom warning via `exc_msg`

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: Sorry for the extraneous comment. No, `cell` is being removed as per the title of the PR.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1138,12 +1112,36 @@ def to_ase_atoms(self) -> 'ase.Atoms | None':
...
+                ase_atoms.set_pbc([False, False, False])
+
+            if self.lattice_vectors is not None:
+                ase_atoms.set_cell(self.lattice_vectors.to('angstrom').magnitude)
+            else:
+                logger.info('No lattice_vectors found in original ModelSystem data.')
+        else:
+            # Use specified representation
+            if not self.representations:
+                raise ValueError('No representations available in ModelSystem.')
```
ladinesa: not sure how this is handled downstream but I would suggest not raising errors

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1155,9 +1153,9 @@ def to_ase_atoms(self) -> 'ase.Atoms | None':
                     cell_section.lattice_vectors.to('angstrom').magnitude
                 )
             else:
-                logger.info('No lattice_vectors found in cell[0].')
-        else:
-            logger.warning('No cell section available in ModelSystem.')
+                logger.info(
+                    f'No lattice_vectors found in representations[{representation_index}].'
```
ladinesa: no formatted logs

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1565,3 +1538,155 @@ def is_molecule(self) -> bool:
                     return False
 
         return True
+
+
+@log
+def from_ase_atoms(ase_atoms: ase.Atoms) -> ModelSystem:
```
ladinesa: this should at least be a static method.

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -404,11 +411,14 @@ def get_k_line_density(
...
         """
         # Initial check
-        if not KSpaceFunctionalities().validate_reciprocal_lattice_vectors(
-            reciprocal_lattice_vectors=reciprocal_lattice_vectors,
-            logger=logger,
-            check_grid=True,
-            grid=self.grid,
+        if (
+            reciprocal_lattice_vectors is None
+            or not KSpaceFunctionalities().validate_reciprocal_lattice_vectors(
```
ladinesa, line 420: could be in a separate mr but looks like `validate_reciprocal_lattice_vectors` should be a static method

tests/conftest.py
```diff
@@ -347,7 +350,7 @@ def model_system() -> ModelSystem:
 
 
 @pytest.fixture(scope='session')
-def atomic_cell() -> AtomicCell:
+def atomic_cell() -> Representation:
```
ladinesa: should you redefine fixture too atomic_representation?

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: I would rather leave it empty and log that.

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: Honestly, this is legacy stuff that has little usage in our schema. It just interfaces with `ase`...

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: I will be resolving this convo, deciding to keep the class. If you want to discuss further, feel free to re-open.

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: I'm kind of torn here. For sure, I don't want this method operating directly on the `ModelSystem` object, as it introduces a whole host of unnecessary state checks. By instantiating from a new object we are sure to have a clean setup, that is simply filled up.

So, this method can stand by itself. Granted, given that `to_ase_atoms` is in the `ModelSystem` namespace, one could expect its dual within the same namespace.
`staticmethod` is indeed compatible with generating a new `ModelSystem` object, but from the interface alone, one would expect it to write to the object.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1027,12 +955,50 @@ class ModelSystem(System):
...
+        description="""
+        Alternative representations of the original model system, e.g., in a different simulation box or crystal cell.
+        """,
+    )
+
     sub_systems = SubSection(sub_section=SectionProxy('ModelSystem'), repeats=True)
 
+    # Backward compatibility: add cell as a subsection that references Representation
+    cell = SubSection(sub_section=Representation.m_def, repeats=True)
+
```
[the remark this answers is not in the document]
ndaelman-hu: Added as a property after the `ModelSystem` definition

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: done

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: You're right. It's a small correction, I'll add it here. Thx for spotting this.

tests/conftest.py
ndaelman-hu: done

src/nomad_simulations/schema_packages/model_system.py
ladinesa: I think a static method is really more apt for this.

ndaelman-hu: Okay, it does not really clash with the clean state. It's more a question of managing expectations.

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: like this?

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1027,12 +955,50 @@ class ModelSystem(System):
...
+            cell = atoms.get_cell()
+            if self.representations and len(self.representations) > 0:
+                cell_section = self.representations[0]
+                cell_section.volume = cell.volume * ureg.angstrom**3
+        except Exception as exc:
+            logger.warning(
+                'Failed to extract geometric-space data from ASE cell.',
+                exc_info=exc,
+            )
+
```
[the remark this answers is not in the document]
ndaelman-hu: @JFRudzinski do you think that we should retain quantities like vector length and angles? Note that they are also encoded in the lattice vectors. For cyrstals, the other symmetry descriptors also cover all the relevant angular information.

src/nomad_simulations/schema_packages/model_system.py
ladinesa: something like
```
@log(exc_msg=''Failed to extract geometric-space data from ASE cell.'')
def get_geometric_space_for_cell
```
Though I have to update the log utl so it does not append the exact error but would add it as exc_info. Maybe you can include this tiny change in l58 of utils.py i.e. `_logger.warning(f'{_exc_msg}', exc_info=e)`

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: swapped error raising for logging debug and info messages

tests/test_model_system.py
```diff
@@ -218,14 +378,20 @@ def test_normalize(self):
...
-            if len(sys.cell) >= 2:
-                assert sys.cell[1].type in ['primitive', 'conventional']
+            # Check if primitive or conventional cells are present in the symmetry section
+            if sys.symmetry and (
+                sys.symmetry.primitive_cell or sys.symmetry.conventional_cell
+            ):
+                if sys.symmetry.primitive_cell:
+                    assert sys.symmetry.primitive_cell.name == 'primitive'
+                if sys.symmetry.conventional_cell:
+                    assert sys.symmetry.conventional_cell.name == 'conventional'
```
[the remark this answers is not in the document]
ndaelman-hu: done

src/nomad_simulations/schema_packages/utils/utils.py
```diff
@@ -56,7 +56,7 @@ def wrapper(*args, **kwargs):
                     },
                 )
             except Exception as e:
-                _logger.warning(f'{_exc_msg} {e}')
+                _logger.warning(f'{_exc_msg}', exc_info=e)
```
ladinesa: sorry you can undo this change as it appears it produced some unwanted failures in test, I can update this in a separate mr.

ndaelman-hu:
@JFRudzinski I merged in the latest updates from `develop`. I will do a quick double-check of the interface, but the first checks indicate a complete rollout. Should be merged soon.

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: @ladinesa Can I resolve this?

ndaelman-hu:
So, one last point that keeps coming up: "encountered questions about how coordinate systems are handled across different frameworks". I also wanted to know whether our approach was consistent with community standards. This led to a comprehensive investigation of coordinate system conventions in computational materials science.

##  OPTIMADE API

In version 1.2.0 (the current stable release), OPTIMADE specifies only `cartesian_site_positions` for atomic positions, with no fractional coordinate support. The specification recommends that "the **first lattice vector SHOULD be set along x** and the second on the xy-plane" for databases that don't define an absolute Cartesian system (https://schemas.optimade.org/releases/v1.2.0/v1.2/properties/optimade/structures/lattice_vectors). This may reflect that different simulation codes may output structures with different orientations.

The development version (v1.3.0-rc.1), however, includes a new `fractional_site_positions` property alongside the existing Cartesian positions (https://github.com/Materials-Consortia/OPTIMADE/blob/develop/optimade.rst), similar to NOMAD's dual storage approach.

 ## ASE (Atomic Simulation Environment)

[ASE](https://ase-lib.org/ase/cell.html) stores cell vectors as row vectors in a 3×3 matrix, where `cell[i, j]` represents the jth Cartesian coordinate of the ith lattice vector. Positions are stored in Cartesian coordinates by default, with fractional coordinates computed on-demand via `scaled_positions()` and `cartesian_positions()` methods. Critically, ASE makes no specification about the absolute orientation of the Cartesian frame—it's **implicit and determined by how the structure was defined**. This matches NOMAD's approach exactly, which is why conversion between ASE and NOMAD is trivial with no coordinate transformations needed.

##  Spglib and the Transformation Matrix

This package is the origin of `transformation_matrix` and `origin_shift` (had to track back to git commit 8ccf1e6). These properties were introduced "to define the conventional cell" and appeared in **two different classes** with different documentation. In `GeometricSpace` (now `Representation`), the language about "custom coordinates system" vs "global coordinates system" is unclear. It suggests that any modification should refer to the "representative" system's frame. Meanwhile, in `GlobalCrystalSymmetry` they were clearly described as relating "the origin of `ModelSystem.cell` to the conventional unit cell that was used to **derive the space group symmetry**." The latter should be reserved for the `Symmetry` section.

The actual implementation showed these values come directly from spglib's `_get_spglib_origin_shift()` and `_get_spglib_transformation_matrix()` methods. Spglib's documentation
  (https://spglib.readthedocs.io/en/latest/definition.html and https://spglib.readthedocs.io/en/stable/dataset.html) interpretes **these transformations operate entirely in fractional coordinate space, not Cartesian space**. As spglib's documentation explicitly states, "the transformation matrix doesn't rotate a crystal in Cartesian coordinates, but just changes the choices of basis vectors."

Spglib thus takes an input cell and produces a standardized conventional cell that matches crystallographic conventions for the identified space group. The transformation relates **how fractional coordinates map between these cells via x_std = P x_input + p, where both cells are expressed in the same implicit Cartesian frame**. The lattice vectors transform as (a_std, b_std, c_std) = 
  (a_input, b_input, c_input) P^-1. Crucially, **Cartesian positions of atoms don't change**—they're just described with different fractional coordinates relative to different lattice vectors.

##  DFT Simulation Codes

Major DFT codes (e.g. VASP, Quantum ESPRESSO, CASTEP) handle coordinate input. All three codes accept both Cartesian and fractional coordinates in their input files. While the materials modeling community (https://mattermodeling.stackexchange.com/questions/1063) agrees that input format doesn't affect performance, fractional coordinates are strongly recommended for periodic systems because they better preserve crystallographic symmetries during geometry optimization. These symmetries are crucial for computational efficiency.

Internally, different codes likely use different representations optimized for their algorithms, but the conversion cost between representations is negligible. The takeaway is that **the community standard is to work with fractional coordinates for periodic systems** when possible, but codes are flexible about input format.

@EBB2675 explained me that in most molecular codes, the basis is also implicit, and rather arbitrary. Even when alignment is highly relevant, e.g. spectroscopy of polarized light, the code will enforce the orientation of the molecules via their positions rather than adjusting some reference frame. I had a similar experience with Gaussian regarding reference frames. Still, we decided that it would be good to add a comment for cases where the reference frame and crystal lattice diverge, or only the former is defined. This can then be added when we encounter a parsing example. 

##  Implications for NOMAD

This investigation validated the new "implicit Cartesian frame" approach that I defined here. All major frameworks (ASE, OPTIMADE, spglib) use the same model: an implicit Cartesian reference frame that isn't stored explicitly, with both lattice vectors and positions expressed in this frame. The confusion in our documentation arose from inheriting spglib's terminology about "global" vs "custom"  coordinate systems, when in reality both systems share the same Cartesian basis and differ only in their choice of lattice vectors.

Our documentation updates

- clarify that the Cartesian frame is implicit and conventional rather than stored
- simplify mathematical descriptions to avoid overwhelming users
- explain that `transformation_matrix` and `origin_shift` relate lattice vectors within the same Cartesian frame rather than transforming between different Cartesian bases

JFRudzinski:
Thank you for the summary @ndaelman-hu This all looks reasonable to me :+1:

PR #270, opened 2025-11-04
Post-HF workflows :  Coupled Cluster and Configuration Interaction

### LINKED ISSUE #26: Add post-HF methods
We should start looking into post-HF (workflow with an input HF solution) as a first step towards measuring accuracy. QMC would be the other benchmark methods.
Target codes would be [Molpro](https://www.molpro.net/manual/doku.php?id=the_closed_shell_ccsd_program) and Turbomole (quantum chemistry community), as well as [CC4S](https://manuals.cc4s.org/user-manual/algorithms/CoupledCluster.html#org1398503), and other embedded versions.

Most of these techniques are defined in terms of their orders, as well as the approach (variational, perturbative) and any restrictions on the excitations. Targets are:
- Moller-Plesset
- Coupled Cluster with potential connection to RPA and SOSEX
- (Configuration Interaction)

The aim to provide a modular and easily interpretable search.

Annotating search by properties, such as `size-extensivity` is also desirable.

### LINKED ISSUE #133: Add Hartree Fock
HF and its flavors could/should be added to ModelMethodElectronic

EBB2675:
This PR introduces post-HF workflows. It adds a generic HF+ (BeyondHF) base workflow and two concrete specializations:

    HF → CC (Coupled Cluster)

    HF → CI (Configuration Interaction)

currently, a minimal amount of information for each ModelMethod (`CoupledCluster`, `ConfigurationInteraction`) has been added to avoid complexity. Later on, the quantities will be enriched.


TLCFEM (approved):

JFRudzinski (approved):

JFRudzinski:
@ladinesa just some context: this is largely based off of the beyond_dft module code-wise, so in principal it should be fine, but have a quick check when you have the chance (note: there is no corresponding change in the parser suite, this does not affect anything there).

My plan would be merge this one first, then rebase and merge my migrate_md PR nomad-simulations, then merge my migrate_md PR nomad-parser-plugins

ladinesa (approved):

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -1183,3 +1183,192 @@ class DMFT(ModelMethodElectronic):
...
+        type=MEnum('(T)', '[T]', '(T0)', '[T0]', '(Q)'),
+        description="""
+        Label for the perturbative corrections:
+          - '(T)'   : standard perturbative triples
+          - '[T]'   : Brueckner-based or other variant
+          - '(T0)'  : approximate version of (T)
+          - '[T0]'  : approximate, typically for Brueckner references
+          - '(Q)'   : perturbative quadruples, e.g., CCSDT(Q)
+        """,
+    )
```
ndaelman-hu, line 1309: If I understand correctly, these are alternative representations of the same information, correct?
I think that within the NOMAD ecosystem, this is the recommended approach.

As we started discussing with `BaseSpinOrbitalState`, however, we might want to consider using one canonical representation for storage and recuperating the other at query time (when downloaded?). I think the same idea could apply here.

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -1183,3 +1183,192 @@ class DMFT(ModelMethodElectronic):
...
+        Spin-component scaling approach for perturbation methods:
+          - SCS   : spin-component scaled (Grimme's approach, https://doi.org/10.1002/wcms.1110)
+          - SOS   : spin-opposite scaled
+          - custom: user-defined scaling factors
+        Typically used for MP2; SCS/SOS variants also exist for some approximate CC models.
+        """,
+    )
+
+
+class CoupledCluster(ModelMethodElectronic):
```
ndaelman-hu, line 1255: Looks all good. There are few CC codes for solids,. which likely need a few more fields.
I wouldn't delay this PR for them right now, but just open an issue to revisit when we cover them.

ndaelman-hu (approved):

src/nomad_simulations/schema_packages/model_method.py
EBB2675: the current implementation is only approximately 5% of what coupled cluster needs, so yes. 

also, Denis sent a list of relevant quantitites for solids a couple of months back, feel free to send yours and i can synthesize them together :rocket:

src/nomad_simulations/schema_packages/model_method.py
EBB2675: not the same information, (T) and [T] are quite different approximations

both have the same order = 3, but different methods

EBB2675: (later on, this MEnum will be enriched with more correction types based on third order)

EBB2675: (T)', '[T]', '(T0)', '[T0]' are all different, with the order = 3

PR #275, opened 2025-11-05
Add labels to identify pull request scope

### LINKED ISSUE #274: Add labels to identify pull request scope

JosePizarro3:
@JFRudzinski please check this out. With this yaml + changing the current labels + assigning labels ONLY in pull requests, the management of the package can be done more easy. This is so that when a new release is done, we know exactly which number to pump:

> Given a version number MAJOR.MINOR.PATCH, increment the:
> 
> MAJOR version when you make incompatible API changes
> MINOR version when you add functionality in a backward compatible manner
> PATCH version when you make backward compatible bug fixes

Feel free to check and see whether some other potential labels are necessary. I think these are more than enough, but others can be defined if we find that are important later on.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,10 @@
+changelog:
+  exclude:
+    authors: [github-actions, pre-commit-ci]
+  categories:
+    - title: 💥 Incompatible API Changes
+      labels: [breaking, refactor]
+    - title: 💡 Added Functionalities (Backward Compatible)
+      labels: [feature, enhancement, DX, UX, docs, tests]
+    - title: 🛠 Backward Compatible Bug Fixes And Minor Changes (Backward Compatible)
+      labels: [dependencies, fix, datamodel, pkg]


JosePizarro3:
@ladinesa what do you think about this? It helped me to control better which numbered release to have. Right now the tags here (I don't know in other parser repos) are a bit unstructured, and with this one will save time trying to understand what number to bump.

JFRudzinski:
I have seen this @JosePizarro3 ... I just haven't had time to look into it yet, I will though

JosePizarro3:
> I have seen this @JosePizarro3 ... I just haven't had time to look into it yet, I will though

Yes, just saying to Alvin in case he wants to add it in the parsers

ladinesa:
> @ladinesa what do you think about this? It helped me to control better which numbered release to have. Right now the tags here (I don't know in other parser repos) are a bit unstructured, and with this one will save time trying to understand what number to bump.

this is automatic vesioning based on changelog? sorry no experience with this.

JosePizarro3:
It is a very dumb way of categorizing with labels pull requests. Then, when a new release is drafted and the notes are automatically generated, they are grouped under those categories in release.yml. This way it is easy to see if you need to pump the MAJOR, the MINOR, or the PATCH version (MAJOR.MINOR.PATCH).

I guess one can see that when adding those and creating a new release.

It is a super small scope change. Only the labels will change and need to be added in the open pull requests

ladinesa:
> It is a very dumb way of categorizing with labels pull requests. Then, when a new release is drafted and the notes are automatically generated, they are grouped under those categories in release.yml. This way it is easy to see if you need to pump the MAJOR, the MINOR, or the PATCH version (MAJOR.MINOR.PATCH).
> 
> I guess one can see that when adding those and creating a new release.
> 
> It is a super small scope change. Only the labels will change and need to be added in the open pull requests

ah i see, not really auto versioning will only generate labels. I am ok with this.

JFRudzinski:
is the suggestion to ONLY have these labels, or simply add the new labels and then there is a protocol that you have to assign at least one label from this group for each PR?

I am not necessarily ok with removing some of the existing labels since I use them for project management, although they should be improved.  In the latter case though, I would add a sort of category to the label, e.g., "version-refactor" so they can be easily identified

JosePizarro3:
> is the suggestion to ONLY have these labels, or simply add the new labels and then there is a protocol that you have to assign at least one label from this group for each PR?
> 
> I am not necessarily ok with removing some of the existing labels since I use them for project management, although they should be improved. In the latter case though, I would add a sort of category to the label, e.g., "version-refactor" so they can be easily identified

Only these.

But we can add more, only that I felt some of the labels (e.g., `ModelSystem` or `needs relevance check`) are too unstructured. We could group those into other category like you suggested, and agree on what number to pump in the semantic tag.

JFRudzinski:
ok, I see now, I am definitely open to refactoring the labels to help the management process. I would just like to wait a little longer until we are closer to the 1.0.0 milestone. Before that the versioning is a bit arbitrary anyway because the schema is not really being used yet.

I will add this to the 1.0.0 milestones in any case

JFRudzinski:
Thanks for the suggestion @JosePizarro3 

I am going to close this PR for now to avoid it becoming stale, as I am not ready to commit to this approach without a solution for how we can still effectively use labels more generally.

I created #317 to keep the idea in mind as we progress.

PR #283, opened 2025-11-20
assign `representative_system_index` to zero for singleton cases

### LINKED ISSUE #282: Clarification needed on `representative_system_index` behavior for singleton `ModelSystem` cases
While working on quantum-chemistry parsers (which typically produce exactly one `ModelSystem`), I noticed that the current normalization logic sets:

- `is_representative = True`

- `representative_system_index = -1`

when no representative is pre-marked, even in the singleton case.

For molecular simulations (entire QC domain), the representative system is almost always the only system.
Therefore, it would be more intuitive for downstream users and consumers if normalization produced:

- `representative_system_index = 0`

instead of -1.

My question is: 
**Should I always explicitly set `is_representative = True` from the parser side for the singleton case? (If so, maybe we can note this in the schema docstring)**

otherwise, i can update the logic from the schema side and create a PR. 

@JFRudzinski @ladinesa

EBB2675:
- For a single `ModelSystem` with no representative set, normalization now uses `representative_system_index = 0` instead of -1.

- Multi-system behavior (sentinel -1 when no rep is marked) is unchanged.

- Updated the corresponding test to match the new singleton behavior.

**Rationale**

The majority of entries typically has a single `ModelSystem.`
Storing` representative_system_index = 0` for this case is more intuitive, avoids downstream handling of -1, and aligns with how the representative system is conceptually interpreted.

### DIFF
--- a/src/nomad_simulations/schema_packages/general.py
+++ b/src/nomad_simulations/schema_packages/general.py
@@ -277,6 +277,14 @@ class Simulation(BaseSimulation, Schema):
         if len(rep_idx) == 0:
             self.model_system[-1].is_representative = True
             self.representative_system_index = -1
+
+            # - Singleton: store concrete index 0
+            # - Multi-system: keep sentinel -1 ("last element")
+            if len(self.model_system) == 1:
+                self.representative_system_index = 0
+            else:
+                self.representative_system_index = -1
+
         elif len(rep_idx) > 1:
             logger.warning(
                 'Multiple representative systems found, one allowed.'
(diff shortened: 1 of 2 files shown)


src/nomad_simulations/schema_packages/general.py
```diff
@@ -277,6 +277,14 @@ def _validate_and_set_representative_system(self) -> None:
...
             self.model_system[-1].is_representative = True
             self.representative_system_index = -1
+
+            # - Singleton: store concrete index 0
+            # - Multi-system: keep sentinel -1 ("last element")
+            if len(self.model_system) == 1:
+                self.representative_system_index = 0
+            else:
+                self.representative_system_index = -1
+
```
JFRudzinski, line 287: The logger statement below this is confusing. Actually, there are several rep systems allowed from the perspective of is_rep, but not from rep_ind, should we change the name of one of these, like is_normalized (really to be norm)...idk, not this but something

JFRudzinski (approved):
Fine to merge now and address the naming thing later, or if you have a good option for renaming (see my comment), please address now

PR #284, opened 2025-11-28
Guard ModelSystem.normalize against numpy array inputs

Bernadette-Mohr:
The check for `lattice_vectors` used to fail if the input `Quantity` contains a `np.ndarray`: `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`.
Now it handles both 2D lists and arrays. All tests pass locally.

FIXME: This still doesn't cover edge cases, scalar values, etc!

### DIFF
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -1410,7 +1410,7 @@ class ModelSystem(System, Representation):
 
         # Check and normalize periodic boundary conditions based on lattice vectors
         # Handle None, empty arrays, empty lists, and pint Quantities
-        lattice_vectors_empty = (
+        is_lattice_vectors_empty: bool = (
             self.lattice_vectors is None
             or (
                 hasattr(self.lattice_vectors, 'size') and self.lattice_vectors.size == 0
@@ -1420,7 +1420,7 @@ class ModelSystem(System, Representation):
                 and len(self.lattice_vectors) == 0
             )
         )
-        if lattice_vectors_empty and self.periodic_boundary_conditions:
+        if is_lattice_vectors_empty and self.periodic_boundary_conditions:
             logger.warning(
                 'Lattice vectors are not defined but periodic boundary conditions are set. Unsetting PBC.'
             )
--- a/src/nomad_simulations/schema_packages/model_system.py
+++ b/src/nomad_simulations/schema_packages/model_system.py
@@ -1409,7 +1409,18 @@ class ModelSystem(System, Representation):
         super().normalize(archive, logger)
 
         # Check and normalize periodic boundary conditions based on lattice vectors
-        if not self.lattice_vectors and self.periodic_boundary_conditions:
+        # Handle None, empty arrays, empty lists, and pint Quantities
+        lattice_vectors_empty = (
+            self.lattice_vectors is None
+            or (
+                hasattr(self.lattice_vectors, 'size') and self.lattice_vectors.size == 0
+            )
+            or (
+                hasattr(self.lattice_vectors, '__len__')
+                and len(self.lattice_vectors) == 0
+            )
+        )
+        if lattice_vectors_empty and self.periodic_boundary_conditions:
             logger.warning(
                 'Lattice vectors are not defined but periodic boundary conditions are set. Unsetting PBC.'
             )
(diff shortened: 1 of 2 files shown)


src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1409,7 +1409,18 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
         super().normalize(archive, logger)
 
         # Check and normalize periodic boundary conditions based on lattice vectors
-        if not self.lattice_vectors and self.periodic_boundary_conditions:
+        # Handle None, empty arrays, empty lists, and pint Quantities
+        lattice_vectors_empty = (
```
ndaelman-hu: I like this variable name, communicates the intent much clearer.
Since we denote most `bool` variables with the prefix `is_` do you mind adding it? Also add type hinting pls.

ndaelman-hu (changes requested):
Thx for the patch, is even more extensive that I would've first gone for!

What worries me is that our tests failed to catch this. Could you pls also take a look at `test_normalize` under `test_model_system.py` and see to extend it (or split out this to a separate test)? Then I consider this fixed.

tests/test_model_system.py
```diff
@@ -516,6 +516,54 @@ def test_normalize(self):
...
+        sys3.positions = np.array([[0, 0, 0]]) * ureg.angstrom
+        sys3.periodic_boundary_conditions = [True, False, False]
+        sys3.lattice_vectors = [[], []]  # empty list
+        sys3.particle_states.append(AtomsState(chemical_symbol='H'))
+        caplog.clear()
+        with caplog.at_level(logging.WARNING):
+            sys3.normalize(EntryArchive(), logger=logger)
+
+        assert sys3.periodic_boundary_conditions == []
+
```
ndaelman-hu: I suggest using `pytest.mark.paramatrize`, then the testing body will be much simpler.

Regarding _cases covered_, I would also add a cover for one that does not get reset, combos with `None` or `periodic_boundary_conditions` as a `np.array`.
The values inside `periodic_boundary_conditions` don't matter much, even `[False]` will pass.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1420,7 +1420,7 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
                 and len(self.lattice_vectors) == 0
             )
         )
-        if lattice_vectors_empty and self.periodic_boundary_conditions:
+        if is_lattice_vectors_empty and self.periodic_boundary_conditions:
```
ndaelman-hu, line 1422: Looking at the test, does `self.periodic_boundary_conditions` pass when it's a `np.array`?

ndaelman-hu (commented):
2 minor comments. maybe we should give `periodic_boundary_conditions` the same thoroughness

ndaelman-hu (approved):
Thank you for adding the tests! So, from the test cases I can tell that PBC being a `np.array` does not pose any issues :)

Bernadette-Mohr:
> Thank you for adding the tests! So, from the test cases I can tell that PBC being a `np.array` does not pose any issues :)

I agree. Since the tests pass, I think we can go withut adding additional safety measures in `model_system.normalize` as long as we don't encounter actual errors at some point in the future. In that case, we still can do something like:
```
is_pbc_set: bool = (
    self.periodic_boundary_conditions is not None
    and (
        (hasattr(self.periodic_boundary_conditions, 'size') and self.periodic_boundary_conditions.size > 0)
        or (hasattr(self.periodic_boundary_conditions, '__len__') and len(self.periodic_boundary_conditions) > 0)
        or self.periodic_boundary_conditions  # fallback for non-sequence truthy values
    )
)
if is_lattice_vectors_empty and is_pbc_set:
```

ndaelman-hu:
> > Thank you for adding the tests! So, from the test cases I can tell that PBC being a `np.array` does not pose any issues :)
> 
> I agree. Since the tests pass, I think we can go withut adding additional safety measures in `model_system.normalize` as long as we don't encounter actual errors at some point in the future. In that case, we still can do something like:
> 
> ```
> is_pbc_set: bool = (
>     self.periodic_boundary_conditions is not None
>     and (
>         (hasattr(self.periodic_boundary_conditions, 'size') and self.periodic_boundary_conditions.size > 0)
>         or (hasattr(self.periodic_boundary_conditions, '__len__') and len(self.periodic_boundary_conditions) > 0)
>         or self.periodic_boundary_conditions  # fallback for non-sequence truthy values
>     )
> )
> if is_lattice_vectors_empty and is_pbc_set:
> ```

I think because we are looking for a positive. The strangeness mostly comes from properly filtering / defining the negative (i.e. empty) cases.

We should eventually extend the method you built as a general utility. Feel free to open an issue for that. As far as the bug here is concerned, I think you patched it sufficiently.

PR #288, opened 2025-12-03
Add pseudo potential

ndaelman-hu:
Some users have been requesting pseudopotential support. This was available in the runschema and should therefore be ported anyhow. The schema was further extended to include additional metadata with codes like CASTEP and QE in mind.
It also leverages other schema-generic pieces for relativity and DFT annotation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -172,3 +172,7 @@ todos*
 questions*
 warnings*
 important*
+
+# Claude Code artifacts
+.claude/agents/
+*_contributions_*.md
--- a/src/nomad_simulations/schema_packages/atoms_state.py
+++ b/src/nomad_simulations/schema_packages/atoms_state.py
@@ -1134,7 +1134,7 @@ class AtomsState(ParticleState):
     electronic_state = SubSection(sub_section=ElectronicState.m_def)
 
     pseudopotential = Quantity(
-        type=Reference(SectionProxy('Pseudopotential')),
+        type=Reference(SectionProxy('nomad_simulations.schema_packages.numerical_settings.Pseudopotential')),
         description="""
         Reference to the pseudopotential used for this atomic species in plane-wave DFT
         calculations. The referenced `Pseudopotential` section is defined in `numerical_settings`
--- a/src/nomad_simulations/schema_packages/atoms_state.py
+++ b/src/nomad_simulations/schema_packages/atoms_state.py
@@ -1134,7 +1134,11 @@ class AtomsState(ParticleState):
     electronic_state = SubSection(sub_section=ElectronicState.m_def)
 
     pseudopotential = Quantity(
-        type=Reference(SectionProxy('nomad_simulations.schema_packages.numerical_settings.Pseudopotential')),
+        type=Reference(
+            SectionProxy(
+                'nomad_simulations.schema_packages.numerical_settings.Pseudopotential'
+            )
+        ),
         description="""
         Reference to the pseudopotential used for this atomic species in plane-wave DFT
         calculations. The referenced `Pseudopotential` section is defined in `numerical_settings`
--- a/src/nomad_simulations/schema_packages/atoms_state.py
+++ b/src/nomad_simulations/schema_packages/atoms_state.py
@@ -6,7 +6,7 @@ import ase
 import numpy as np
 import pint
 from nomad.datamodel.metainfo.basesections.v2 import Entity
-from nomad.metainfo import MEnum, Quantity, SectionProxy, SubSection
+from nomad.metainfo import MEnum, Quantity, Reference, SectionProxy, SubSection
 from nomad.units import ureg
 
 if TYPE_CHECKING:
@@ -1132,6 +1132,16 @@ class AtomsState(ParticleState):
     )
 
     electronic_state = SubSection(sub_section=ElectronicState.m_def)
+
+    pseudopotential = Quantity(
+        type=Reference(SectionProxy('Pseudopotential')),
+        description="""
+        Reference to the pseudopotential used for this atomic species in plane-wave DFT
+        calculations. The referenced `Pseudopotential` section is defined in `numerical_settings`
+        and contains metadata such as pseudopotential type (PAW, ultrasoft, norm-conserving),
+        cutoff energy, and XC functional used to generate the pseudopotential.
+        """,
+    )
 
     @log
     def get_label(self) -> str | None:
--- a/src/nomad_simulations/schema_packages/basis_set.py
+++ b/src/nomad_simulations/schema_packages/basis_set.py
@@ -333,9 +333,20 @@ class AtomCenteredFunction(ArchiveSection):
 
 class EffectiveCorePotential(BasisSetComponent):
     """
-    TREXIO-style ECP storage:
+    TREXIO-style ECP storage for quantum chemistry calculations with Gaussian basis sets.
+
+    Effective Core Potentials (ECPs) replace core electrons with analytical potentials
+    expressed as sums of Gaussian functions. This representation is used in quantum chemistry
+    codes (Gaussian, ORCA, Molpro, Q-Chem, NWChem) to reduce computational cost while
+    maintaining accuracy for valence electrons.
+
+    Storage format (TREXIO convention):
       - Per-nucleus arrays: z_core[nucleus], max_ang_mom_plus_1[nucleus]
       - Flat list of projector items: size ecp_num with nucleus_index, ang_mom, exponent, coefficient, power
+
+    Note: This class stores analytical ECPs for Gaussian basis set codes. It does NOT represent
+    pseudopotentials used in plane-wave DFT codes (PAW, ultrasoft, norm-conserving). For metadata
+    about plane-wave pseudopotentials, see the `Pseudopotential` class in numerical_settings.py.
     """
 
     # Optional human label
--- a/src/nomad_simulations/schema_packages/basis_set.py
+++ b/src/nomad_simulations/schema_packages/basis_set.py
@@ -1,6 +1,6 @@
 import itertools
 from collections.abc import Iterable
-from typing import TYPE_CHECKING, Any, Optional
+from typing import TYPE_CHECKING, Any
 
 from scipy import constants as const
 
--- a/src/nomad_simulations/schema_packages/numerical_settings.py
+++ b/src/nomad_simulations/schema_packages/numerical_settings.py
@@ -15,6 +15,7 @@ if TYPE_CHECKING:
     from nomad.metainfo import Section
     from structlog.stdlib import BoundLogger
 
+from nomad_simulations.schema_packages.model_method import RelativityModel
 from nomad_simulations.schema_packages.model_system import ModelSystem
 from nomad_simulations.schema_packages.utils import log
 
@@ -1143,15 +1144,11 @@ class Pseudopotential(NumericalSettings):
         """,
     )
 
-    relativistic = Quantity(
-        type=MEnum('non-relativistic', 'scalar-relativistic', 'full-relativistic', 'unavailable'),
-        shape=[],
+    relativistic_treatment = SubSection(
+        sub_section=RelativityModel.m_def,
         description="""
-        Level of relativistic effects included in pseudopotential generation:
-        - `'non-relativistic'`: No relativistic corrections (light elements)
-        - `'scalar-relativistic'`: Scalar relativistic effects (spin-orbit coupling neglected)
-        - `'full-relativistic'`: Complete relativistic treatment (heavy elements, spin-orbit)
-        - `'unavailable'`: Relativistic treatment not specified or unknown
+        Relativistic treatment used during pseudopotential generation.
+        Does not imply anything about how treatment of the valence electrons.
         """,
     )
 
--- a/src/nomad_simulations/schema_packages/numerical_settings.py
+++ b/src/nomad_simulations/schema_packages/numerical_settings.py
@@ -1032,6 +1032,83 @@ class FrozenCore(NumericalSettings):
     )
 
 
+class Pseudopotential(NumericalSettings):
+    """
+    Section containing the parameters for pseudopotentials used in electronic structure calculations.
+    Pseudopotentials are used to approximate the potential of the core electrons and the nucleus,
+    allowing for a more efficient treatment of valence electrons.
+    """
+
+    name = Quantity(
+        type=str,
+        shape=[],
+        description="""
+        Native code name of the pseudopotential.
+        """,
+    )
+
+    type = Quantity(
+        type=MEnum('US V', 'US MBK', 'PAW'),
+        shape=[],
+        description="""
+        Pseudopotential classification.
+
+        | abbreviation | description | DOI |
+        | ------------ | ----------- | --------- |
+        | `'US'`       | Ultra-soft  | |
+        | `'PAW'`      | Projector augmented wave | |
+        | `'V'`        | Vanderbilt | https://doi.org/10.1103/PhysRevB.47.6728 |
+        | `'MBK'`      | Morrison-Bylander-Kleinman | https://doi.org/10.1103/PhysRevB.41.7892 |
+        """,
+    )
+
+    norm_conserving = Quantity(
+        type=bool,
+        shape=[],
+        description="""
+        Denotes whether the pseudopotential is norm-conserving.
+        """,
+    )
+
+    cutoff = Quantity(
+        type=np.float64,
+        shape=[],
+        unit='joule',
+        description="""
+        Minimum recommended spherical cutoff energy for any plane-wave basis set
+        using the pseudopotential.
+        """,
+    )
+
+    xc_functional_name = Quantity(
+        type=str,
+        shape=['*'],
+        description="""
+        Name of the exchange-correlation functional used to generate the pseudopotential.
+        Follows the libxc naming convention.
+        """,
+    )
+
+    l_max = Quantity(
+        type=np.int32,
+        shape=[],
+        description="""
+        Maximum angular momentum of the pseudopotential projectors.
+        """,
+    )
+
+    lm_max = Quantity(
+        type=np.int32,
+        shape=[],
+        description="""
+        Maximum magnetic momentum of the pseudopotential projectors.
+        """,
+    )
+
+    def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
+        super().normalize(archive, logger)
+
+
 class IntegralDecomposition(ArchiveSection):
     """
     A general class for integral decomposition techniques that approximate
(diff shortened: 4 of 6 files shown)


ndaelman-hu:
@EBB2675 I used the sections `RelativityModel` and `XCFunctional`, which you recently added.
Pls also remark on how you see their wider spread usage, as well as import handling.

ndaelman-hu:
@JFRudzinski if you want, I can test this schema out right-away in a few parsers. The first candidates would be: VASP, FHIaims, CASTEP, and QE. Aside from the obvious parsing extension, this has a 2-pronged advantage:

- some codes distinguish multiple cutoffs for different levels of precision. I already indicated that this should be extended on their side. This could be the actual test.
- I might also get feedback from the CASTEP devs. This isn't necessarily contingent on us extending the parser, but it would direct their focus.

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -1032,6 +1032,170 @@ class FrozenCore(NumericalSettings):
...
+    name = Quantity(
+        type=str,
+        shape=[],
+        description="""
+        Native code name of the pseudopotential.
+        """,
+    )
+
+    type = Quantity(
+        type=MEnum('US V', 'US MBK', 'PAW'),
```
EBB2675: The description table lists 'US', 'PAW', 'V', 'MBK', and enum is `MEnum('US V', 'US MBK', 'PAW')`

This mismatch might come across as confusing. Right now it’s not clear whether:

- `type` is meant as combination of ultrasoft + builder (US V, US MBK), or
-	you conceptually want a primary type (PAW, US, NC) and a separate “family/builder” classification (V, MBK, …).

I’d at least align the table with the actual enum values or explicitly say “'US V' = Vanderbilt ultrasoft; 'US MBK' = Morrison-Bylander-Kleinman ultrasoft”.

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -1032,6 +1032,170 @@ class FrozenCore(NumericalSettings):
     )
 
 
+class Pseudopotential(NumericalSettings):
+    """
+    Section containing metadata for pseudopotentials used in plane-wave DFT calculations.
+
+    Pseudopotentials approximate the potential of core electrons and the nucleus, enabling
+    efficient treatment of valence electrons in codes like VASP, Quantum ESPRESSO, and CASTEP.
+    Common types include PAW (Projector Augmented Wave), ultrasoft (US), and norm-conserving (NC).
```
EBB2675:
```suggestion
    Common types include PAW (Projector Augmented Wave), ultrasoft (US), and norm-conserving (NC).
    Norm-conserving pseudopotentials are represented by setting `norm_conserving = True` and choosing the appropriate value in `type`.
```

EBB2675: The docstring talks about “norm-conserving (NC), ultrasoft (US), and PAW” as three types. 

In the schema, norm-conservation is encoded only via the `norm_conserving: bool` flag, while the type enum distinguishes the representation and does not contain an explicit NC value. It might help to add one sentence above that makes this explicit so future contributors don’t expect “NC” to appear as a separate entry in the `type` enumeration.

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -1032,6 +1032,170 @@ class FrozenCore(NumericalSettings):
...
+        description="""
+        Maximum angular momentum of the pseudopotential projectors.
+        """,
+    )
+
+    lm_max = Quantity(
+        type=np.int32,
+        shape=[],
+        description="""
+        Maximum magnetic momentum of the pseudopotential projectors.
```
EBB2675:
```suggestion
        Maximum magnetic quantum number of the pseudopotential projectors.
```

EBB2675 (approved):
looks great, and also the usage of `RelativityModel` looks perfect 👍 

i just added a couple of super minor docstring clarifications

JFRudzinski (commented):
It looks good...

I am just wondering a bit about the categorization of pseudo potentials as numerical settings vs. Model method.

I would have thought the identity of the pseudo potential should live as a model method, as they are used to build the Hamilton components. 

Things like cutoff and such, ie, how you evaluate the potentials, I would agree are numerical settings

There is a clear analogy with my force field schema, please check and then we can discuss if it's useful to align these better

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: Good catch! Typically people care more so about the primary type, not that much the numerical / builder representation.
Adding the latter would also require some added verification logic. Better to V and MDK out.

On that note, I also re-evaluated NC, as it is sometimes combined with PAW. In particular when PAW pseudopotentials are used for GW.

ndaelman-hu:
> It looks good...
> 
> I am just wondering a bit about the categorization of pseudo potentials as numerical settings vs. Model method.
> 
> I would have thought the identity of the pseudo potential should live as a model method, as they are used to build the Hamilton components.
> 
> Things like cutoff and such, ie, how you evaluate the potentials, I would agree are numerical settings
> 
> There is a clear analogy with my force field schema, please check and then we can discuss if it's useful to align these better

We discussed and agreed that the PP can stay in `numerical_settings.py`.

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -1032,6 +1032,197 @@ class FrozenCore(NumericalSettings):
     )
 
 
+class Pseudopotential(NumericalSettings):
+    """
+    Section containing high-level metadata (type, cutoff energy, XC functional) that identifies which pseudopotential was used.
+    Pseudopotentials approximate the potential of core electrons and the nucleus, enabling
+    efficient treatment of valence electrons in codes like VASP, Quantum ESPRESSO, and CASTEP.
```
ndaelman-hu:
```suggestion
    efficient treatment of valence electrons in plane-wave codes.
```

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -1032,6 +1032,197 @@ class FrozenCore(NumericalSettings):
...
+
+    NC-PAW: PAW variant with norm-conserving partial waves. While more expensive than standard
+    PAW, NC-PAW provides better scattering properties for high-energy states, making it more
+    suitable for calculations requiring accurate unoccupied states.
+
+    NC-PAW-GW: NC-PAW optimized specifically for GW and BSE calculations. These include additional
+    projectors at higher energies to accurately describe quasiparticle states far above the Fermi
+    level. Standard PAW and US pseudopotentials systematically underestimate scattering into
+    high-energy unoccupied states, which is critical for GW many-body perturbation theory.
+
```
ndaelman-hu: should go under `type` description

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -1032,6 +1032,197 @@ class FrozenCore(NumericalSettings):
...
+        type=str,
+        shape=[],
+        description="""
+        Electronic configuration used to generate the pseudopotential (e.g., "3s1 3d0.5" or "3p6 3d7 4s1").
+        Documents the valence electron occupations used during generation.
+        The configuration string may only show the outermost valence orbitals explicitly.
+        """,
+    )
+
+    norm_conserving = Quantity(
```
ndaelman-hu:
```suggestion
    is_norm_conserving = Quantity(
```

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -1032,6 +1032,197 @@ class FrozenCore(NumericalSettings):
...
+        unit='joule',
+        description="""
+        Recommended cutoff energy for the plane-wave basis set using this pseudopotential.
+
+        When multiple cutoff recommendations exist in the source (e.g., coarse/medium/fine
+        or min/max), store the most representative value. Use the medium/standard setting
+        when in doubt. Codes with sophisticated multi-level cutoff systems should extend this at
+        the parser-specific schema level.
+        """,
+    )
```
ndaelman-hu: add `cutoff_target` to clarify the reference?

.gitignore
ndaelman-hu, line 1: @JFRudzinski are you fine with us adding the default AI folders to the gitignore?

src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -333,9 +333,20 @@ def __init__(self, *args, **kwargs):
 
 class EffectiveCorePotential(BasisSetComponent):
     """
-    TREXIO-style ECP storage:
+    TREXIO-style ECP storage for quantum chemistry calculations with Gaussian basis sets.
+
+    Effective Core Potentials (ECPs) replace core electrons with analytical potentials
+    expressed as sums of Gaussian functions. This representation is used in quantum chemistry
+    codes (Gaussian, ORCA, Molpro, Q-Chem, NWChem) to reduce computational cost while
```
ndaelman-hu, line 340:
```suggestion
    codes to reduce computational cost while
```

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu, line 1: Note: this design appears semantically most desirable, but should first be tested in at least 1 parser

ndaelman-hu (commented):
A few of my own thoughts. Will be implemented, then this PR is ready for merging.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: applied

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: applied

src/nomad_simulations/schema_packages/atoms_state.py
ndaelman-hu: being tested on VASP: https://github.com/FAIRmat-NFDI/nomad-parser-plugins-simulation/pull/129

.gitignore
JFRudzinski: yes

JFRudzinski (commented):
At a glance it looks good for me :+1:  Someone more of an expert should review in more detail though

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -1032,6 +1032,215 @@ class FrozenCore(NumericalSettings):
...
+
+    cutoff_target = Quantity(
```
EBB2675: These code-specific strings for plane-wave energy cutoffs are not scientific concepts, but naming conventions in specific codes. Users of PP libraries do not interpret them scientifically, so they are not really necessary for interoperability. Parser authors can resolve these labels internally and store a clean, consistent value in cutoff.

(If in the future a PP format appears that includes multiple scientifically distinct cutoff values, then a slightly different design would be more suitable, for example:

`cutoffs = SubSection(PPCutoff, repeats=True)` )

In particular:

- Does **ENMAX** mean something different than “recommended cutoff”? → No

- Does **ecutwfc** string have a different unit or meaning? → No

- Does knowing the original label help downstream analyses? → No


I am flexible on this, but if we move away from a code-agnostic design, I would need to quickly revisit my quantum-chemistry related implementations and add similar code-specific conventions wherever they are important, similar to these cutoff indicating strings 👍

EBB2675 (changes requested):
I’d gently lean towards dropping `cutoff_target`: labels like ENMAX/ecutwfc are code-specific rather than scientific, and parsers can already map them internally to a single, clean cutoff value. But I am flexible if others see a strong use case.

if we go for this kind of design, I can have a quick PR on QC schema support, to include analogous, community-known, code-specific conventions as well 👍

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: I quite like your `cutoffs = SubSection(PPCutoff, repeats=True)` suggestion. I guess I still need a representative cutoff for querying: people will definitely want to search pseudopotentials by element and cutoff, those are its main parameters. Or we use nested queries? I could try that.

Either way,  it's not that much a question of "if". Codes like VASP and CP2K provide cutoff ranges or multiple (precision-dependent) cutoffs, respectively, instead of a single representative cutoff. The question I was trying to answer isn't really "Does ENMAX mean something different than “recommended cutoff”? → No", but rather "Which cutoff does NOMAD consider representative: ENMIN or ENMAX?". I guess you did pick up on that to some extent. 

I do not have a much of view on how the choice of `cutoff_target` or `PPCutoff` affects the rest of the schema, though. I hadn't thought about that scope yet, thx for bringing it up.

EBB2675: cant they query by code name VASP etc instead of cutoff_target?

EBB2675: i have also checked my previous VASP simulations from my phd, and there were also certain cases in which i didnt resort to any ENMAX or ENMIN, but choose another cutoff

EBB2675: from my understanding, these recommended cutoffs are provided to guide the user, but at the end, isnt the calculation done by what is set in `ENCUT` ?

if the user does not set `ENCUT` explicitly in the `INCAR`, it indeed takes a default value from POTCAR's largest ENMAX, but still the OUTCAR prints the ENCUT, which is the final cutoff used in any scenario

EBB2675:
To keep the schema code-agnostic while preserving queryability, we can model cutoffs as *typed physical concepts* rather than code-native labels like `ENMAX` or `ecutwfc`.

### a proposed schema shape, iteration 1 by EBB:

```python
class PPCutoff(ArchiveSection):
    # What physical expansion this cutoff controls
    cutoff_kind = Quantity(
        type=MEnum(
            "wavefunction",     # e.g. QE ecutwfc, VASP ENCUT
            "charge_density",   # e.g. QE ecutrho (often ~4x for NC, higher for US)
            "augmentation",     # PAW augmentation / projector-related cutoffs (if applicable)
            "response",         # response-function basis cutoffs (e.g. ENCUTGW-like)
            "other",
        )
    )

    # What the value means in context
    cutoff_role = Quantity(
        type=MEnum(
            "used",             # actually used in the run (runtime value)
            "recommended",      # vendor/library recommended default
            "recommended_min",  # lower bound recommendation
            "recommended_max",  # upper bound recommendation
            "standard",         # library "standard" recommendation (SSSP-like)
            "stringent",        # library "stringent" recommendation
            "soft",             # explicitly soft setting
            "hard",             # explicitly hard setting
        )
    )

    value = Quantity(type=np.float64, unit="joule")
```

And in `Pseudopotential`:

```python
cutoffs = SubSection(sub_section=PPCutoff.m_def, repeats=True)
```

### here i include some example code-to-semantics mappings

- **VASP `ENMAX`**  
  maps to: `cutoff_kind="wavefunction"`, `cutoff_role="recommended"`  
  (or `cutoff_role="recommended_max"` if we explicitly treat it as an upper recommended bound)

- **VASP `ENMIN`**  
  maps to: `cutoff_kind="wavefunction"`, `cutoff_role="recommended_min"`

- **QE `ecutwfc`**  
  maps to: `cutoff_kind="wavefunction"`, `cutoff_role="used"` (runtime input)

- **QE `ecutrho`**  
  maps to: `cutoff_kind="charge_density"`, `cutoff_role="used"` (runtime input)

### with this kind of skeleton, i am able to do the following as a user:

- “Show me PAW pseudopotentials with **recommended wavefunction** cutoff < X”
- “Show me entries where **used wavefunction** cutoff < **recommended**”
- “Show me pseudopotentials with **recommended_min**/**recommended_max** ranges and filter by max”


@ndaelman-hu @JFRudzinski

ndaelman-hu:
> To keep the schema code-agnostic while preserving queryability, we can model cutoffs as typed physical concepts rather than code-native labels like ENMAX or ecutwfc.

I like the general layout. This is basically a more grounded version of the OG `PhysicalProperty` and a pattern that has been applied in a few other places of the schema too.

I just have some questions / feedback regarding the `MEnum` values:

- `cutoff_kind`: would change "other" to "undecided" or "unknown", as it might be very very hard to establish for each code.
- `cutoff_role`:
  - would remove "used", as this is already extracted to `PlaneWaveBasisSet`. Or I define a more general `Cutoff` section. 
  - what is the distinction between "standard, stringent", "hard", and "soft"? I guess the first two have to do with the precision settings. Unfortunately, precision terminology is very code-dependent. We'd need a special research effort to standardize that. Hard and soft instead refer to kinds of pseudopotentials.

EBB2675:
On `cutoff_kind`:
Yes, I agree that "other" is too vague. Renaming it to "unknown" (or "undecided") makes more sense semantically, since the ambiguity usually comes from lack of information rather than a genuinely different physical meaning :+1: 

On `cutoff_role = "used"`:
Agreed. If the actually used cutoff is already captured via `PlaneWaveBasisSet`, then it should not be duplicated here. In that case, `PPCutoff` should strictly represent pseudopotential-provided or library-provided guidance, not runtime settings :+1: please make sure to make this clear in the main docstring

About the MEnums:
I only included standard/stringent and hard/soft as examples to show how extendable the pattern is for the future, not as a strong claim that we already have a cross-code standard for these categories, so feel free to ditch them :+1:

ndaelman-hu:
@EBB2675 @JFRudzinski I applied Esma's suggestions. I just didn't get around to testing the nested queries yet.
Given the current timeframe and our aim to focus PR scope, I would open this as a follow-up issue.

Feel free to review, else I merge after lunch.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: This has been discussed in a different thread. We clarified that this PR aims to describe PP metadata, not the general basis set. Am resolving this thread.

EBB2675:
`PPCutoff` looks great

a quick question: how can i refer to the used cutoffs for different properties: wf, charge density? What does the `PlaneWaveBasisSet` record? is it only for the wavefunction cutoff?

ndaelman-hu:
> `PPCutoff` looks great
> 
> a quick question: how can i refer to the used cutoffs for different properties: wf, charge density? What does the `PlaneWaveBasisSet` record? is it only for the wavefunction cutoff?

Yeah, after pushing, I realized that I hadn't updated the basis set yet. Had been too engrossed in the query-side of things.
Will get back to you on this today.

ndaelman-hu:
> > `PPCutoff` looks great
> > a quick question: how can i refer to the used cutoffs for different properties: wf, charge density? What does the `PlaneWaveBasisSet` record? is it only for the wavefunction cutoff?
> 
> Yeah, after pushing, I realized that I hadn't updated the basis set yet. Had been too engrossed in the query-side of things. Will get back to you on this today.

I checked, but the distinction between the electronic structure property is supposed to be made at a higher level, in `BasisSetContainer`. There is a note there stating:

>  TODO: add reference to `electronic_structure`,
>  specifying to which electronic structure representation the basis set is applied
>  e.g. wavefunction, density, grids for subroutines, etc.

Given the current changes to the schema, this might be re-evaluated.
This means evaluating the whole basis set structure, which is beyond the scope of this PR.

All comments have been addressed, I'm merging this.

PR #292, opened 2025-12-04
migrate force calculations

### LINKED ISSUE #289: Migrate `ForceCalculations`
Based on our discussion about the semantics of _numerical settings_, we concluded that `ForceCalculations` should be moved from `numerical_settings.py` to `force_field.py`.

One potential target is `ForceField`. If modularity is important, they could be kept separate though. Regardless, atm there is no connection between both sections!

JFRudzinski:


src/nomad_simulations/schema_packages/force_field.py
```diff
@@ -1191,6 +1275,14 @@ class ForceField(ModelMethod):
...
     )
 
+    force_calculations = SubSection(
+        sub_section=ForceCalculations.m_def,
+        repeats=True,
+        description="""
+        Parameters describing how this ForceField model is evaluated during a simulation.
+        """,
+    )
+
```
ndaelman-hu: If `force_calculations` are accessed via `ForceField`, then I don't think they need to inherit from `NumericalSettings` directly: those are access directly under `model_method.numerical_settings`

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu, line 1: This file was just moved verbatim, correct?

tests/test_force_field.py
ndaelman-hu, line 1: I consider this test quite redundant, could be removed.

ndaelman-hu (approved):
Pls double-check the `ForceCalcualtions` inheritance. Test is redundant here.
Approval as there's no need for a follow-up review, unless you add completely different tests.

src/nomad_simulations/schema_packages/numerical_settings.py
JFRudzinski: yes

src/nomad_simulations/schema_packages/force_field.py
JFRudzinski: yes, that was stupid, I just remove the subsection but leave the inheritance

tests/test_force_field.py
JFRudzinski: removing it

src/nomad_simulations/schema_packages/force_field.py
ndaelman-hu: It depends on how you want it accessed. I  just wouldn't have to parallel ways, as it may confuse people.

PR #293, opened 2025-12-05
Remove empty normalize functions

JFRudzinski:
**Remove empty `normalize()` functions that only call `super().normalize()`**

Removes 17 boilerplate `normalize()` functions across 8 files that contained only `super().normalize()` calls without additional logic. Retains the `normalize()` function in `MuffinTinRegion` (basis_set.py) which has the `@check_normalized` decorator that enforces specific execution order. Includes formatting fixes for trailing whitespace and blank lines.

### DIFF
--- a/src/nomad_simulations/schema_packages/basis_set.py
+++ b/src/nomad_simulations/schema_packages/basis_set.py
@@ -967,10 +967,7 @@ class MuffinTinRegion(BasisSetComponent, Mesh):
         super().__init__(*args, **kwargs)
         self.mt_r_min = None
 
-    @check_normalized
-    def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
-        super().normalize(archive, logger)
-        # TODO: add spherical specification, once supported in `Grid`
+    # TODO: add spherical specification in normalization, once supported in `Grid`
 
 
 class BasisSetContainer(NumericalSettings):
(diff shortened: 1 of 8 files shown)


src/nomad_simulations/schema_packages/basis_set.py
```diff
@@ -967,10 +967,7 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.mt_r_min = None
 
-    @check_normalized
-    def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
-        super().normalize(archive, logger)
-        # TODO: add spherical specification, once supported in `Grid`
```
ndaelman-hu, line 973: Note that this normalization has a decorator (as does the initializer). They are here to enforce a specific execution of the normalizer. I'd be cautious removing them here. Let us double-check this case (pls remind me).

ndaelman-hu (commented):
I would leave 1 normalization still in. Alternatively, we investigate it's role in the normalization chain for LAPW.

src/nomad_simulations/schema_packages/basis_set.py
JFRudzinski: reverted

ndaelman-hu (approved):
Thank you for reverting. If you want me to re-optimize this, I can add it as an issue, but I consider it very low-priority.

PR #318, opened 2026-01-15
Fix normalization order for KSpace subsections (KMesh and KLinePath)

### LINKED ISSUE #126: Minor fixes
While I was debugging using the Wannier90 parser, I got some errors to be fixed (see [nomad-parser-wannier90#3](https://github.com/FAIRmat-NFDI/nomad-parser-wannier90/pull/3)):

- [ ] Change order of normalization for 1st `KSpace` and then its subsections (`KMesh` and `KLinePath`).
- [x] Fix warning in `spectral_profile.py` adding condition for `if self.name == 'ElectronicDensityOfStates'` for total DOS.
- [ ] Overwrite FermiLevel and call for `ElectronicBandStructure.normalize()` once DFT+TB workflow is resolved.
- [x] No idea why, but in the frontend, the `SinglePoint.inputs` and `outputs` section are not shown.
- [x] Fix README
- [x] Fix equality operator
- [x] Move testing files to subfolders to mimic the src structure
- [x] Delete copyright at the beginning of each file

ndaelman-hu:
## Purpose

Fixes issue #126 where `KMesh` and `KLinePath` subsections could not access `reciprocal_lattice_vectors` during normalization because NOMAD normalizes subsections before their parent sections. This bug manifested in workflows combining DFT codes (providing crystal structure) with post-processing tools like Wannier90.

## Scope

**Included**

- Add `normalize_reciprocal_lattice_vectors()` method to `KSpace` class with `@log` decorator
- Update `KSpace.normalize()` to resolve `reciprocal_lattice_vectors` before calling `super().normalize()`
- Update `KMesh.normalize()` and `KLinePath.normalize()` to call parent's `normalize_reciprocal_lattice_vectors()` early
- Fix unit preservation by accessing `reciprocal_lattice_vectors` directly from parent instead of via `m_xpath`
- Add comprehensive unit tests for normalization order independence

**Out of scope**

- Changes to other numerical settings classes
- Modifications to Wien2k, Wannier90, or other parser implementations
- Schema changes beyond normalization logic

## Context & Links

- Closes #126
- Related to Wannier90 parser development: https://github.com/FAIRmat-NFDI/nomad-parser-wannier90/pull/3
- Tested with entry_id: `zDRjaWxRjMLLsRhcoq7VbrU-ki5o`

## Reviewer Notes

**Key design decisions:**

1. **`@log` decorator usage**: The `normalize_reciprocal_lattice_vectors()` method uses `@log` to keep the method signature clean while providing logger access via `self.method_name.__annotations__['logger']`

2. **Normalization order**: `KSpace.normalize()` calls `normalize_reciprocal_lattice_vectors()` BEFORE `super().normalize()` so that when the framework triggers subsection normalization, the vectors are already available

3. **Direct attribute access**: Changed from `m_xpath('m_parent.reciprocal_lattice_vectors')` to `self.m_parent.reciprocal_lattice_vectors` to preserve units (pint quantities)

4. **Idempotency**: `normalize_reciprocal_lattice_vectors()` checks if vectors are already set and returns early, allowing safe multiple calls from both parent and children

## Status

- [x] Ready for review

## Breaking Changes

- [x] None

This is a backward-compatible fix that only changes the internal normalization order. All existing tests pass, and the behavior for cases where `reciprocal_lattice_vectors` were successfully resolved remains unchanged.

## Dependencies / Blockers

- [x] None

## Testing / Validation

- [x] Unit tests: Added 2 new tests (`test_normalization_order_with_kspace` for both `KMesh` and `KLinePath`) that explicitly test subsection normalization before parent
- [x] Integration tests: All 34 existing numerical_settings tests pass (32 existing + 2 new)
- [x] Manual testing: Tested with Wannier90 parser on entry_id `zDRjaWxRjMLLsRhcoq7VbrU-ki5o` - no regressions, normalization completes successfully
- [x] Code quality: `ruff format` and `ruff check` pass

**Test coverage:**
- Normalization order independence (new tests)
- Backward compatibility with existing tests
- Edge cases: missing ModelSystem, missing lattice vectors (warnings logged, no crashes)


ndaelman-hu:
**Test inconsistency fixed**: KLinePath test now sets `reciprocal_lattice_vectors=None` to properly test resolution from ModelSystem.

**Logger access pattern**: The `logger = self.method_name.__annotations__['logger']` pattern is used 56 times across the codebase (atoms_state.py, model_system.py, spectral_profile.py, workflow/*.py). This is the established project convention for `@log` decorated methods.

@ladinesa - is this pattern acceptable or should we use a different approach?

ndaelman-hu:
@ladinesa could you pls provide feedback to this PR?

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -482,6 +482,13 @@ def resolve_k_line_density(
         return None
 
     def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
+        # Ensure parent KSpace has normalized reciprocal_lattice_vectors first
+        k_space = self.m_parent
+        if k_space is not None and hasattr(
```
ladinesa: maybe look at the def used by the sub-section instead instead of looking for the attr

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -762,14 +770,22 @@ def linspace_segments(
         self.points = new_points
 
     def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
+        # Ensure parent KSpace has normalized reciprocal_lattice_vectors first
+        k_space = self.m_parent
+        if k_space is not None and hasattr(
```
ladinesa: same

tests/test_numerical_settings.py
```diff
@@ -161,6 +161,38 @@ class TestKMesh:
...
+        )
+        # Normalize ModelSystem first so it's available for KSpace
+        simulation.model_system[0].normalize(EntryArchive(), logger)
+
+        k_space = simulation.model_method[0].numerical_settings[0]
+        k_mesh = k_space.k_mesh[0]
+
+        # Explicitly normalize KMesh before KSpace (simulating NOMAD's subsection-first order)
+        # This should trigger the fix where KMesh calls k_space.normalize_reciprocal_lattice_vectors()
+        k_mesh.normalize(EntryArchive(), logger)
```
ladinesa: why the need to initialize EntryArchive here and not pass simulation

tests/test_numerical_settings.py
```diff
@@ -314,6 +346,39 @@ class TestKLinePath:
     Test the `KLinePath` class defined in `numerical_settings.py`.
     """
 
+    def test_normalization_order_with_kspace(self):
```
ladinesa: can't you someone merge and parameterize these tests?

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: The `hasattr` check serves as defensive programming to ensure safe operation in edge cases. While `KMesh` and `KLinePath` are formally defined as subsections of `KSpace`, the runtime check provides safety if `m_parent` is unexpectedly None or not a `KSpace` instance during testing.

There's precedent in `physical_property.py:285` where `hasattr(contribution, 'normalize')` is used for similar defensive checking.

If you prefer a more explicit approach, I can change it to `isinstance(k_space, KSpace)`. Both provide runtime safety - let me know your preference.

tests/test_numerical_settings.py
ndaelman-hu: Good call! I've refactored both tests into a single parameterized test `test_kspace_subsection_normalization_order` that handles both `KMesh` and `KLinePath` cases. 

The new implementation:
- Uses `@pytest.mark.parametrize` with clear test IDs (`KMesh`, `KLinePath`)
- Eliminates code duplication while maintaining comprehensive coverage
- Passes both test cases successfully

This aligns with the project testing standards for using parameterization with edge cases. Thanks for the suggestion!

src/nomad_simulations/schema_packages/numerical_settings.py
ladinesa: It would be nice to avoid these duck-typing as much as possible.

tests/test_numerical_settings.py
ndaelman-hu: Good question! Currently we're creating separate `EntryArchive()` instances for each normalize call (model_system, then subsection).

Are you suggesting we should:
- **Normalize the whole simulation** and test the full cascade? (e.g., `simulation.normalize(archive, logger)`)
- **Or keep it specialized** to testing just the subsection's normalize, but reuse a single archive?

The current approach gives us precise control to normalize the subsection before its parent KSpace (to test issue #126's edge case), but I'm open to a different pattern if you prefer!

ladinesa:
> @ladinesa - is this pattern acceptable or should we use a different approach?

I needed to inject the logger into the msection somehow and the annotations seem to be the perfect place. We do not need a wrapper for this in my opinion.

tests/test_numerical_settings.py
ladinesa: The archive arg for the `normalize`strictly should be the root for the msection. This would test for any unusual behavior of the implementation. If it uses  `archive` then your test may get it wrong.

ndaelman-hu:
@ladinesa applied your suggestions as I understood them. Ready for follow-up review.

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -482,6 +482,11 @@ def resolve_k_line_density(
         return None
 
     def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
+        # Ensure parent KSpace has normalized reciprocal_lattice_vectors first
+        k_space = self.m_parent
+        if k_space is not None and k_space.m_def == KSpace.m_def:
```
ladinesa: do we check m_def or isinstance(k_space, KSpace) is sufficient?

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -762,14 +768,20 @@ def linspace_segments(
         self.points = new_points
 
     def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
+        # Ensure parent KSpace has normalized reciprocal_lattice_vectors first
+        k_space = self.m_parent
+        if k_space is not None and k_space.m_def == KSpace.m_def:
```
ladinesa: same

ladinesa:
LGTM please merge once all are resolved.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: I think that `isinstance` suffices, yes. We simply want to ensure that the correct interface is followed here.
It is also better than `m_def`, as interfaces are inherited (when Liskov substitution is observed).

So I changed it back to use `isinstance`.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: Changed to isinstance here as well.

JFRudzinski (approved):

PR #323, opened 2026-02-04
Md fix

amirgolp:
fixes related to topology normalizer


JFRudzinski (approved):

JFRudzinski:
@amirgolp - just fyi, I was too fast to merge this, now that I look again these functions don't need the logger passed as a keyword argument because they use a decorator which passes the logger appropriately. So I reverted these changes

PR #328, opened 2026-02-17
implement `NonlocalCorrelation`

### LINKED ISSUE #278: Extend DFT classification with libvdwxc
The title explains the intent well. I would follow the approach laid out in PR https://github.com/nomad-coe/nomad-simulations/pull/237

EBB2675:
## Purpose

This PR introduces a dedicated `NonlocalCorrelation` schema section for nonlocal correlation terms in DFT and aligns XC partner linking with typed references.

## Scope

**Included**

_1. Added `NonlocalCorrelation` as an explicit schema section_
- Introduces a dedicated section to represent nonlocal correlation contributions explicitly.
- Supports the following variants via `NonlocalCorrelation.type`:
  - `VV10`
  - `rVV10`
  - `vdW-DF`
  - `vdW-DF2`
  - `vdW-DF-cx`
  - `optB88-vdW`
  - `optB86b-vdW`
- Adds `NonlocalCorrelation.xc_partner_ref` as a typed reference to the paired `XCFunctional`.

_2. Standardized XC partner linkage for dispersion-related contributions_
- `EmpiricalDispersionModel` now uses `xc_partner_ref` instead of the old string field.
- `DFT.normalize` backfills missing `xc_partner_ref` values for:
  - `EmpiricalDispersionModel`
  - `NonlocalCorrelation`
- The partner reference is set to the active `dft.xc` section when available.

_3. Keeps nonlocal-correlation population parser-driven_
- This PR does **not** infer `NonlocalCorrelation` from `xc.functional_key`.
- This PR does **not** rewrite or strip `xc.functional_key`.
- Parsers are expected to create `NonlocalCorrelation` explicitly when nonlocal correlation is present.

## Design Notes

The earlier normalization-based inference for `+VV10` / `+rVV10` was removed. Based on review discussion, it is better to keep nonlocal-correlation population parser-driven for now, since many codes expose these settings through separate input flags or fields rather than embedding them in the XC functional label itself.

## Breaking / Schema Changes

- `EmpiricalDispersionModel.xc_partner` was replaced by `EmpiricalDispersionModel.xc_partner_ref`.

## Testing / Validation

- Added unit coverage in `tests/test_model_method.py` for:
  - setting `NonlocalCorrelation.xc_partner_ref` during `DFT.normalize` when it is missing

## Status

- [x] Ready for review

## Dependencies / Blockers

- [x] None


src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -596,6 +629,38 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        if nonlocal_corr_addon:
+            existing_nonlocal = None
+            for contribution in self.contributions or []:
+                if not isinstance(contribution, NonlocalCorrelation):
+                    continue
+                if contribution.type in (None, nonlocal_corr_addon):
+                    existing_nonlocal = contribution
+                    break
+
+            if existing_nonlocal is None:
```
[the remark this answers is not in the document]
EBB2675: thx for the suggestion, i will wait for nathan's comment

ndaelman-hu:
This PR (among other things) semantizes `NonlocalCorrelation` in our schema, giving it clearer visiblity and standardization.

> We currently infer nonlocal correlation only from functional_key patterns +VV10 / +rVV10 (via split_xc_and_addons()), but NonlocalCorrelation.type also includes vdW-DF variants (vdW-DF, vdW-DF2, vdW-DF-cx, optB88-vdW, optB86b-vdW). Should we (a) extend inference to recognize these variants in functional_key, (b) restrict the enum to what we infer automatically (VV10/rVV10) and expect vdW-DF variants to be set explicitly by parsers, or (c) keep both but document that only VV10/rVV10 are inferred today? @ndaelman-hu

Good questions — I did some (admittedly cursory) research into how codes report these functionals. I may well be missing cases, so take this with a grain of salt.

From what I could find, no more than one or two programs or libraries actually attach a nonlocal correlation tag to the functional name. It's more common in the literature, but most codes — VASP, QE, SIESTA, ABINIT — set it through separate input fields or flag combinations. So in practice it may be better to just have the parser set `NonlocalCorrelation` explicitly?

I understand the impulse to automate this and make life easier for parser developers, but the automation may be somewhat premature here. That said, I don't have a strong opinion on this question by itself — it's the second one that gives me a clearer preference.

> If an existing NonlocalCorrelation contribution conflicts with what we infer from functional_key, we currently create a second entry, which is not ideal. What would you suggest? @ndaelman-hu

I'm not entirely sure what you mean by creating multiple entries (more on that below), but it sounds like it could be a deeper-lying bug that would take significant time to trace. It may just be easier to remove the normalization logic and defer to the parser — in that case there should only ever be one `NonlocalCorrelation` section.

We could of course retain the inference if it's much more commonly needed in the QC community than my quick search suggests. But then we'd need to investigate the duplication or design cleanup logic to merge entries, which is more involved.

### Question to @EBB2675 

With "a second entry" do you mean two separate archives or two subsections within one? And did you notice this when running a parsing example? I ask because the behavior sounds in line with issues I've seen with the mapping annotations parser — I could see that being the underlying cause.

EBB2675:
> From what I could find, no more than one or two programs or libraries actually attach a nonlocal correlation tag to the functional name. It's more common in the literature, but most codes — VASP, QE, SIESTA, ABINIT — set it through separate input fields or flag combinations. So in practice it may be better to just have the parser set `NonlocalCorrelation` explicitly?


that sounds nicer, i will remove these normalizations and re-tag you for final look :+1:

EBB2675:
hi gentlemen @JFRudzinski and @ndaelman-hu 

this PR is ready for review, I have removed the old complicating normalization logic for attaching `NonlocalCorrelation` flags to `DFT`. Please see the updated description and let me know if there is anything needed to be added/removed

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -576,6 +621,15 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
         if self.xc is None:
             self.xc = XCFunctional()
 
+        for contribution in self.contributions or []:
```
ndaelman-hu: `or []` so the repeating subsection is not (automatically) instantiated?

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -576,6 +621,15 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+            if (
+                isinstance(
+                    contribution, (EmpiricalDispersionModel, NonlocalCorrelation)
+                )
+                and contribution.xc_partner_ref is None
+            ):
```
ndaelman-hu, line 630: Would swap the order here: Python will jump the if-body as soon as it encounters a false statement.

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -264,8 +266,8 @@ class EmpiricalDispersionModel(BaseModelMethod):
     )
 
     # TODO later: link to XCComponent(s)
-    xc_partner = Quantity(
-        type=str,
+    xc_partner_ref = Quantity(
+        type=Reference(SectionProxy('XCFunctional')),
```
ndaelman-hu, line 270: Minor remark: `SectionProxy` is only necessary here because `XCFunctional` is defined below.
You don't have to, but if you move this class underneath it, you can just use `Reference(XCFunctional)`. Even `Reference` might be redundant then.

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -329,6 +331,49 @@ class RelativityModel(BaseModelMethod):
...
+            'optB88-vdW',
+            'optB86b-vdW',
+        ),
+        description="""
+        Identifier of the nonlocal correlation functional / variant.
+        """,
+    )
+
+    xc_partner_ref = Quantity(
+        type=Reference(SectionProxy('XCFunctional')),
```
ndaelman-hu, line 370: Same remark as above.

ndaelman-hu (approved):
PR updates the connection between `NonLocalCorrelation` and `XCFunctional`.
Technically in order, only small nitpicks.

JFRudzinski (approved):

src/nomad_simulations/schema_packages/model_method.py
EBB2675: i know, thanks alot!

src/nomad_simulations/schema_packages/model_method.py
EBB2675: removed or

src/nomad_simulations/schema_packages/model_method.py
EBB2675: doesnt Python short-circuit "and" from left to right
 
 So:
	1.	it first checks whether contribution is one of those two classes
	2.	only if that is True does it evaluate `contribution.xc_partner_ref` is None

That matters because `self.contributions` can contain other contribution types too. If you swap the order, then Python may try to read `xc_partner_ref` on an object that does not even define that attribute. That could raise an `AttributeError`

PR #338, opened 2026-02-26
Add quantity `mass` to AtomsState

### LINKED ISSUE #337: Duplicate/conflicting Quantity definitions in AtomsState and CGBeadState
## Charge
- `AtomsState.charge`: the real, physical charge of an atom or ion: an (integer) number of electrons gained or lost. Type: `int`, unit: `elemental charge`. Normalized to an integer value by ASE downstream.
- `AtomsState.partial_charge`: the continuous, non-integer charge arising from unequal electron sharing due to electronegativity differences. Type: `float`, unit: `elemental charge`. Excluded from ASE normalization. Required by several simulation methods.
- `CGBeadState.charge`: the charge assigned to a coarse-grained bead. Type: `float`, unit: `coulomb`. **Naming/usage conflict — see below.**

## Mass
- `AtomsState.mass`: the atomic mass, populated from force field parameters by MD parsers. Type: `float`, unit: `kg`. **No ASE normalization**.
- `CGBeadState.mass`: the total mass of the particle, also populated from force field parameters. Same type and unit as `AtomsState.mass`.

### Problems
- `CGBeadState.charge` is conceptually equivalent to `AtomsState.partial_charge`, not `AtomsState.charge`. The naming conflict risks incorrect normalization: CG particles do not necessarily carry integer charges, and their charge values are not necessarily reflected in the atomic references used by ASE.
- Defining both `charge` and `partial_charge` across `AtomsState` and `CGBeadState` is redundant.
- The same redundancy applies to `mass` being defined separately in both `AtomsState` and `CGBeadState`.

### Questions
- Should `mass` and `partial_charge` be lifted to `ParticleState`, so each is defined only once? The unit difference in the charge case would need to be handled at parsing/normalization time.
- Does it semantically make sense for `ParticleState` to carry a `partial_charge` quantity? Renaming it to plain `charge` would require renaming `AtomsState.charge`, which is used elsewhere (NMR) and would likely cause ripple effects.

Bernadette-Mohr:
## Purpose

Add a Quantity 'mass' to AtomsState, containing the atomic mass, unit 'kg'. Needed for molecular dynamics atoms definitions and downstream calculation of MolecularDynamicsResults quantities via the center of mass.

## Scope

- Definition of Quantity 'mass' in AtomsState

**Out of scope**

- Discussion of organization and semantics: mass is now defined twice in the same way, both in AtomsState and CGBeadState. Should it be elevated to ParticleState to prevent code repetitions?

## Context & Links

- Related issue: https://github.com/FAIRmat-NFDI/nomad-simulations/issues/337

## Reviewer Notes

See related issue

## Status

- [ ] Draft / In progress
- [X] Ready for review
- [ ] Blocked (explain):

## Breaking Changes

- [X] None
- [ ] (explain):

## Dependencies / Blockers

- [X] None
- [ ] Open design questions (explain):
- [ ] External dependencies (explain):

## Testing / Validation

- [ ] None (explain):
- [ ] Unit tests
- [ ] Integration tests
- [X] Manual testing (explain): Required for updating `MolecularDynamicsResults` to process `archive.data`, usage tested there.

### DIFF
--- a/src/nomad_simulations/schema_packages/atoms_state.py
+++ b/src/nomad_simulations/schema_packages/atoms_state.py
@@ -620,10 +620,10 @@ class ElectronicState(Entity):
         section_def=BaseSpinOrbitalState.m_def,
         description="""
         The actual quantum state descriptor at this level of the hierarchy.
-        This `BaseSpinOrbitalState` (typically `SphericalSymmetryState`) defines the 
+        This `BaseSpinOrbitalState` (typically `SphericalSymmetryState`) defines the
         quantum numbers and properties of the state in a modular fashion.
         Child states in `sub_states` may inherit quantum numbers from this parent descriptor.
-        
+
         For example, if parent has `spin_orbit_state=SphericalSymmetryState(n=3, l=2)`,
         a child might only specify `spin_orbit_state=SphericalSymmetryState(ml=-2)`,
         with n=3, l=2 implied from the parent.
@@ -637,7 +637,7 @@ class ElectronicState(Entity):
         - For single orbitals: typically 1 (if ml, ms specified) or 2*l+1 (orbital only)
         - For manifolds: sum of constituent state degeneracies
         - For symmetry-adapted states: dimension of irreducible representation
-        
+
         Can be computed from `spin_orbit_state` or summed from `sub_states`.
         For correlated systems, represents the many-body state degeneracy.
         """,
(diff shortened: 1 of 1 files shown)


EBB2675 (changes requested):
if `AtomsState.mass` may not necessarily correspond to an actual atomic mass, but can also be a parameter for _particles_, i suggest lifting it up to `ParticleState`, otherwise it may be confusing

PR #339, opened 2026-02-26
Archive data update

### LINKED ISSUE #334: Typo in integrator name: `conjuga_nt_ gradient` instead of `conjuga_te_ gradient`
There is a recurrent typo of the integrator name, source of correct spelling: https://en.wikipedia.org/wiki/Conjugate_gradient_method

Affected are:

- nomad-FAIR:
  - nomad/datamodel/metainfo/simulation/legacy_workflows.py
  - nomad-FAIR/nomad/datamodel/metainfo/simulation/workflow.py
- nomad-parser-plugins-simulation:
  - docs/parsers/h5md/h5md_expl.md
- nomad-parser-plugins-atomistic:
  - atomisticparsers/lammps/parser.py
- nomad-simulations:
  - src/nomad_simulations/schema_packages/workflow/geometry_optimization.py
  - src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
- nomad-schema-plugin-simulation-workflow:
  - simulationworkflowschema/geometry_optimization.py
  - simulationworkflowschema/molecular_dynamics.py

Bernadette-Mohr:
## Purpose
Complete the migration of the MD workflow normalizer from the legacy `archive.run` schema to `archive.data` (`nomad-simulations`), adapt `archive_to_universe` to support current `atoms_state` definitions.

## Scope

**Included**

- `MolecularDynamicsResults.normalize()` and all helper methods (`_get_molecular_rdfs`, `_get_molecular_msds`, `get_molecular_rgs`, `_populate_output_rgs`) ported from `archive.run[].system/calculation` to `archive.data.model_system/outputs`
- `archive_to_universe`: rewritten to build an MDAnalysis `Universe` from `archive.data`; updated to handle current `atoms_state` structure
- `get_bond_list_from_model_contributions`: ported from commented-out legacy stub to working implementation against `ForceField.contributions`
- Masses and partial charges extracted from `ParticleParametersContainer` and passed to the universe
- `ModelSystem.particle_states` reassignment logic generalised to handle mixed atomic/CG label sets
- Dead code removed: legacy `Property.type` assignment on RDF/MSD objects, commented-out `atomsgroup_ref` block (references `AtomsGroup`, which no longer exists in the new schema)
- Tests added/updated for `archive_to_universe`, `ModelSystem`, and `ForceField` against new schema
- Fix typo in MolecularDynamicsMethod.integrator_type enum: 'conjugant_gradient' → 'conjugate_gradient' (closes #334)

**Out of scope**

- Replacing `atomsgroup_ref` with a `Reference(ModelSystem)` link on `RadiiOfGyration`
- Full test coverage of `MolecularDynamicsResults.normalize()` (requires MDAnalysis fixture trajectories)

## Status

- [x] Ready for review

## Breaking Changes

- [x] None

## Dependencies / Blockers

- [x] None

## Testing / Validation

- [x] Unit tests (`test_molecular_dynamics`, `test_model_system`, `test_force_field`)
- [x] Manual testing: GROMACS all-atom and CG trajectories parsed end-to-end


tests/test_force_field.py
```diff
@@ -1193,3 +1193,32 @@ def test_missing_units_skip_derivation():
...
+
+def test_particle_parameters_effective_mass_stores_and_units():
+    """effective_mass on ParticleParameters retains kg units."""
+    pp = ParticleParameters(particle_type='OW')
+    pp.effective_mass = (15.999 * ureg.amu).to('kg')
+    assert pp.effective_mass is not None
+    assert np.isclose(pp.effective_mass.to('amu').magnitude, 15.999, rtol=1e-5)
+
+
+def test_particle_parameters_effective_mass_none_by_default():
```
EBB2675: this test can be safely removed 👍

tests/utils/test_molecular_dynamics.py
```diff
@@ -0,0 +1,210 @@
...
+    result = getattr(mm, 'partial_charges', None)
+    assert result is None
+
+
+def test_getattr_guard_forcefield_effective_masses_returns_value():
+    """getattr returns the value when ForceField.effective_masses is set."""
+    ff = ForceField()
+    masses_amu = np.array([15.999, 1.008, 1.008])
+    ff.effective_masses = (masses_amu * ureg.amu).to('kg')
+
```
EBB2675: `ForceField` does not declare a `effective_masses` quantity

tests/utils/test_molecular_dynamics.py
```diff
@@ -0,0 +1,215 @@
...
+    assert result is not None
+    assert len(result) == 3
+    assert np.isclose(result.to('amu').magnitude, masses_amu, rtol=1e-5).all()
+
+
+def test_getattr_guard_forcefield_partial_charges_returns_value():
+    """getattr returns the value when ForceField.partial_charges is set."""
+    ff = ForceField()
+    charges = np.array([-0.82, 0.41, 0.41])
+    ff.partial_charges = charges * ureg.elementary_charge
```
EBB2675: `ForceField` does not declare a `partial_charges` quantity in the schema. These tests pass because Python allows attaching dynamic attributes to the instance, not because the metainfo path is valid, please fix

tests/test_force_field.py
```diff
@@ -1193,3 +1193,32 @@ def test_missing_units_skip_derivation():
...
+
+def test_particle_parameters_partial_charge_stores_and_units():
+    """partial_charge on ParticleParameters retains elementary_charge units."""
+    pp = ParticleParameters(particle_type='OW')
+    pp.partial_charge = -0.82 * ureg.elementary_charge
+    assert pp.partial_charge is not None
+    assert np.isclose(pp.partial_charge.to('elementary_charge').magnitude, -0.82)
+
+
+def test_particle_parameters_partial_charge_none_by_default():
```
EBB2675: this test can be safely removed 👍

EBB2675 (commented):
**Review 1:**  the new `getattr` checks for `ForceField.partial_charges` / `effective_masses` do not seem to exercise real schema quantities. Those names are not declared on `ForceField`; Python just allows attaching dynamic attributes to the instance at runtime. The actual schema path here is `ParticleParameters.{partial_charge,effective_mass}`


**Question 1:** the current mass/charge lookup is built from `ParticleParameters.particle_type` and then resolved via `ParticleState.label`. Is it guaranteed upstream that these two are always normalized to the same string?

I checked the parser side, and I could not find that guarantee. The shared MD parser utilities appear to normalize atomic particle labels to chemical symbols like H/O, and I did not find parser-side wiring that would realign FF particle types such as OW/HW onto `ParticleState.label` or populate `ParticleParametersContainer` in a way that enforces that equivalence. So from what I can see, the current lookup seems to work only if particle_type == label by construction. If that alignment is guaranteed somewhere else, it would be helpful to state that explicitly; otherwise it seems safer to use `species_scope` as the authoritative mapping instead of relying on string equality.

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -789,7 +793,7 @@ class MolecularDynamicsMethod(SimulationWorkflowMethod):
     integrator_type = Quantity(
         type=MEnum(
             'brownian',
-            'conjugant_gradient',
+            'conjugate_gradient',
```
EBB2675, line 807: Changing `'conjugant_gradient'` to `'conjugate_gradient'` is small, but it is **still** a schema/API change. Please either keep changes like this in a separate small PR, or call them out explicitly in the PR title/description. 

In the current PR, this change is neither reflected in the title nor mentioned in the description, so it is easy to miss during review and can surprise downstream consumers that rely on the old enum value

EBB2675: or you can tie the issue #334 to here so that we understand that this issue was solved in this PR

Bernadette-Mohr:
> **Review 1:** the new `getattr` checks for `ForceField.partial_charges` / `effective_masses` do not seem to exercise real schema quantities. Those names are not declared on `ForceField`; Python just allows attaching dynamic attributes to the instance at runtime. The actual schema path here is `ParticleParameters.{partial_charge,effective_mass}`
> 
> **Question 1:** the current mass/charge lookup is built from `ParticleParameters.particle_type` and then resolved via `ParticleState.label`. Is it guaranteed upstream that these two are always normalized to the same string?
> 
> I checked the parser side, and I could not find that guarantee. The shared MD parser utilities appear to normalize atomic particle labels to chemical symbols like H/O, and I did not find parser-side wiring that would realign FF particle types such as OW/HW onto `ParticleState.label` or populate `ParticleParametersContainer` in a way that enforces that equivalence. So from what I can see, the current lookup seems to work only if particle_type == label by construction. If that alignment is guaranteed somewhere else, it would be helpful to state that explicitly; otherwise it seems safer to use `species_scope` as the authoritative mapping instead of relying on string equality.

The string-equality assumption is not guaranteed on purpose. FF types do not automatically align with chemical symbols. `archive_to_universe` now resolves masses/charges via `species_scope` first (the identity-based reference list populated by `ParticleParametersContainer.normalize()`), falling back to `particle_type == ps.label` string match only when `species_scope` is not yet populated (e.g., in tests that skip normalization). Since `archive_to_universe` is called from `MolecularDynamicsResults.normalize()`, `ParticleParametersContainer.normalize()` will have already run and `species_scope` should be authoritative.

EBB2675:
> The string-equality assumption is not guaranteed on purpose. FF types do not automatically align with chemical symbols. `archive_to_universe` now resolves masses/charges via `species_scope` first (the identity-based reference list populated by `ParticleParametersContainer.normalize()`), falling back to `particle_type == ps.label` string match only when `species_scope` is not yet populated (e.g., in tests that skip normalization). Since `archive_to_universe` is called from `MolecularDynamicsResults.normalize()`, `ParticleParametersContainer.normalize()` will have already run and `species_scope` should be authoritative.

Perfect! My local testing for the non-happy path gave an error while giving the review, but starting from `ab42bfe70486832f927b1ebdf4d7fcb53e6c64cf` it behaves correctly :+1:

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -338,34 +340,26 @@ def create_empty_universe(
     return universe
 
 
-# TODO: update run to data
 def archive_to_universe(
     archive,
-    system_index: int = 0,
+    system_index: int = 0,  # TODO: Shouldn't this be -1 (representative_system)?
```
JFRudzinski: set to None default and then set to representative_system_index if None?

JFRudzinski (approved):
Looks good to me 👍

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
Bernadette-Mohr: Here, the system is accessed for the topology, which is not required to be attached to the representative system. I have tightened the controls to not rely on any subsystem index that might not be accurate.

JFRudzinski:
@Bernadette-Mohr What is the status here? Is this being merged or are you waiting on something?

Bernadette-Mohr:
> @Bernadette-Mohr What is the status here? Is this being merged or are you waiting on something?

I'm principal it's ready to merge, but I added more changes after your review, so I wanted to give you a last chance to object.

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -400,45 +392,146 @@ def archive_to_universe(
...
         return None
+    particle_states = sec_system_top.particle_states if sec_system is not None else None
+    if not particle_states:
+        LOGGER.warning(
+            'No particle_states found in topology system. Cannot build MDA universe.'
+        )
+        return None
+    particle_names = [ps.label for ps in particle_states]
+    particle_types = [
+        getattr(ps, 'chemical_symbol', None) or getattr(ps, 'bead_symbol', 'CGX')
```
JFRudzinski:
```suggestion
ps.chemical_symbol or ps.bead_symbol or 'CGX'
```

please use nomadic style instead of python natives

src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
```diff
@@ -400,45 +392,146 @@ def archive_to_universe(
...
+            )
+        elif pp is not None and pp.effective_mass is not None:
+            _masses_list.append(
+                ureg.convert(
+                    pp.effective_mass.magnitude, pp.effective_mass.units, ureg.amu
+                )
+            )
+        else:
+            _missing_masses += 1
+            symbol = getattr(ps, 'chemical_symbol', None)
```
JFRudzinski: same here, please go through and remove getattr for metainfo sections

JFRudzinski (approved):
Looks good, just some minor styling changes, after that you can merge

PR #353, opened 2026-03-18
change the section reference direction of `Pseudopotential` vs `AtomsState`

EBB2675:
## Purpose

This PR removes the pseudopotential reference from `AtomsState` and moves that linkage to `Pseudopotential` via a forward `species_scope` reference. This keeps `AtomsState` free of method-specific metadata and aligns pseudopotential assignment with other method-side scoping patterns already used in the schema.

## Scope

**Included**

- Remove `AtomsState.pseudopotential`
- Add `Pseudopotential.species_scope` referencing `AtomsState`
- Add regression coverage for the new reference direction and archive serialization
- Update schema docs to reflect the new ownership model

## Context & Links

- Schema cleanup: pseudopotential assignment belongs with method/numerical settings, not particle-state definition
- This follows the same forward-reference pattern already used by method-side sections such as basis set and force field atom parameters

## Reviewer Notes

- No parser was populating `AtomsState.pseudopotential` in `nomad-parser-plugins-simulation@develop`
Other branches have not been checked.

## Status

- [x] Ready for review

## Breaking Changes

- [x] (explain):  `AtomsState.pseudopotential` was removed; consumers must read/write pseudopotential assignment through `Pseudopotential.species_scope`

## Dependencies / Blockers

- [x] None

## Testing / Validation

- [x] Unit tests

### DIFF
--- a/docs/schema/particle_states.md
+++ b/docs/schema/particle_states.md
@@ -110,3 +110,4 @@ classDiagram
 | `slater_integrals` | m_float64(float64) (shape: [3]) | <details><summary>Value of the Slater integrals [F0, F2, F4] in spherical harmonics used to derive</summary>Value of the Slater integrals [F0, F2, F4] in spherical harmonics used to derive<br>the local Hubbard interactions:<br>u_interaction = ((2.0 / 7.0) ** 2) * (F0 + 5.0 * F2 + 9.0 * F4) / (4.0*np.pi)<br>u_interorbital_interaction = ((2.0 / 7.0) ** 2) * (F0 - 5.0 * F2 + 3.0 * 0.5 * F4) / (4.0*np.pi)<br>j_hunds_coupling = ((2.0 / 7.0) ** 2) * (5.0 * F2 + 15.0 * 0.25 * F4) / (4.0*np.pi)<br>See e.g., Elbio Dagotto, Nanoscale Phase Separation and Colossal Magnetoresistance,<br>Chapter 4, Springer Berlin (2003).</details> |
 | `double_counting_correction` | m_str(str) | Name of the double counting correction algorithm applied. |
 
+
--- a/docs/schema/particle_states.md
+++ b/docs/schema/particle_states.md
@@ -70,7 +70,6 @@ classDiagram
 | `charge` | m_int32(int32) | <details><summary>Formal integer charge of the atom, defined as the number of extra</summary>Formal integer charge of the atom, defined as the number of extra<br>electrons (negative) or holes (positive) relative to the neutral atom.<br>For neutral atoms `charge = 0`.<br>Note: for `CoreHole` systems we do not consider the charge of the atom even if<br>we do not store the final `ElectronicState` where the electron was excited to.</details> |
 | `spin` | m_int32(int32) | Total spin quantum number, S. |
 | `label` | m_str(str) | User- or program-package-defined identifier for this atomic site. e.g. 'H1', 'H1a', 'C_eq'. It doesn't replace `chemical_symbol`, but merely gives users a more specialized token for the unique site name. |
-| `pseudopotential` | Reference | <details><summary>Reference to the pseudopotential used for this atomic species in plane-wave DFT</summary>Reference to the pseudopotential used for this atomic species in plane-wave DFT<br>calculations. The referenced `Pseudopotential` section is defined in `numerical_settings`<br>and contains metadata such as pseudopotential type (PAW, ultrasoft, norm-conserving),<br>cutoff energy, and XC functional used to generate the pseudopotential.</details> |
 
 ### `CGBeadState`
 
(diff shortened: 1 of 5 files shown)


src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -1054,6 +1054,16 @@ class Pseudopotential(NumericalSettings):
         """,
     )
 
+    species_scope = Quantity(
+        type=Reference(
+            SectionProxy('nomad_simulations.schema_packages.atoms_state.AtomsState')
+        ),
+        shape=['*'],
```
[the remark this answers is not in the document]
EBB2675: done, danke

docs/schema/particle_states.md
```diff
@@ -70,7 +70,6 @@ classDiagram
...
 | `spin` | m_int32(int32) | Total spin quantum number, S. |
 | `label` | m_str(str) | User- or program-package-defined identifier for this atomic site. e.g. 'H1', 'H1a', 'C_eq'. It doesn't replace `chemical_symbol`, but merely gives users a more specialized token for the unique site name. |
-| `pseudopotential` | Reference | <details><summary>Reference to the pseudopotential used for this atomic species in plane-wave DFT</summary>Reference to the pseudopotential used for this atomic species in plane-wave DFT<br>calculations. The referenced `Pseudopotential` section is defined in `numerical_settings`<br>and contains metadata such as pseudopotential type (PAW, ultrasoft, norm-conserving),<br>cutoff energy, and XC functional used to generate the pseudopotential.</details> |
```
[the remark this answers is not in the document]
EBB2675: we weren't populating these anyhow so i ll just skip this

tests/test_numerical_settings.py
```diff
@@ -74,6 +78,46 @@ def mock_xpath(path, dict=False):
             assert k_space.reciprocal_lattice_vectors == result
 
 
+class TestPseudopotential:
+    def test_species_scope_references_atoms_state(self):
```
ladinesa: maybe parameterize this to test also negative cases.

tests/test_numerical_settings.py
```diff
@@ -74,6 +78,46 @@ def mock_xpath(path, dict=False):
...
+        entry = EntryArchive(
```
ladinesa: if this is common, maybe make it a fixture or have a function to create it

tests/test_numerical_settings.py
EBB2675: done

tests/test_numerical_settings.py
EBB2675: will do a general test cleanup in an upcoming PR, there are similar cases in many other tests as well

ladinesa (approved):

tests/test_numerical_settings.py
```diff
@@ -74,6 +78,79 @@ def mock_xpath(path, dict=False):
             assert k_space.reciprocal_lattice_vectors == result
 
 
+class TestPseudopotential:
```
ndaelman-hu: I think this test (as is) is not necessary:

- What is this test targeting? That we can set references?
- Are none or empty edge cases here?

If we were to enforce that when PPs are present, they match the no. `AtomsState`s would make more sense to test. This is just illustrative: I think that such acheck is overkill atm.

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -1054,6 +1055,14 @@ class Pseudopotential(NumericalSettings):
         """,
     )
 
+    species_scope = Quantity(
+        type=AtomsState,
+        shape=['*'],
+        description="""
+        References to the `AtomsState` sections using this pseudopotential.
+        """,
+    )
```
ndaelman-hu, line 1064: I don't think a single Pseudopotential can encode multiple `AtomState`s. The inverse direction is in principle possible.

```suggestion
    species_scope = Quantity(
        type=AtomsState,
        description="""
        References to the `AtomsState` sections using this pseudopotential.
        """,
    )
```

ndaelman-hu (approved):
I checked, and the VASP PP PR does not contain any references yet. That was slated for later on. So this should not break with the changes there.

src/nomad_simulations/schema_packages/numerical_settings.py
EBB2675: so if there are multiple carbon atoms (with the same pseuodopotential), you just want to reference one of their AtomsState instances? what will happen to the other carbons and their AtomsState instances?

tests/test_numerical_settings.py
EBB2675: i can remove the tests 👍

src/nomad_simulations/schema_packages/numerical_settings.py
EBB2675: shouldnt one pseudopotential refer to the AtomsState's of all relevant carbons or all relevant hydrogens in the system?

JFRudzinski (approved):

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: > shouldnt one pseudopotential refer to the AtomsState's of all relevant carbons or all relevant hydrogens in the system?

Not necessarily. When trying to capture special electronic states, e.g. a core-hole, one would leverage a different PP of that specific atom. In codes like VASP, the PP are the element definition. I guess that the `AtomsState`, meanwhile, is a bit coarser, referring either to the physical element (or its state), or whatever semantic distinction the user decides to make.

> so if there are multiple carbon atoms (with the same pseuodopotential), you just want to reference one of their AtomsState instances? what will happen to the other carbons and their AtomsState instances?

Wouldn't you have a single `AtomsState` for all carbon atoms? If one receives special treatment and should be singled out, this will also be reflected in their PP.

EBB2675: > Wouldn't you have a single `AtomsState` for all carbon atoms? 

no, every atom has its own AtomsState instance. this class does not refer to the atom type

ndaelman-hu: Okay, thank you for refreshing my memory. Then the shape can stay. Or, we route the PP via the atom type. That would be in line with how most solid state codes think of it anyway.

EBB2675: This came up again recently in both MD and QC contexts, for example atom parameters in MD and atom-centered basis sets in QC. For the following two reasons, we decided to resort to&stay with the implementation style of `species_scope` in the `BasisSetComponent` implementation for these fields:

**Reason 1)** Although these quantities are usually assigned by atom type in most codes, it is still technically possible to assign different functions, scalars, or potentials to atoms of the same type. For example, one may choose to treat an active site with a more accurate set of functions, scalars, or potentials, while assigning less accurate ones to other atoms of the same species elsewhere in the system.

**Reason 2)** Even if we introduce an atom-type label layer such as

`PP` -> `AtomType` -> `AtomsState`

we still ultimately need a way to reference individual atoms, so that navigation to the atom level remains possible.

I will merge this to be able to consistent with other functions/scalars/potentials assignment to particles/atoms in the data model, feel free to create an issue to add another atom type layer to all of them :+1:

PR #363, opened 2026-03-24
implement Broken symmetry density functional theory

EBB2675:
## Purpose

add a `BSDFT` model-method class for the broken-symmetry calculation step, without introducing the full workflow orchestration yet.

In NOMAD semantics, `BSDFT` is intrinsically a workflow: a standard high-spin `DFT` calculation is performed first, and the broken-symmetry determinant is then constructed as a follow-up calculation on top of that reference. This PR does not implement that workflow yet. Instead, it introduces the method-level schema needed to represent the broken-symmetry calculation itself, so that a later `DFT` -> `BSDFT` workflow can be modeled cleanly and explicitly.

## Scope

**Included**

- Add `BSDFT(DFT)` to the model-method schema
- Add `BrokenSymmetryCenter` with atom/site reference and local spin base classes from `atoms_state.py`
- Add `BSDFT.spin_centers`
- Add `BSDFT.total_spin_projection` in doubled form
- Normalize `BSDFT.name = 'BSDFT'`
- Validate BSDFT-specific constraints:
_unrestricted determinant_
_spin-polarized calculation_
_at least two spin centers_
_at least one up and one down center_
- Add unit-test coverage for BSDFT behavior
- Add a minimal documentation/example snippet for BSDFT

**Out of scope**

- Any BS workflow class
- Linking a BSDFT calculation to a high-spin reference
- Exchange-coupling (J) extraction
- ⟨S²⟩ handling
- Projected-spin or spin-Hamiltonian analysis

## Context & Links

- Prepares for a later workflow-level orchestration similar to `DFT` -> `GW`
- High-spin reference remains standard `DFT`; only the broken-symmetry calculation is specialized as `BSDFT`
- Reuses existing spin-state schema concepts from `atoms_state.py` instead of introducing a parallel ad hoc spin-sign model

## Reviewer Notes

BSDFT is modeled as a subclass of DFT, not as a flag/subsection on DFT, to keep the method identity explicit and make future workflow detection straightforward.

Validation currently warns on inconsistent BSDFT metadata rather than rewriting or dropping parsed values, matching the general schema style.

## Status

- [x] Ready for review

## Breaking Changes

- [x] None

## Dependencies / Blockers

- [x] None

## Testing / Validation

- [x] Unit tests

### DIFF
--- a/docs/snippets/model_method/bs_dft_example.py
+++ b/docs/snippets/model_method/bs_dft_example.py
@@ -1,0 +1,20 @@
+# docs-snippet: runnable
+from nomad_simulations.schema_packages.atoms_state import AtomsState
+from nomad_simulations.schema_packages.model_method import BSDFT, BrokenSymmetryCenter
+
+
+def build_bs_dft_example() -> tuple[object, BSDFT]:
+    """Create a plain DFT-like high-spin reference context and a BSDFT method."""
+    fe1 = AtomsState(chemical_symbol='Fe', label='Fe1')
+    fe2 = AtomsState(chemical_symbol='Fe', label='Fe2')
+
+    method = BSDFT(
+        determinant='unrestricted',
+        is_spin_polarized=True,
+        total_spin_projection=0,
+    )
+    method.spin_centers = [
+        BrokenSymmetryCenter(atom_ref=fe1, spin_sign='up', label='site_a'),
+        BrokenSymmetryCenter(atom_ref=fe2, spin_sign='down', label='site_b'),
+    ]
+    return fe1, method
--- a/docs/snippets/model_method/bs_dft_example.py
+++ b/docs/snippets/model_method/bs_dft_example.py
@@ -1,5 +1,8 @@
 # docs-snippet: runnable
-from nomad_simulations.schema_packages.atoms_state import AtomsState
+from nomad_simulations.schema_packages.atoms_state import (
+    AtomsState,
+    SphericalSymmetryState,
+)
 from nomad_simulations.schema_packages.model_method import BSDFT, BrokenSymmetryCenter
 
 
@@ -14,7 +17,15 @@ def build_bs_dft_example() -> tuple[object, BSDFT]:
         total_spin_projection=0,
     )
     method.spin_centers = [
-        BrokenSymmetryCenter(atom_ref=fe1, spin_sign='up', label='site_a'),
-        BrokenSymmetryCenter(atom_ref=fe2, spin_sign='down', label='site_b'),
+        BrokenSymmetryCenter(
+            atom_ref=fe1,
+            spin_state=SphericalSymmetryState(ms_quantum_number=0.5),
+            label='site_a',
+        ),
+        BrokenSymmetryCenter(
+            atom_ref=fe2,
+            spin_state=SphericalSymmetryState(ms_quantum_number=-0.5),
+            label='site_b',
+        ),
     ]
     return fe1, method
--- a/docs/snippets/model_method/bs_dft_example.py
+++ b/docs/snippets/model_method/bs_dft_example.py
@@ -12,8 +12,8 @@ def build_bs_dft_example() -> tuple[object, BSDFT]:
     fe2 = AtomsState(chemical_symbol='Fe', label='Fe2')
 
     method = BSDFT(
-        determinant='unrestricted',
-        is_spin_polarized=True,
+        determinant='unrestricted',  # required for consistency check
+        is_spin_polarized=True,  # required for consistency check
         total_spin_projection=0,
     )
     method.spin_centers = [
--- a/src/nomad_simulations/schema_packages/model_method.py
+++ b/src/nomad_simulations/schema_packages/model_method.py
@@ -18,7 +18,11 @@ if TYPE_CHECKING:
     from nomad.datamodel.datamodel import EntryArchive
     from structlog.stdlib import BoundLogger
 
-from nomad_simulations.schema_packages.atoms_state import CoreHole, ElectronicState
+from nomad_simulations.schema_packages.atoms_state import (
+    AtomsState,
+    CoreHole,
+    ElectronicState,
+)
 from nomad_simulations.schema_packages.data_types import (
     positive_float,
     positive_int,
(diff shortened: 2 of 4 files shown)


src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -763,6 +769,119 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        if len(spin_centers) < 2:
+            logger.warning(
+                'BSDFT requires at least two spin_centers to define a broken-symmetry assignment.'
+            )
+            return False
+
+        signs = {center.resolve_spin_sign() for center in spin_centers}
+        if None in signs:
+            logger.warning(
+                'BSDFT spin_centers require spin_state with a non-zero ms_quantum_number.'
```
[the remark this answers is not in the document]
EBB2675: you are absolutely right :robot:

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -763,6 +769,120 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+    def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
+        super().normalize(archive, logger)
+
+        self.name = 'BSDFT'
+
+        if self.determinant != 'unrestricted':
+            logger.warning('BSDFT requires determinant to be unrestricted.')
+
+        if self.is_spin_polarized is not True:
+            logger.warning('BSDFT requires is_spin_polarized to be True.')
```
ndaelman-hu: Should the warning also trigger for unset determinants or `is_spin_polarized`?

ndaelman-hu: Nitpick: personally try to always you the code highlight for code-specific terminology. Is not applied uniformly throughout our schema, or NOMAD pydocs.


```suggestion
        if self.determinant != 'unrestricted':
            logger.warning('BSDFT requires determinant to be unrestricted.')

        if self.is_spin_polarized is not True:
            logger.warning('BSDFT requires `is_spin_polarized` to be True.')
```

docs/snippets/model_method/bs_dft_example.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def build_bs_dft_example() -> tuple[object, BSDFT]:
+    """Create a plain DFT-like high-spin reference context and a BSDFT method."""
+    fe1 = AtomsState(chemical_symbol='Fe', label='Fe1')
+    fe2 = AtomsState(chemical_symbol='Fe', label='Fe2')
+
+    method = BSDFT(
+        determinant='unrestricted',
+        is_spin_polarized=True,
```
ndaelman-hu: maybe add an inline code comment saying that these fields have to be set or else a consistency warning will trigger:


```suggestion
        determinant='unrestricted',  # required for consistency check
        is_spin_polarized=True,      # required for consistency check
```

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -763,6 +769,120 @@ def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
...
+        type=np.int32,
+        description="""
+        Total spin projection M_S of the broken-symmetry determinant, stored in doubled
+        form to preserve half-integer values (e.g. M_S = 1/2 is stored as 1).
+        This is not the total spin quantum number S of a spin eigenstate.
+        """,
+    )
+
+    def _validate_spin_centers(self, logger: 'BoundLogger') -> bool:
+        spin_centers = self.spin_centers or []
```
ndaelman-hu, line 857: It's kinda clear from the warning messages what is being checked, but doesn't hurt to add it as a pydocs here too.

tests/test_model_method.py
```diff
@@ -622,6 +624,175 @@ def test_exact_exchange_mixing_factor_resolution(
...
+        )
+
+        bsdft.normalize(EntryArchive(), logger=logger)
+
+        assert any(
+            entry['event']
+            == 'BSDFT spin_centers must provide a resolvable spin sign (e.g. via a '
+            'SphericalSymmetryState spin_state with non-zero ms_quantum_number).'
+            for entry in log_output.entries
+        )
```
ndaelman-hu: Could be combined via parametrization.

tests/test_model_method.py
```diff
@@ -622,6 +624,175 @@ def test_exact_exchange_mixing_factor_resolution(
...
+
+        bsdft.normalize(EntryArchive(), logger=logger)
+
+        assert bsdft.name == 'BSDFT'
+        assert bsdft.jacobs_ladder == 'GGA'
+        assert {c.canonical_label for c in (bsdft.xc.components or [])} == {
+            'XC_GGA_X_PBE',
+            'XC_GGA_C_PBE',
+        }
```
ndaelman-hu: Is there a reason to presume that it wouldn't? I guess we could add tests verifying that the parent normalizer is run (which applies to the vast majority of cases). I think such a test is beyond the scope of this PR, but feel free to open an issue for it.

ndaelman-hu (approved):
Adds a ModelMethod section that semantically connects method and electronic structure. Will be extended to a workflow piece down the line.

Mostly small nitpicky feedback on docs and tests.

ndaelman-hu:
Just as a question: isn't `BrokenSymmetryCenter` semantically closer to a piece in `ModelSystem` than `ModelMethod`.

src/nomad_simulations/schema_packages/model_method.py
EBB2675: sure, i will also make the warning a bit more verbose in an upcoming cleanup commit :+1:

docs/snippets/model_method/bs_dft_example.py
EBB2675: sure, i will also make these more verbose in an upcoming cleanup commit :+1:

src/nomad_simulations/schema_packages/model_method.py
EBB2675: will add a brief explanation :+1:

tests/test_model_method.py
EBB2675: will do :+1:

EBB2675:
> Just as a question: isn't `BrokenSymmetryCenter` semantically closer to a piece in `ModelSystem` than `ModelMethod`.

It is defined by means of a domain-specific method, rather than general physics and chemistry, so i thought its safer to add it to model method

Otherwise model system might be full of method dependent physical spaces from many different potential domains, which can make the distinction difficult down the line

tests/test_model_method.py
EBB2675: will remove this test :+1:

src/nomad_simulations/schema_packages/model_method.py
EBB2675: `self.determinant != 'unrestricted'` warns for both a wrong value and an unset value like `None`

`self.is_spin_polarized is not True` warns for both `False` and an unset value like `None`

ndaelman-hu: Indeed. I just wanted to verify with you that it should also trigger when unset. As in, do you think it necessary for these fields to be set. Anyhow, I just emphasized this point, resolve how you see fit.

EBB2675: aah now i see, yes, we think it is necessary :+1:

PR #369, opened 2026-03-26
remove particle reassignment in `ModelSystem.normalize()`

### LINKED ISSUE #248: Reassess practice of reassigning general ParticleStates
- [ ] check robustness of current reassignment
- [x] see what parser-focused implementation of AtomState/CGBeadState setting would look like in the mapper
- [ ] consider maintenance/correctness of is_atomic() -- could also set is_atomic once in normalizer before reassignment and then simply use the cache

In whatever implementation, need to determine protocol/method for 
- [ ] enable passing of quantities from ParticleState to XState

EBB2675:
## Purpose
This PR removes schema-side reassignment of generic `ParticleState` entries into `AtomsState` or `CGBeadState` during `ModelSystem` normalization, and tightens the `is_atomic()` gate to be strictly type-based.

The old behavior was dangerous because schema normalization was reinterpreting parser-provided data from labels alone. That could silently force generic particles into `CGBeadState`, and it also allowed atomic-only downstream normalization to proceed when labels merely looked atomic. The fallback decision now belongs to parsers, and `is_atomic()` now reflects actual particle-state typing rather than label coincidence.

## Scope

**Included**

- Remove schema-side particle-state reassignment from `ModelSystem.normalize()`
- Remove the internal helper methods used for generic `ParticleState` -> `AtomsState/CGBeadState` conversion
- Change `ModelSystem.is_atomic()` so it returns True only when all relevant particle states are `AtomsState`
- Apply the stricter is_atomic() logic to both root systems and subsystems via particle_indices
- Update `ModelSystem` docstring to reflect the new policy
- Adjust tests to match the new atomic gating behavior
- Remove tests that only existed to validate the old reassignment behavior

## Context & Links

- Related issue: #248 Reassess practice of reassigning general `ParticleStates`
- Companion parser PR: parser-side fallback now performs the `ParticleState` -> `AtomsState/CGBeadState` decision explicitly before schema normalization 

## Reviewer Notes

The key change is policy, not just code removal:

- Schema normalization no longer changes parser-provided particle-state classes.
- is_atomic() is no longer label-based. It is now strictly type-based.
- This means a system is considered atomic only when it is actually represented by AtomsState, not when its labels happen to coincide with element symbols. That is intentional and aligns the schema with parser responsibility.
 
- The old logic was problematic because if a parser emits the following:
```
ModelSystem(
    particle_states=[
        ParticleState(label='B'),
        ParticleState(label='B'),
        ParticleState(label='A'),
    ]
)
```

Particle states were immeditely re-assigned as `AtomsState(chemical_symbol='B')` via schema thanks to element symbol of boron, which would be unwanted if they wanted to label the particle as 'B'.

## Status

- [x] Ready for review

## Breaking Changes

- [x] (explain): Parser flows that previously relied on schema-side reassignment or label-based atomic gating must now provide the intended particle-state type explicitly.
This schema PR has a parser counterpart and should be considered together with it.
https://github.com/FAIRmat-NFDI/nomad-parser-plugins-simulation/pull/166

## Dependencies / Blockers

- [x] None

## Testing / Validation

- [x] Integration tests
- [x] Parser tests


src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1687,19 +1698,20 @@ def to_ase_atoms(
...
+            for ASE conversion
           - positions from the top-level positions quantity (always from ModelSystem, not from representations)
           - periodic boundary conditions and lattice vectors from either ModelSystem directly (if representation_index=None)
             or from the specified alternative representation (if representation_index is an integer)
         """
         logger = self.to_ase_atoms.__annotations__['logger']
         symbols = self.get_symbols()
         if not symbols:
-            logger.error('Cannot generate ASE Atoms without chemical symbols.')
+            logger.error('Cannot generate ASE Atoms without particle label tokens.')
```
ladinesa: I think we should translate into ASE Atoms language here, which uses chemical symbols rather than particles and label tokens? Although in nomad, we generalize this quantity.

EBB2675: good catch, will do now

ladinesa (approved):

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1597,37 +1597,48 @@ def is_atomic(self) -> bool:
...
 
         self._cache['is_atomic'] = is_atomic
         return is_atomic
 
     def get_symbols(self) -> list[str]:
         """
-        Access to particle labels, irrespective of specific child class.
+        Access to particle label tokens, irrespective of specific child class.
+
+        For atomic systems these are typically chemical symbols. For coarse-grained
```
JFRudzinski:
```suggestion
        For atomic systems these must be chemical symbols. For coarse-grained
```

JFRudzinski: right?

JFRudzinski (approved):
Thank you!

PR #385, opened 2026-05-05
canonical-label recognition of `AtomCenteredBasisSet` based on a vendored BSE-name registry

### LINKED ISSUE #383: vendor minimal Basis Set Exchange metadata for basis set normalization and linking
add a small vendored metadata layer from Basis Set Exchange to `AtomCenteredBasisSet`in `nomad-simulations` so parsed basis set names can be canonicalized and linked to a stable external reference (without relying on live API calls at runtime)

EBB2675:
## Purpose

Add a lightweight basis-set name canonicalization layer for `AtomCenteredBasisSet`.
The implementation vendors a minimal registry generated from Basis Set Exchange names,
without requiring `basis_set_exchange` at runtime.

During `AtomCenteredBasisSet.normalize`, the raw parsed `basis_set` is preserved.
If the label matches the vendored canonical name registry unambiguously,
`canonical_basis_set` is filled with the canonical display name. Derived
`canonical_basis_set` metadata is cleared before each normalization lookup to avoid
stale values.

## Scope

**Included**:
- Adds a vendored basis-set name registry generated from BSE canonical names.
- Adds lookup utilities for relaxed, unambiguous label matching against registry keys and canonical names.
- Adds `canonical_basis_set` metadata during `AtomCenteredBasisSet.normalize`.
- Preserves the raw parsed `basis_set` label.
- Adds a `dev-electronic` extra for BSE registry regeneration.
- Adds focused tests for known labels, unknown labels, registry shape, and normalization behavior.

**Out of scope:**
- Numerical basis-function validation.
- Program-specific basis-set variants.
- Curated alias equivalence such as Pople shorthand mapping.
- Runtime dependency on BSE.

## Reviewer Notes

This currently canonicalizes labels only. It does not prove numerical equivalence of
basis-set definitions. Alias support is intentionally omitted from the schema


src/nomad_simulations/schema_packages/utils/basis_set_exchange/__init__.py
ndaelman-hu, line 1: No need to create an `__init__` file for a comment. Pls remove.

src/nomad_simulations/schema_packages/utils/basis_set_exchange/registry.py
```diff
@@ -0,0 +1,93 @@
...
+_DASHES = str.maketrans(
+    {
+        '\u2010': '-',
+        '\u2011': '-',
+        '\u2012': '-',
+        '\u2013': '-',
+        '\u2014': '-',
+        '\u2212': '-',
+    }
+)
```
ndaelman-hu: Seems overkill here and I recommend removing it. This handles unicode to ascii conversion.

I don't think we've encountered such an issue in our parsers before (as most files are already ascii). If we do encounter it, the solution should exist at a more widespread level in the parsers.

If this makes up for the JSON registry, the dashes should simply be standardized there.

pyproject.toml
ndaelman-hu, line 1: Why aren't `bse` and `pylibxc` added as dependencies here?
I would generate a special case, e.g. `dev-electronic` so we can regenerate the JSON files.

src/nomad_simulations/schema_packages/utils/basis_set_exchange/export_registry.py
```diff
@@ -0,0 +1,88 @@
...
+        import basis_set_exchange as bse
+        from basis_set_exchange import misc
+    except ImportError as exc:
+        raise SystemExit(
+            'basis_set_exchange is required only to regenerate this registry. '
+            'Install it in a development environment, e.g. '
+            '`uv run --with basis-set-exchange python -m '
+            'nomad_simulations.schema_packages.utils.basis_set_exchange.export_registry`.'
+        ) from exc
+
```
ndaelman-hu, line 29: The lazy loading is appropriate here, but instructions of the error message aren't. `pyproject.toml` should be extended, see below.

src/nomad_simulations/schema_packages/utils/basis_set_exchange/registry_min.json
ndaelman-hu, line 1: Regarding the `aliases` key, could you elaborate for what scenarios exactly you are future-proofing?
I guess it would be common alternatives used across simulation packages, e.g. 6-31G** vs 6-31G(d,p). But wouldn't this kind of mapping typically be a parser responsibility?

If there really are commonly used aliases that can be preempted, I'd like to see them added here on in an immediate follow-up PR. Else, a bunch of the logic can be simplified here.

We should overall be clear in what counts as an alias and what's a quirk of a code: do 6-31Gdp and 6_31G_dp merit their own alias or are they too specific? I don't want a situation where we continuously update aliases to accommodate a new code package.

ndaelman-hu: Related note, you currently have no approach for how to programmatically add aliases to the registry. This is necessary, as `export_registry.py` will overwrite this file later on.

src/nomad_simulations/schema_packages/utils/basis_set_exchange/build.py
ndaelman-hu, line 1: `build.py` is kinda weird name. This just converts the registry lookup to the schema... I'd merge this with `registry.py`.

ndaelman-hu (commented):
Overall good addition with a similar structure as the libxc.
We should clarify the `aliases` point before merging. The other comments are more so code cleaning.

EBB2675:
@ndaelman-hu Thanks for the detailed review! I addressed the comments by:

- removing the placeholder `__init__.py`
- dropping the Unicode dash normalization
- adding a dedicated `dev-electronic` extra for BSE registry regeneration
- merging the small `build.py` helper into `registry.py`
- removing the unused `aliases` field and alias lookup logic
- updating `export_registry.py` so regenerated JSON keeps the simplified schema and the lazy-import error points to `dev-electronic`
- clearing derived `canonical_basis_set` before normalization lookup and logging exception details if canonicalization fails

(I kept `pylibxc` out of `dev-electronic`. The existing LibXC registry is already vendored, with provenance documented as generated from the LibXC source tree rather than from `pylibxc`)

tests/test_basis_set.py
ndaelman-hu, line 1: Nice edge case addition

EBB2675:
if there is nothing else, i am ready to merge this @JFRudzinski @ndaelman-hu

src/nomad_simulations/schema_packages/utils/basis_set_exchange/registry.py
```diff
@@ -2,29 +2,21 @@
...
+    canonical_name: str
 
 
 def _normalize_label(label: str) -> str:
-    label = unicodedata.normalize('NFKC', label or '').translate(_DASHES)
-    return _IGNORED_LABEL_CHARS.sub('', label.casefold())
+    return _IGNORED_LABEL_CHARS.sub('', (label or '').casefold())
 
 
 @lru_cache(maxsize=1)
```
ndaelman-hu, line 22: The `@lru_cache` decorator will cache exceptions if `registry_min.json` is missing or malformed, causing all subsequent lookups to fail. Consider wrapping the file I/O in a try-except that returns
  an empty dict on failure, preventing the error state from being cached. This ensures graceful degradation rather than repeated exceptions

ndaelman-hu (approved):
Requests were addressed in latest commits.
Just some minor stuff I found more, no need for additional review.

src/nomad_simulations/schema_packages/utils/basis_set_exchange/registry.py
EBB2675: Hmm I checked this both against `functools.lru_cache` behavior and with a minimal cached function that raises on every call. Repeated calls were cache misses and `currsize` stayed `0`, which is expected: `lru_cache` caches successful return values, not raised exceptions. So, if `registry_min.json` is missing or malformed, `_registry()` will raise on each call rather than caching an error state

I'd prefer not to return an empty registry here because a missing or malformed JSON file is a packaging/data issue that should remain visible. The `AtomCenteredBasisSet.normalize()` path already degrades gracefully by preserving `basis_set`, clearing the derived `canonical_basis_set`, and logging canonicalization failures with `exc_info=True`

ndaelman-hu: > I'd prefer not to return an empty registry here because a missing or malformed JSON file is a packaging/data issue that should remain visible.

Sounds reasonable. Should we make checking the JSON integrity part of a (special) test stack then?

> The AtomCenteredBasisSet.normalize() path already degrades gracefully by preserving basis_set

That is the most important part for daily usage.

EBB2675: Good point :+1: I added this to the normal lightweight BSE registry test. It loads `registry_min.json` and checks the expected schema, non-empty string values, and unique keys

(I avoided a regeneration check in regular CI for obvious reasons)

we can revisit whether this belongs in a dedicated/special test stack once we get to restructuring and cleaning up the test suite more broadly

ndaelman-hu: Thx, I'll resolve this thread then

PR #402, opened 2026-06-19
Prune long trajectory

Bernadette-Mohr:
## Purpose
Long trajectories caused parsing to hang indefinitely. Pair generation for molecular RDF calculation failed if codes didn't provide clear molecule type identifiers (Lammps).

### Root cause: incorrect molecule type detection for LAMMPS archives

`get_fragtypes()` in `mdanalysisparser.py` used order-sensitive array comparison to
identify molecule types by composition. For LAMMPS, where atom ordering within a
molecule is not canonical, each molecule was assigned a unique type. For
hexane/cyclohexane (36 molecules), this produced 666 RDF pairs instead of 3,
causing the normalizer to hang.

**Fix:** sort atom type arrays before comparing so composition-based matching is
order-independent.

### Downstream fix: molecule names in LAMMPS hierarchy

`archive_to_universe` reconstructs MDAnalysis moltypes from `mol.name`. The LAMMPS
parser was setting each molecule's name to its integer molecule ID, so even after
`get_fragtypes` produced 2 unique types, the 36 unique IDs reintroduced 36 unique
moltypes at the archive level.

**Fix:** `_create_molecule` accepts a `moltype_name` parameter; `_create_molecule_group`
passes `str(moltype)` so all molecules within a group share the same name — matching
the convention already used by parsers with named topologies (e.g. GROMACS).

### Safety net: frame downsampling for large trajectories

`_get_molecular_rdfs` in `MolecularDynamicsResults` always passed `n_prune=1` for
trajectories where every model system maps to one universe frame. For large
trajectories this makes RDF computation scale linearly with frame count.

**Fix:** when `universe.n_frames > 100`, `n_prune` is increased to
`ceil(n_frames / 100)` so at most ~100 frames are sampled per RDF interval.

### Files changed

| File | Location |
|------|----------|
| `parsers/utils/mdanalysisparser.py` | `get_fragtypes()` |
| `parsers/lammps/parser.py` | `_create_molecule()`, `_create_molecule_group()` |
| `nomad_simulations/workflow/molecular_dynamics.py` | `MolecularDynamicsResults._get_molecular_rdfs()` |

## Status

- [x] Ready for review

## Breaking Changes

- [x] None

## Dependencies / Blockers

- [x] None

## Testing / Validation

- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
+++ b/src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
@@ -1172,12 +1172,19 @@ class MolecularDynamicsResults(SerialWorkflowResults):
             return self.radial_distribution_functions
 
         n_traj_split = 10  # number of intervals to split trajectory into for averaging
+        _ANALYSIS_MAX_FRAMES = 100
         try:
             n_prune = int(
                 self._universe.trajectory.n_frames / len(archive.data.model_system)
             )
         except Exception:
             n_prune = 1
+        n_universe_frames = self._universe.trajectory.n_frames
+        if n_universe_frames > _ANALYSIS_MAX_FRAMES:
+            n_prune = max(
+                n_prune,
+                int(np.ceil(n_universe_frames / _ANALYSIS_MAX_FRAMES)),
+            )
 
         interval_indices: list[
             np.ndarray
--- a/src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
+++ b/src/nomad_simulations/schema_packages/utils/molecular_dynamics.py
@@ -495,13 +495,13 @@ def archive_to_universe(
             _charges_list.append(0.0)
     if _missing_masses:
         LOGGER.warning(
-            '%d particle(s) missing mass; atomic particles fall back to ASE defaults, '
-            'CG particles default to 0.0.',
-            _missing_masses,
+            'particles missing mass; atomic fall back to ASE defaults, CG default to 0.0',
+            count=_missing_masses,
         )
     if _missing_charges:
         LOGGER.warning(
-            '%d particle(s) missing charge; defaulting to 0.0.', _missing_charges
+            'particles missing charge; defaulting to 0.0',
+            count=_missing_charges,
         )
 
     masses = np.array(_masses_list)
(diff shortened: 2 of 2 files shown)


JFRudzinski:
I don't have enough time to look deep enough into this right now, but I would like to discuss with you soon ... there is already some protections against long traj / large systems in the specific calculation of derived properties, but maybe it needs to be refined, or added to

src/nomad_simulations/schema_packages/workflow/molecular_dynamics.py
```diff
@@ -1172,12 +1172,19 @@ def _get_molecular_rdfs(
             return self.radial_distribution_functions
 
         n_traj_split = 10  # number of intervals to split trajectory into for averaging
+        _ANALYSIS_MAX_FRAMES = 100
         try:
             n_prune = int(
```
JFRudzinski, line 1177: Just reminding myself what is happening here:

data.model_system is already pruned compared to the raw trajectory based on a maximum cummulative number of atoms -- https://github.com/FAIRmat-NFDI/nomad-parser-plugins-simulation/blob/develop/src/nomad_simulation_parsers/parsers/utils/mdparserutils.py#L116

which then determines the archive_sampling_rate -- 
https://github.com/FAIRmat-NFDI/nomad-parser-plugins-simulation/blob/develop/src/nomad_simulation_parsers/parsers/utils/mdparserutils.py#L185

max_cum_atoms is set to 2.5M somewhat arbitrarily I think

Now you are adding a further restriction, this time directly in the schema package to ensure that n_frames is also lower than 100

Also it should be noted that there is already a restriction of the total number of pairs per frame for any rdf -- https://github.com/FAIRmat-NFDI/nomad-simulations/blob/develop/src/nomad_simulations/schema_packages/utils/molecular_dynamics.py#L812

I'm still a little unsure why you have the issue ... without the n_prune, how many frames do you have? If the number of molecules is already below the threshold, I don't see how calculating over many frames matter, unless n_frames is orders of magnitude greater than 100? But then I would expect the max_cum_mols restriction to do the pruning for us

Could you provide some specifics about the system you are having issues with in terms of n_frames raw, n_frames in data, n_atoms per frame, n_mols per moltype per frame?

JFRudzinski:
@Bernadette-Mohr What happened with this one? It's no longer relevant?

Bernadette-Mohr:
Yes, it is not relevant anymore. The issues I had addressed here are either non-issues altogether, or better handled elsewhere.

PR #405, opened 2026-07-06
Use SeeKpath for high-symmetry k-point resolution

ndaelman-hu:
## Purpose

Replace the ASE-based high-symmetry k-point resolution with SeeKpath. ASE's special-point generator classifies the bare cell from its metrics alone and cannot accept an external symmetry analysis, so it can disagree with the spglib-derived symmetry NOMAD stores; its `BravaisLattice` constructors additionally enforce Setyawan-Curtarolo parameter conventions that standardized cells do not generally satisfy - the source of the three failure modes reported in #403. SeeKpath keys k-point selection to the space group from the same spglib analysis behind NOMAD's stored symmetry data, so classification and k-points cannot disagree by construction.

## Scope

**Included**

- `resolve_high_symmetry_points` reimplemented on `seekpath.get_path` (HPKOT recipe)
- Transformation of SeeKpath's coordinates into the input cell's reciprocal basis (fractional coordinates of `KSpace.reciprocal_lattice_vectors`), including rotated-frame handling via spglib's `std_rotation_matrix`
- New dependency `seekpath>=2.1.0`
- Tests: the three reported bug scenarios, three coordinate-basis cases (conventional BCC, axis-permuted FCC primitive, rigidly rotated BCC primitive), and three reference structures distilled to their spglib primitive cells

**Out of scope**

- Ecosystem adoption of HPKOT label conventions in parsers and legacy workflows (e.g. suffixed labels such as `X_1`)
- Any `KLinePath` label translation layer for backward compatibility

## Context & Links

- Base PR: #403
- Alternative implementation kept for reference: `fix-kmesh-use-ase-get-special-points` (ASE `get_special_points`); to be closed once this is merged

## Reviewer Notes

- Labels follow the HPKOT standard and can differ from Setyawan-Curtarolo where the classifications differ (e.g. a conventional cubic cell with a body-centered basis yields BCC labels); point sets are more comprehensive.
- Validated against all eight provided reference structures: every structure resolves correctly (checked against an independent primitive-cell reference via rotation-invariant norms and Gram matrices). ASE's metric classification contradicts spglib on four of them, including two `Imma` structures that land in different lattice classes from sub-0.1 degree metric noise.

## Status

- [ ] Draft / In progress
- [x] Ready for review
- [ ] Blocked (explain):

## Breaking Changes

- [ ] None
- [x] (explain): `high_symmetry_points` labels follow HPKOT instead of Setyawan-Curtarolo; consumers matching specific label names may need updating.

## Dependencies / Blockers

- [x] None
- [ ] Open design questions (explain):
- [ ] External dependencies (explain):

## Testing / Validation

- [x] Unit tests
- [x] Manual testing (explain): all eight provided reference structures validated against an independent primitive-cell SeeKpath reference; three added as regression fixtures.


src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -409,99 +254,99 @@ def resolve_high_symmetry_points(
...
+                    logger.warning('SeeKpath returned empty special points dictionary.')
+                    return None
+
+                # SeeKpath expresses `point_coords` in fractional coordinates of the
+                # reciprocal basis of its own standardized primitive cell, which may
+                # differ from the input cell by a lattice-vector choice (e.g., a
+                # conventional or permuted setting) and, after spglib idealization,
+                # by a rigid rotation. Transform each point to the input cell's
+                # reciprocal basis, which is what `KSpace.reciprocal_lattice_vectors`
+                # and the `high_symmetry_points` schema contract refer to.
```
ndaelman-hu: Do you mean that the quantities define it as such?

src/nomad_simulations/schema_packages/numerical_settings.py
```diff
@@ -409,99 +254,99 @@ def resolve_high_symmetry_points(
...
+                    rotation = np.array(dataset.std_rotation_matrix)
+
+                # k_frac_input = ((k_frac_prim @ B_prim) @ R) @ B_input^-1
+                to_input_basis = (
+                    reciprocal_primitive @ rotation @ np.linalg.inv(reciprocal_input)
+                )
+                special_points = {
+                    label: (np.array(coords) @ to_input_basis).tolist()
+                    for label, coords in special_points.items()
+                }
```
ndaelman-hu: Could we split this into its own helper function pls?

tests/test_numerical_settings.py
```diff
@@ -176,31 +174,34 @@ def test_resolve_high_symmetry_points(self, model_systems_input, expected_result
...
 
-        When ASE detects a lattice type, it reorders vectors to satisfy
-        crystallographic conventions (e.g., a < b < c for orthorhombic).
-        This test verifies that k-points are successfully generated when
-        ASE and spglib agree on the lattice type.
+        SeeKpath uses spglib's symmetry analysis to generate comprehensive
+        k-point sets including intermediate points with numeric suffixes.
         """
         from ase import Atoms
         from ase.cell import Cell
```
ndaelman-hu: 1. the naming was not updated
2. can't this be merged with the test above?

tests/test_numerical_settings.py
```diff
@@ -312,34 +299,50 @@ def test_resolve_high_symmetry_points_consistency_check_agreement(self, log_outp
             model_systems=[model_system], logger=logger
         )
 
-        # Should get cubic k-points
+        # Should get cubic k-points (SeeKpath uses HPKOT naming: H/P/N instead of M/R/X)
```
ndaelman-hu: Is the k-point nomeclature optional? Does SeeKpath provide an option to retain the older convention?

ndaelman-hu (commented):
rethink the test structure and see whether aggregation make ssense

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: Yes: the `high_symmetry_points` quantity description defines the values as "in units of `reciprocal_lattice_vectors`". Reworded the comment to reference the quantity description directly. The description itself still documented the old ASE-based behavior, so it was rewritten as well.

tests/test_numerical_settings.py
ndaelman-hu: No. seekpath only implements `recipe='hpkot'` (any other value raises), so there is no option for Setyawan-Curtarolo labels. `with_time_reversal` does not affect naming either; it only extends the path for non-centrosymmetric groups when set to `False`. A label translation layer would have to live on the NOMAD side, which this PR declares out of scope.

src/nomad_simulations/schema_packages/numerical_settings.py
ndaelman-hu: Done in 15a12e4: extracted as `KSpaceFunctionalities._to_input_reciprocal_basis`, with the transformation documented in its docstring.

tests/test_numerical_settings.py
ndaelman-hu: Done in 15a12e4: the three SimpleNamespace integration tests are merged into a single parametrized `test_resolve_high_symmetry_points_seekpath` with exact label-set assertions, sharing a `make_kpoint_model_system` helper with the coordinate tests. Degenerate-input cases (empty list, non-representative system) were added to `test_resolve_high_symmetry_points`.

PR #414, opened 2026-07-16
fix band gap extraction minor bugs

### LINKED ISSUE #413: DOS-derived band gap is always wrong: inverted sign in extract_band_gap and LUMO stored under HOMO cache key
There are two bugs in `ElectronicDensityOfStates` (`spectral_profile.py`) that break every band gap derived from a DOS.

1. `extract_band_gap` computes `homo - lumo` and clamps negative values to 0. Since HOMO < LUMO in any gapped system, the difference is always negative, so every insulator gets a gap of 0.0. The sibling `ElectronicEigenvalues.extract_band_gap` does it the right way round (`lumo - homo`).

2. In `resolve_energies_origin`, the loop that searches for the lowest unoccupied energy stores its result in `m_cache['highest_occupied_energy']` instead of `'lowest_unoccupied_energy'` (line 306). The LUMO is therefore never cached, so on the DOS-only path `extract_band_gap` returns None, and the HOMO gets overwritten with the conduction band edge, which also makes `energies_origin` wrong.

I checked both with a synthetic gapped DOS (band edges at ±1 eV, Fermi level at 0): after `resolve_energies_origin` the cache holds +0.99 eV as the HOMO, no LUMO, and `energies_origin` is +0.99 eV instead of -1 eV. Seeding the cache with HOMO = -5 eV and LUMO = -3 eV gives a gap of 0.0 instead of 2 eV.

Fix is two lines: use `lumo - homo` in `extract_band_gap`, and store the LUMO search result under the right cache key.

EBB2675:
# Title

Fix DOS-derived band gap: sign inversion and wrong cache key

## Purpose

Band gaps derived from the electronic DOS were always 0 or missing. `extract_band_gap` computed `homo - lumo` (negative for any gapped system, so clamped to 0), and the LUMO search in `resolve_energies_origin` stored its result under the HOMO cache key, so the LUMO was never cached and `energies_origin` pointed at the conduction band edge.

## Scope

**Included**

- `spectral_profile.py`: compute `lumo - homo` in `extract_band_gap` (consistent with `ElectronicEigenvalues`), store the LUMO search result under `lowest_unoccupied_energy`, docstring updated.

## Reviewer Notes

DOS-derived gap values and `energies_origin` change for all affected entries; anything processed with the old code needs re-normalization. No existing test covered this path, which is why both bugs went unnoticed.

## Status

- [x] Ready for review

## Breaking Changes

- [x] None

## Dependencies / Blockers

- [x] None

## Testing / Validation

- [x] Manual testing (explain): reproduced both bugs with a synthetic gapped DOS (band edges ±1 eV, Fermi at 0), verified the fix gives the right HOMO/LUMO and a ~2 eV gap.

### DIFF
--- a/src/nomad_simulations/schema_packages/properties/spectral_profile.py
+++ b/src/nomad_simulations/schema_packages/properties/spectral_profile.py
@@ -365,7 +365,7 @@ class ElectronicDensityOfStates(DOSProfile):
         band_gap = None
         homo = self.m_cache.get('highest_occupied_energy')
         lumo = self.m_cache.get('lowest_unoccupied_energy')
-        if homo and lumo:
+        if homo is not None and lumo is not None:
             band_gap = ElectronicBandGap()
             band_gap.is_derived = True
             band_gap.physical_property_ref = self
--- a/src/nomad_simulations/schema_packages/properties/spectral_profile.py
+++ b/src/nomad_simulations/schema_packages/properties/spectral_profile.py
@@ -290,7 +290,6 @@ class ElectronicDensityOfStates(DOSProfile):
                     except IndexError:
                         break
                     if value > configuration.dos_intensities_threshold:
-                        idx = idx if idx == idx_descend else idx + 1
                         self.m_cache['highest_occupied_energy'] = energies_points[idx]
                         break
                     idx -= 1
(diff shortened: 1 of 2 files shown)


JFRudzinski (approved):

ladinesa (approved):

EBB2675:
FYI @ndaelman-hu 

please let me know if you'd have any additions to it

Issue #423, opened 2026-07-23
SCF convergence quantities have inconsistent definitions and physically unsafe normalization fallbacks


EBB2675:
https://github.com/FAIRmat-NFDI/nomad-simulations/blob/develop/src/nomad_simulations/schema_packages/outputs.py
 
- [ ]  `delta_density_rms` description is copy-pasted from `delta_potential_rms`. We need to define precisely: (a) the residual  (is it ρₙ − ρₙ₋₁?); (b) the norm used to reduce ρ(r) to a scalar (e.g. √∫|Δρ|²d³r vs. a discrete grid sum); and (c) the normalization (per grid point, per electron, per volume?), since this determines both the value and the units. `coulomb` only makes sense for an integrated charge residual; document that assumption or change the unit to a charge-density.

- [ ]  `delta_force_abs` is populated with final per-atom force magnitudes rather than force changes between successive SCF iterations:
```python
force_values = self.total_forces[-1].value
force_norms = ((force_values**2).sum(axis=1)) ** 0.5
return force_norms
```

- [ ] See #424 : `_compute_energy_deltas` incorrectly treats repeated generic `TotalEnergy` properties in `Outputs.total_energies` as an ordered SCF iteration history, instead of deriving the differences from `scf_steps.energies_total`.

```python
energy_values = [e.value for e in self.total_energies]
for i in range(1, len(energy_values)):
    delta = np.abs(energy_values[i] - energy_values[i - 1])
```

minimal snippet:

```python
import numpy as np
from nomad.units import ureg
from structlog import get_logger

from nomad_simulations.schema_packages.outputs import Outputs, SCFSteps
from nomad_simulations.schema_packages.properties import TotalEnergy, TotalForce

outputs = Outputs(
    total_energies=[
        TotalEnergy(label="DFT", value=1.0 * ureg.joule),
        TotalEnergy(label="DFT + dispersion", value=2.0 * ureg.joule),
    ],
    total_forces=[
        TotalForce(value=np.array([[3.0, 4.0, 0.0]]) * ureg.newton),
    ],
    scf_steps=SCFSteps(),
)

outputs.normalize(archive=None, logger=get_logger())

print(outputs.scf_steps.delta_energies_total)
print(outputs.scf_steps.delta_force_abs)
```
gives 
```
[1.0] joule
[5.0] newton
```
[...]

PR #424, opened 2026-07-23
Fix SCF energy deltas to use SCF step energies

EBB2675:
## Purpose

Fix SCF energy-delta normalization so `SCFSteps.delta_energies_total` is derived from the SCF iteration energy history, not from generic output energy sections.

### Included

* Make `_compute_energy_deltas` source-agnostic by accepting an explicit `energies` argument.
* Call `_compute_energy_deltas` from `Outputs.normalize` with `scf_steps.energies_total`.
* Update the convergence tests to:

  * provide the SCF iteration history through `scf_steps.energies_total`;
  * verify that generic `Outputs.total_energies` entries are not treated as SCF iterations;
  * verify the resulting convergence behavior.
* Update tutorial snippets that demonstrate automatic SCF energy-delta population.
* Clarify the schema semantics of:

  * `Outputs`;
  * `TotalEnergy`;
  * `SCFSteps.energies_total`;
  * `SCFSteps.delta_energies_total`.
* Add a conceptual geometry-optimization example distinguishing:

  * SCF iterations within one configuration;
  * ordered workflow configurations represented by separate `WorkflowOutputs` sections.
* Update the schema-documentation generator and template so these explanations remain part of the generated documentation.

### Out of scope

* Changing force-convergence fallback behavior.
* Changing `_rms` SCF quantity names, units, or descriptions.

## Context and links

Part of #423.

Previously, repeated `Outputs.total_energies` entries could be interpreted as an ordered sequence of SCF iterations. This is unsafe because `total_energies` may instead contain separately labelled energy variants, such as DFT and dispersion-corrected total energies.

The schema now makes the relevant sequence axes explicit:

* `SCFSteps.energies_total` contains the ordered SCF iteration energies within one system configuration.
* Separate `Outputs` or `WorkflowOutputs` sections represent separate system configurations or workflow steps.
* Repeated `Outputs.total_energies` entries do not currently have a defined ordering or sequence meaning.

## Reviewer notes

The change remains intentionally narrow. Existing `delta_force_abs` behavior and `_rms` metadata are left unchanged for separate follow-up work.

Seven parsers map `SCFSteps.delta_energies_total` directly: VASP, Crystal, exciting, ABINIT, FHI-aims, Quantum ESPRESSO/PWscf, and AMS. Together, these contain ten mapping declarations: three for VASP, two for Quantum ESPRESSO/PWscf, and one for each of the others.

Each parser sets the quantity explicitly, so the normalizer inference, which only runs when the field is unset, was already a no-op for them.

GPAW is the eighth parser that constructs `scf_steps`, and the only one that maps neither `delta_energies_total` nor `energies_total`. Its `get_energies` method returns one total energy plus its contributions. Consequently:

* `Outputs.total_energies` contains only one entry, so the previous `len(total_energies) >= 2` guard returned `None`;
* `scf_steps.energies_total` remains unset, so the new inference also returns `None`.

GPAW's `delta_energies_total` is therefore unset both before and after this change.

FHI-aims similarly returns one total energy plus components rather than sibling total-energy values, and maps the SCF energy delta directly regardless.

## Status

* [x] Ready for review

## Breaking changes

* [x] `delta_energies_total` is no longer inferred from `Outputs.total_energies`.
* [x] It is inferred only when `scf_steps.energies_total` contains at least two values.
* [x] No current parser is affected.

## Dependencies and blockers

* [x] None

## Testing and validation

* [x] Unit tests
* [x] Tutorial snippets and docs updated

### DIFF
--- a/docs/schema/outputs.md
+++ b/docs/schema/outputs.md
@@ -2,6 +2,37 @@
 
 **Purpose:** Base output structure and common property definitions
 
+## Notes
+
+One `Outputs` section represents the calculated properties of one system
+configuration, identified through `model_system_ref`. A sequence of
+configurations, such as geometry-optimization steps, is represented using
+multiple output sections, normally `WorkflowOutputs`, ordered by
+`WorkflowOutputs.step`.
+
+One `TotalEnergy` represents the converged total energy of that configuration.
+Energy components belong inside that `TotalEnergy` through its `contributions`
+subsections. The listed contributions do not have to be exhaustive, so the total
+energy is not necessarily equal to their sum. Although `Outputs.total_energies`
+has `repeats=True`, the schema currently defines neither an ordering nor a
+sequence meaning for repeated entries.
+
+`SCFSteps.energies_total` is the ordered sequence of total energies from the SCF
+iterations within one configuration. `SCFSteps.delta_energies_total` contains
+differences between consecutive values from that SCF sequence.
+
+Conceptual geometry-optimization layout:
+
+```text
+WorkflowOutputs(step=0)
+    TotalEnergy for configuration 0
+    SCFSteps.energies_total for the SCF iterations of configuration 0
+
+WorkflowOutputs(step=1)
+    TotalEnergy for configuration 1
+    SCFSteps.energies_total for the SCF iterations of configuration 1
+```
+
 
 ## Relationship map
 
@@ -74,7 +105,7 @@ classDiagram
 
 | Section | Description | MetaInfo |
 |---|---|---|
-| `Outputs` | Output properties of a simulation. | [Open in MetaInfo browser](https://nomad-lab.eu/prod/v1/develop/gui/analyze/metainfo/nomad_simulations/section_definitions@nomad_simulations.schema_packages.outputs.Outputs){:target="_blank"} |
+| `Outputs` | Output properties of one system configuration. | [Open in MetaInfo browser](https://nomad-lab.eu/prod/v1/develop/gui/analyze/metainfo/nomad_simulations/section_definitions@nomad_simulations.schema_packages.outputs.Outputs){:target="_blank"} |
 
 | Quantity | Type | Description |
 |---|---|---|
@@ -89,8 +120,8 @@ classDiagram
 
 | Quantity | Type | Description |
 |---|---|---|
-| `energies_total` | m_float64(float) (shape: ['*']) | Total energy at each SCF step. |
-| `delta_energies_total` | m_float64(float) (shape: ['*']) | <details><summary>Absolute change of total energy between consecutive SCF steps.</summary>Absolute change of total energy between consecutive SCF steps. When<br>derived from `energies_total`, `delta_energies_total[i] =<br>abs(energies_total[i + 1] - energies_total[i])`, so N SCF energies<br>produce N - 1 energy deltas.</details> |
+| `energies_total` | m_float64(float) (shape: ['*']) | Ordered sequence of total energies from the SCF iterations within one system configuration. |
+| `delta_energies_total` | m_float64(float) (shape: ['*']) | <details><summary>Absolute change of total energy between consecutive SCF steps.</summary>Absolute change of total energy between consecutive SCF steps. When<br>derived from `energies_total`, the values follow<br>$\Delta E_i = \lvert E_{i+1} - E_i \rvert$, so N SCF energies<br>produce N - 1 energy deltas.</details> |
 | `delta_potential_rms` | m_float64(float) (shape: ['*']) | Root mean square of change of potential energy at each SCF step. |
 | `delta_density_rms` | m_float64(float) (shape: ['*']) | Root mean square of change of potential energy at each SCF step. |
 | `delta_wavefunction_rms` | m_float64(float) (shape: ['*']) | Root mean square of change of wavefunction coefficients at each SCF step. Dimensionless quantity representing convergence of orbital coefficients. |
--- a/docs/schema/outputs.md
+++ b/docs/schema/outputs.md
@@ -91,10 +91,10 @@ classDiagram
 |---|---|---|
 | `energies_total` | m_float64(float) (shape: ['*']) | Total energy at each SCF step. |
 | `delta_energies_total` | m_float64(float) (shape: ['*']) | Absolute change of total energy at each SCF step. |
-| `delta_potential_rms` | m_float64(float) (shape: ['*']) | Root mean square of change of potential energy at each SCF step. |
-| `delta_density_rms` | m_float64(float) (shape: ['*']) | Root mean square of the integrated charge-density residual at each SCF step. |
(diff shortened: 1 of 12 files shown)


src/nomad_simulations/schema_packages/outputs.py
```diff
@@ -280,16 +283,20 @@ def set_model_method_ref(self) -> ModelMethod | None:
 
     def _compute_energy_deltas(self, logger: 'BoundLogger'):
```
ladinesa: I suggest making this a class method instead to calculate energy differences from various sources

EBB2675:
this is ready to be reviewed @JFRudzinski @ladinesa @ndaelman-hu

src/nomad_simulations/schema_packages/outputs.py
EBB2675: done

docs/snippets/tutorials/part2_foundations/assignment_2_7.py
```diff
@@ -30,9 +29,9 @@
 )
 simulation.model_method.append(method)
 
-outputs = Outputs(scf_steps=SCFSteps())
-for value in [1.0, 1.5, 2.0, 2.1, 2.101]:
-    outputs.total_energies.append(TotalEnergy(value=value * ureg.eV))
+outputs = Outputs(
+    scf_steps=SCFSteps(energies_total=np.array([1.0, 1.5, 2.0, 2.1, 2.101]) * ureg.eV)
```
ndaelman-hu: you can leave out `np.array`, `pint.Quantity` already upcasts from lists

docs/snippets/tutorials/part2_foundations/normalization_exercise_2.py
```diff
@@ -4,15 +4,13 @@
...
 
 logger = utils.get_logger(__name__)
 
 # Pattern based on tests/workflow/test_convergence_targets.py.
-outputs = Outputs(scf_steps=SCFSteps())
-outputs.total_energies.append(TotalEnergy(value=1.0 * ureg.eV))
-outputs.total_energies.append(TotalEnergy(value=1.5 * ureg.eV))
-outputs.total_energies.append(TotalEnergy(value=2.1 * ureg.eV))
+outputs = Outputs(
+    scf_steps=SCFSteps(energies_total=np.array([1.0, 1.5, 2.1]) * ureg.eV)
```
ndaelman-hu: same comment

docs/schema/outputs.md
ndaelman-hu, line 1: Thank you for disambiguating `SCFSteps.energies_total`. Given that this error stems from some semantic vagueness, I would also extend the documentation for the latter and provide an example of how each should look like in a `GeometryOptimization`.

Based on my understanding, the explanation for `TotalEnergy` would go something like:

> A `TotalEnergy` is the converged total energy of a single configuration (one `Outputs`): a value together with an internal `contributions` breakdown of `BaseEnergy` terms. The total is not required to equal the sum of its contributions, because the listed contributions need not be exhaustive. This decomposition is held within one `TotalEnergy`, not distributed across repeated entries. `Outputs.total_energies` permits repetition (`repeats=True`), but the schema defines neither an ordering nor a purpose for multiple entries. The defined sequence axes are `SCFSteps.energies_total` (SCF iterations within a configuration) and the list of Outputs sections (workflow steps, ordered by WorkflowOutputs.step).

and for the `Outputs`

> Each `Outputs` section holds the properties of one system configuration (the `ModelSystem` given by `model_system_ref`); a sequence of configurations is represented as multiple `Outputs` sections, not by repeated subsections within one.

@JFRudzinski does this match how you use it in MD?

src/nomad_simulations/schema_packages/outputs.py
```diff
@@ -60,7 +60,10 @@ class SCFSteps(ArchiveSection):
         type=float,
         unit='joule',
         description="""
-        Absolute change of total energy at each SCF step.
+        Absolute change of total energy between consecutive SCF steps. When
+        derived from `energies_total`, `delta_energies_total[i] =
+        abs(energies_total[i + 1] - energies_total[i])`, so N SCF energies
```
ndaelman-hu: We often times use latex or unicode to write down math formulas. This is legible, but I would keep the style consistent.

src/nomad_simulations/schema_packages/outputs.py
```diff
@@ -278,23 +281,27 @@ def set_model_method_ref(self) -> ModelMethod | None:
                 return model_methods[-1]
         return None
 
-    def _compute_energy_deltas(self, logger: 'BoundLogger'):
+    @staticmethod
+    def _compute_energy_deltas(energies, logger: 'BoundLogger'):
```
ndaelman-hu, line 296: Why the change in interface here?

JFRudzinski (commented):
Looks good to me, but let's wait for Nathan to sign off please

tests/workflow/test_convergence_targets.py
```diff
@@ -724,6 +720,26 @@ def test_energy_delta_computed_from_total_energies(self, archive, logger):
...
+                energies_total=np.array([10.0, 10.1, 10.3]) * ureg.joule
+            ),
+        )
+        archive.data.outputs = [outputs]
+
+        outputs.normalize(archive, logger)
+
+        assert outputs.scf_steps.delta_energies_total is not None
+        assert list(outputs.scf_steps.delta_energies_total.magnitude) == pytest.approx(
+            [0.1, 0.2]
```
ndaelman-hu: Okay, the choice of numbers here is very  confusing, as it appear to compare with `TotalEnergy`s. Would change the deltas to avoid overlapping0.

tests/workflow/test_convergence_targets.py
```diff
@@ -696,20 +696,16 @@ def test_single_path_backwards_compatible(self, archive, logger):
         is_reached = energy_target.normalize(archive, logger)
         assert is_reached is True
 
-    def test_energy_delta_computed_from_total_energies(self, archive, logger):
-        """Test that delta_energies_total is computed from total_energies during normalization."""
+    def test_energy_delta_computed_from_scf_energies_total(self, archive, logger):
```
ndaelman-hu, line 699: Merge with the test added below

ndaelman-hu (commented):
This PR re-alings the `SCFSteps.delta_energies_total` with the semantics of the schema. I don't think we got to fully clarify it yesterday, but we can handle that here in-PR.

I would also update the docs for `TotalEenrgy` and `Outputs`, at least. That aside, some minor code tweaks.

src/nomad_simulations/schema_packages/outputs.py
EBB2675: The interface was changed following Alvin's suggestion to make the calculation source-agnostic. Passing the energy sequence explicitly ensures that normalization uses `scf_steps.energies_total` and avoids coupling the helper to `Outputs.total_energies`. I removed the lambda and kept the explicit helper call for clarity

PR #425, opened 2026-07-23
fix `PlaneWaveBasisSet.compute_cutoff_radius()`

### LINKED ISSUE #422: should plane-wave cutoff radius use hbar/ħ instead of h?
`PlaneWaveBasisSet.compute_cutoff_radius()` currently derives `cutoff_radius` from `cutoff_energy` using Planck's constant `h`:

```python
np.sqrt(2 * m_e * cutoff_energy) / h
```

For a plane wave written as $e^{i\mathbf{k}\cdot\mathbf{r}}$, I would expect this conversion to use $\hbar$:

$$
k_\mathrm{cut} = \frac{\sqrt{2m_eE_\mathrm{cut}}}{\hbar}.
$$

This'd be consistent with `KSpace.reciprocal_lattice_vectors`, which are documented as including the $2\pi$ prefactor.

Should this function use `ħ`/`hbar` instead of `h`? As written, only `cutoff_radius` values derived from `cutoff_energy` through this helper would be smaller by a factor of $2\pi$

EBB2675:
## Purpose

Fix `PlaneWaveBasisSet.compute_cutoff_radius()` to use `hbar` instead of `h` when converting cutoff energy to reciprocal cutoff radius. For plane waves written as `e^{i k·r}`, using `h` made derived `cutoff_radius` values too small by a factor of `2 pi`.

## Scope

**Included**

- Use `hbar` in `compute_cutoff_radius()`.
- Update cutoff test expectations for `500 eV` from `1.823 / angstrom` to `11.456 / angstrom`.

## Context & Links

- Consistent with `KSpace.reciprocal_lattice_vectors`, which include the `2 pi` prefactor.
- Consistent with the documented plane-wave form `e^{i k·r}`.

## Reviewer Notes

Please verify that the `cutoff_radius` convention is intended to follow the plane-wave form in the docstrings (e^{i\mathbf{k}\cdot\mathbf{r}}). Under this convention, `cutoff_radius` is the wavevector magnitude (k_{\mathrm{cut}}), so the conversion from `cutoff_energy` requires (\hbar). Using (h) would instead correspond to spatial frequency in cycles per unit length and would give values smaller by a factor of (2\pi).

## Status

- [x] Ready for review

## Breaking Changes

- [x] Derived `cutoff_radius` values from `cutoff_energy` increase by a factor of `2 pi`.

## Dependencies / Blockers

- [x] None

## Testing / Validation

- [x] Unit tests

### DIFF
--- a/src/nomad_simulations/schema_packages/basis_set.py
+++ b/src/nomad_simulations/schema_packages/basis_set.py
@@ -101,7 +101,7 @@ class PlaneWaveBasisSet(BasisSetComponent, KMesh):
         type=np.float64,
         unit='1/meter',
         description="""
-        Cutoff radius of the plane-wave basis set; the reciprocal-space dual to `cutoff_energy`.     
+        Cutoff radius of the plane-wave basis set; the reciprocal-space dual to `cutoff_energy`.
         Uses the 2*pi-inclusive convention of `KSpace.reciprocal_lattice_vectors`.
         """,
     )
--- a/src/nomad_simulations/schema_packages/basis_set.py
+++ b/src/nomad_simulations/schema_packages/basis_set.py
@@ -116,6 +116,9 @@ class PlaneWaveBasisSet(BasisSetComponent, KMesh):
         The cutoff radius is the wavevector magnitude `k_cut` of a plane wave `e^{i k·r}`,
         obtained from the cutoff energy as `k_cut = sqrt(2 * m_e * E_cut) / hbar`. It follows
         the same 2*pi-inclusive convention as `KSpace.reciprocal_lattice_vectors`.
+"""
+        Compute `cutoff_radius` from `cutoff_energy` as the plane-wave wavevector magnitude
+        `k_cut = sqrt(2 * m_e * E_cut) / hbar` (angular wavevector, 2*pi-inclusive).
         """
         if cutoff_energy is None:
             return None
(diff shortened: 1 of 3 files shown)


src/nomad_simulations/schema_packages/basis_set.py
ndaelman-hu, line 106: Since the whole `h/ħ` ambiguity came from the convention being implicit, could we make the docstring state it explicitly? That way the intended meaning of `cutoff_radius` is self-documenting.


```suggestion
        description="""
        Cutoff radius of the plane-wave basis set; the reciprocal-space dual to `cutoff_energy`.     
        Uses the 2*pi-inclusive convention of `KSpace.reciprocal_lattice_vectors`.
        """,
```

ndaelman-hu, line 120:
```suggestion
"""
        Compute `cutoff_radius` from `cutoff_energy` as the plane-wave wavevector magnitude
        `k_cut = sqrt(2 * m_e * E_cut) / hbar` (angular wavevector, 2*pi-inclusive).
        """
```

ndaelman-hu (approved):
Thanks, the fix looks correct — value checks out against `k = √(2mₑE)/ħ`.

ndaelman-hu:
@EBB2675 Sorry, that old comment was posted too soon. I added the proper suggestions now as a review. Spiritually the same, though.

src/nomad_simulations/schema_packages/basis_set.py
EBB2675: this one i have applied when i saw your first comment, the other one im applying from the suggestion now :+1:

PR #426, opened 2026-07-24
Fix CI failures caused by Ruff

### LINKED ISSUE #427: CI failures caused by unpinned dependency updates
CI reruns started failing without code changes because Ruff 0.16.0 introduced new checks and pymatgen-core 2026.7.24 lacked compatible Linux/Python distributions

EBB2675:
- `include = ["*.py"]` keeps Ruff focused on Python files because Ruff 0.16 started formatting Markdown code blocks.
- `PLR0917` is ignored because Ruff 0.16 newly enabled it, causing 22 failures in existing APIs. Refactoring those signatures is outside this CI-fix PR and could be breaking.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,6 @@ docs = [
 ]
 
 [tool.uv]
-constraint-dependencies = ["pymatgen-core!=2026.7.24"]
 build-constraint-dependencies = ["wheel<0.46"]
 
 [tool.ruff]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,7 @@ dev = [
     'pytest-timeout>=1.4.2',
     'pytest-cov>=2.7.1',
     'pytest-asyncio>=0.21.0',
-    'ruff>=0.6',
+    'ruff==0.15.22',
     'structlog>=1.0',
     'typing-extensions>=4.12',
 ]
@@ -68,6 +68,7 @@ docs = [
 ]
 
 [tool.uv]
+constraint-dependencies = ["pymatgen-core<2026.7.24"]
 build-constraint-dependencies = ["wheel<0.46"]
 
 [tool.ruff]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -51,7 +51,7 @@ dev = [
     'pytest-timeout>=1.4.2',
     'pytest-cov>=2.7.1',
     'pytest-asyncio>=0.21.0',
-    'ruff>=0.6',
+    'ruff==0.15.22',
```
JFRudzinski: How do we manage the updates in the future? Is there some reference from one of the Area D repos for this?

ladinesa: nomad-FAIR does not pin the version.

EBB2675:
hi again @ladinesa , @JFRudzinski 

i updated the PR:

- `include = ["*.py"]` keeps Ruff focused on Python files because Ruff 0.16 started formatting Markdown code blocks.
- `PLR0917` is ignored because Ruff 0.16 newly enabled it, causing 22 failures in existing APIs. Refactoring those signatures is outside this CI-fix PR and could be breaking.
- `pymatgen-core!=2026.7.24` excludes one broken release that lacks compatible Linux/Python distributions while still allowing future fixed versions

pyproject.toml
```diff
@@ -68,9 +68,12 @@ docs = [
 ]
 
 [tool.uv]
+constraint-dependencies = ["pymatgen-core!=2026.7.24"]
```
ndaelman-hu: This release was revoked, so there's no need for this line anymore: https://pypi.org/project/pymatgen-core/2026.7.24/

Moreover, I already confirmed that the builds can now be re-run on the current `develop` without breaking.

ndaelman-hu (approved):
The PR constraints some of Ruff's linting scope. I agree with most additions, but **some have become obsolete now** (not necessarily at the time of the patch). Those should be removed. No need for a follow-up review.

PR #440, opened 2026-07-30
Resolve free-form LibXC XC components; add unidentified marker; fix REVPBE alias

ndaelman-hu:
## Purpose

Let DFT parsers report exchange–correlation as the raw LibXC components a code
emits (individual exchange/correlation parts, by label or by LibXC id) without
each parser owning LibXC taxonomy lookups. The schema resolves those components
— filling `family`/`kind` so `DFT.jacobs_ladder` can be derived — and, when a
component cannot be resolved, keeps it as an explicitly marked placeholder
rather than dropping it. Also corrects the `REVPBE` alias, which collided with
`RPBE`.

## Scope

**Included**

- `XCComponent.normalize`: complete a directly-supplied component from the LibXC
  registry, keyed by either `canonical_label` or `libxc_id`; canonicalize the
  label (e.g. `gga_x_pbe` -> `XC_GGA_X_PBE`). `weight` is left untouched.
- New `XCComponent.unidentified` boolean: a component the registry cannot
  resolve (unknown label or id) is retained as a placeholder and marked
  `unidentified`, so a functional's composition is not misrepresented as
  complete (e.g. a two-part functional does not silently become exchange-only).
- Fix `REVPBE` alias: `XC_GGA_X_RPBE` -> `XC_GGA_X_PBE_R` (revPBE is Zhang–Yang).

**Out of scope**

- The parser changes that emit these components (separate PR in
  `nomad-parser-plugins-simulation`).

## Context & Links

- Schema-side enabler for the DFT XC parser migration; the parser PR depends on
  this and should merge after it.

## Reviewer Notes

- Two resolution keys: `canonical_label` and, as a fallback, `libxc_id` via
  `spec_from_id` — the latter lets codes that expose LibXC ids (e.g. abinit's
  negative `ixc`) hand the schema the id and let the registry, which is more
  complete than any parser-local table, resolve it.
- `unidentified` is set by the normalizer when neither key resolves; a parser
  may also set it directly for a slot it knows exists but cannot name at all
  (no label, no id). Such placeholders survive because they carry a set field —
  the normalizer classifies what reaches it but cannot rescue a component pruned
  for being empty.
- `weight` is deliberately not set by the fill — it is a composition fraction
  the parser owns, meaningful only for explicit mixes.
- The `REVPBE` fix changes what `REVPBE` expands to (previously identical to
  `RPBE`); output for revPBE calculations changes accordingly.

## Status

- [x] Ready for review

## Breaking Changes

- [x] None (`unidentified` is additive). Note: the `REVPBE` alias now expands
  differently — see Reviewer Notes.

## Dependencies / Blockers

- [x] None for this PR. The parser-side PR depends on it.

## Testing / Validation

- [x] Unit tests — `tests/test_model_method.py`: free-form resolution by label
  and by id (parametrized), `unidentified` marking of unresolvable label/id,
  placeholder retention/arity. 190 pass.
- [x] Manual testing — verified end-to-end that octopus/abinit/crystal/exciting
  produce resolved components with correct `family`/`kind` and `jacobs_ladder`.
- Docs regenerate with no diff.

### DIFF
--- a/src/nomad_simulations/schema_packages/model_method.py
+++ b/src/nomad_simulations/schema_packages/model_method.py
@@ -33,6 +33,7 @@ from nomad_simulations.schema_packages.data_types import (
 from nomad_simulations.schema_packages.model_system import ModelSystem
 from nomad_simulations.schema_packages.numerical_settings import NumericalSettings
 from nomad_simulations.schema_packages.utils.libxc.build import (
+    spec_from_id,
     spec_from_label,
 )
 from nomad_simulations.schema_packages.utils.libxc.expand import (
@@ -617,6 +618,51 @@ class XCComponent(ArchiveSection):
          """,
     )
 
+    unidentified = Quantity(
+        type=bool,
+        description="""
+        `True` when the code reports an XC component at this position that cannot
+        be resolved to a LibXC label (family/kind therefore stay unset). The
+        component is retained as a placeholder so the functional's composition
+        (the number of parts) is not misrepresented as complete.
+        """,
+    )
+
+    def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
+        super().normalize(archive, logger)
+
+        # Complete a free-form component from the LibXC registry, keyed by either
+        # the `canonical_label` or the `libxc_id` the code reports. `weight` is a
+        # composition fraction the parser owns, not registry data, so it is
+        # intentionally left untouched.
+        if self.family is not None:
+            return
+
+        spec = None
+        if self.canonical_label:
+            spec = spec_from_label(self.canonical_label)
+        elif self.libxc_id is not None:
+            spec = spec_from_id(int(self.libxc_id))
+
+        if spec is None:
+            # The code reported a component (a label or a LibXC id) the registry
+            # cannot resolve: keep it as a placeholder so it is not silently
+            # dropped, and mark it unidentified.
+            if self.canonical_label or self.libxc_id is not None:
+                logger.debug(
+                    'Unresolvable LibXC component.',
+                    label=self.canonical_label,
+                    libxc_id=self.libxc_id,
+                )
+                self.unidentified = True
+            return
+        self.libxc_id = spec['libxc_id']
+        self.display_name = spec['display_name']
+        self.family = spec['family']
+        self.kind = spec['kind']
+        # canonicalize to the registry spelling (e.g. 'gga_x_pbe' -> 'XC_GGA_X_PBE')
+        self.canonical_label = spec['canonical_label']
+
 
 class XCFunctional(ArchiveSection):
     """
@@ -681,6 +727,12 @@ class XCFunctional(ArchiveSection):
                     )
         except Exception:
             logger.warning('LibXC expansion failed.')
+
+        # Free-form composition: components supplied directly by a parser (raw
+        # LibXC labels the code reports) complete their own taxonomy. Components
+        # created by the name expansion above already carry it and no-op.
+        for comp in self.components or []:
+            comp.normalize(archive, logger)
 
         if self.global_exact_exchange is None:
             alphas = [
--- a/src/nomad_simulations/schema_packages/model_method.py
+++ b/src/nomad_simulations/schema_packages/model_method.py
@@ -641,7 +641,7 @@ class XCComponent(ArchiveSection):
         spec = None
         if self.canonical_label:
             spec = spec_from_label(self.canonical_label)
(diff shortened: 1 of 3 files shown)


src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -616,6 +618,51 @@ class XCComponent(ArchiveSection):
...
+        # the `canonical_label` or the `libxc_id` the code reports. `weight` is a
+        # composition fraction the parser owns, not registry data, so it is
+        # intentionally left untouched.
+        if self.family is not None:
+            return
+
+        spec = None
+        if self.canonical_label:
+            spec = spec_from_label(self.canonical_label)
+        elif self.libxc_id is not None:
```
EBB2675:
```suggestion
        if spec is None and self.libxc_id is not None:
```

EBB2675: When both fields are populated, the `libxc_id` lookup is skipped whenever `canonical_label` is present, even if the label lookup returns `None`. Since the PR describes the ID as a fallback, should we also try it after an unsuccessful label lookup?

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -616,6 +618,51 @@ class XCComponent(ArchiveSection):
...
+                )
+                self.unidentified = True
+            return
+        self.libxc_id = spec['libxc_id']
+        self.display_name = spec['display_name']
+        self.family = spec['family']
+        self.kind = spec['kind']
+        # canonicalize to the registry spelling (e.g. 'gga_x_pbe' -> 'XC_GGA_X_PBE')
+        self.canonical_label = spec['canonical_label']
+
```
EBB2675, line 669:
```suggestion
        self.unidentified = False
```

EBB2675: Could we explicitly clear `unidentified` after successful resolution? Otherwise, a component previously marked as unidentified can retain `unidentified=True` even after all registry fields have been populated

EBB2675 (approved):
Thank you very much

looks good to me, i have included two minor points

src/nomad_simulations/schema_packages/model_method.py
ndaelman-hu: Good catch — done in 9d971d2. Changed the `elif` to `if spec is None and self.libxc_id is not None:`, so a present-but-unresolvable `canonical_label` now falls back to the id before the component is marked `unidentified`.

src/nomad_simulations/schema_packages/model_method.py
ndaelman-hu: Done in 9d971d2. Cleared on successful resolution, guarded as `if self.unidentified: self.unidentified = False` so only a previously-marked placeholder is touched rather than stamping every resolved component.

src/nomad_simulations/schema_packages/model_method.py
```diff
@@ -617,6 +618,51 @@ class XCComponent(ArchiveSection):
...
+    def normalize(self, archive: 'EntryArchive', logger: 'BoundLogger') -> None:
+        super().normalize(archive, logger)
+
+        # Complete a free-form component from the LibXC registry, keyed by either
+        # the `canonical_label` or the `libxc_id` the code reports. `weight` is a
+        # composition fraction the parser owns, not registry data, so it is
+        # intentionally left untouched.
+        if self.family is not None:
+            return
```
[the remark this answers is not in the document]
ndaelman-hu: Declining this one: `family` and `kind` are only ever set together by the registry fill (`spec_from_label`/`spec_from_id`), and no parser maps `family` directly, so a component with `family` set but `kind` missing does not arise in practice. The guard is a safe idempotency check for the current flows.

src/nomad_simulations/schema_packages/utils/libxc/aliases.json
EBB2675, line 7:
```suggestion
    "VWN": ["XC_LDA_X", "XC_LDA_C_VWN"],
    "VWN-RPA": ["XC_LDA_X", "XC_LDA_C_VWN_RPA"],
```

ndaelman-hu: Applied in f314031 — added `"VWN-RPA": ["XC_LDA_X", "XC_LDA_C_VWN_RPA"]`. Paired with fhiaims mapping the VWN-RPA control string to `VWN-RPA` in PR #199 (490026c).

PR #441, opened 2026-07-30
clarify spin terminology

EBB2675:
## Purpose
Clarify spin terminology in the simulation schema by replacing the ambiguous `ModelSystem.total_spin` quantity with `ModelSystem.total_multiplicity`, which represents the spin multiplicity `2S+1`. This avoids conflating spin quantum number, doubled spin representations, and multiplicity.

## Scope

**Included**

- Rename `ModelSystem.total_spin` to `ModelSystem.total_multiplicity`.
- Update the quantity description to define multiplicity as the positive integer `2S+1`.
- Add a TODO noting that `AtomsState.spin` is legacy and needs a future migration/cleanup path.

**Out of scope**

- Redesigning atomic/electronic spin-state representation.
- Removing or changing `AtomsState.spin`.

## Status

- [x] Ready for review

## Breaking Changes

- [x] `ModelSystem.total_spin` was renamed to `ModelSystem.total_multiplicity`; no known parsers are currently affected

## Dependencies / Blockers

- [x] None


src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1478,13 +1478,14 @@ class ModelSystem(System, Representation):
...
+    total_multiplicity = Quantity(
         type=np.int32,
         description="""
-        Total spin quantum number **S** of the system (so Ŝ² ψ = S(S+1) ħ² ψ).
-        Stored as an integer or half-integer represented in doubled form
-        (e.g. singlet → 0, doublet → 1, triplet → 2).
-        Not to be confused with the spin multiplicity 2S+1.
+        Spin multiplicity of the total system, i.e. the value **2S+1**, where S is the
+        total spin quantum number. It describes the system-level total electronic state
+        and is always a positive integer:
```
ndaelman-hu: Pls modify the datatype to enforce this constraint.

src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1478,13 +1478,14 @@ class ModelSystem(System, Representation):
         """,
     )
 
-    total_spin = Quantity(
+    total_multiplicity = Quantity(
```
ndaelman-hu, line 1482: I thought we settled on also adding a conversion function for codes that report total spin instead.

ndaelman-hu (commented):
This PR introduces `total_multiplicity` to supercede `total_spin` at the system level.
I agree with the change, just 2 minor points.

src/nomad_simulations/schema_packages/model_system.py
EBB2675: yes, i noted it as a parser utility

ndaelman-hu: Maybe a utility actually isn't that useful. I had an agent (Opus 5) check the output formats across ab initio code and below are its findings. Since these do not always cleanly interconvert, you might want to extend the system-level descriptors to cover all cases @EBB2675.

---

Every code whose spin target state is actually parsed expresses it as the multiplicity (2S+1). What you find instead of multiplicity in other codes isn't S either — it's one of two different observables (⟨S²⟩ or net magnetization ≈ 2·Sₙ). Here's the breakdown across the ~43 electronic parsers:

Multiplicity, M = 2S+1 (stored as spin_target / spin_target_multiplicity / x_<code>_multiplicity):
Gaussian (Multiplicity = → spin_target), GAMESS (x_gamess_spin_target_multiplicity, docstring literally "M=2S+1, where S is the total spin"), CP2K, NWChem (Spin multiplicity), ORCA (x_orca_multiplicity), Psi4. CPMD refers to it conceptually ("minimum spin multiplicity, i.e. doublet or singlet").

⟨S²⟩ — the expectation value (equals S(S+1) for a pure state; neither S nor 2S+1), stored as spin_S2:
Gaussian, NWChem (<S^2> =), MOPAC ((S**2) =), Psi4 (s2_observed).

Net magnetization ≈ 2·S_z (n↑ − n↓, in Bohr magnetons) — the periodic DFT codes:
Quantum ESPRESSO (total magnetization … Bohr mag/cell), VASP, ABACUS (tot_magnetization), Octopus (total_magnetic_moment), GPAW. GPAW is the one code that then converts this to an explicit S_z: x_gpaw_fixed_spin_Sz = magnetic_moments.sum() / 2.0.

EBB2675: thanks, as mentioned, i will revisit this next week in my parser sprint

src/nomad_simulations/schema_packages/model_system.py
EBB2675: done

src/nomad_simulations/schema_packages/model_system.py
ndaelman-hu: Note that this was a schema point, though. Just open it as an issue if you want to tackle it separately.

PR #438, opened 2026-07-29
Update ruff ignores

ladinesa:
## Purpose
Update ruff ignores list and modify files as prescribed.

## Scope

**Included**

- 

**Out of scope**

-

## Context & Links

#430 

- 

## Reviewer Notes

none
## Status

- [ ] Draft / In progress
- [x] Ready for review
- [ ] Blocked (explain):

## Breaking Changes

- [x] None
- [ ] (explain):

## Dependencies / Blockers

- [x] None
- [ ] Open design questions (explain):
- [ ] External dependencies (explain):

## Testing / Validation

_How was this change validated?_

- [] None (explain):
- [ ] Unit tests
- [x] Integration tests
- [ ] Manual testing (explain):

PR #399, opened 2026-06-10
MOs semantic cleanup

EBB2675:
## Purpose

`MolecularOrbitals` had no canonical location in `Outputs` and inherited the k-resolved `ElectronicEigenvalues` schema while separately defining molecular-orbital quantities.

This PR makes molecular orbitals a first-class output and models them as an independent `PhysicalProperty` with MO-specific dimensions, quantities, and validation.

## Scope

**Included**

- Export `MolecularOrbitals` from `properties/__init__.py`.
- Add `Outputs.molecular_orbitals` as a repeating `SubSection`.
- Change `MolecularOrbitals` from `ElectronicEigenvalues` inheritance to an independent `PhysicalProperty`.
- Keep explicit, strictly positive dimensions:
  - `n_mo`: number of molecular orbitals.
  - `n_ao`: number of atomic orbitals.
- Define per-orbital quantities with shape `['n_mo']`:
  - `orbital_energies`
  - `orbital_occupations`
  - `role`
  - `symmetry`
- Represent spin-polarized results using one `MolecularOrbitals` section per `spin_channel`.
- Rename MO-specific quantities:
  - `mo_energies` -> `orbital_energies`
  - `mo_occupations` -> `orbital_occupations`
  - `mo_spin` -> section-level `spin_channel`
  - `mo_coefficients` -> `coefficients`
  - `mo_coefficients_im` -> `coefficients_im`
  - `mo_class` -> `role`
  - `mo_symmetry` -> `symmetry`
  - `mo_type` -> `kind`
- Store real and imaginary AO-to-MO coefficient matrices as `HDF5Dataset` quantities with expected shape `(n_mo, n_ao)`.
- Infer dimensions during normalization:
  - Infer `n_mo` from a valid coefficient matrix first, then from `orbital_energies`, `orbital_occupations`, `role`, or `symmetry`.
  - Infer `n_ao` from a valid coefficient matrix.
- Log normalization errors for:
  - Invalid coefficient rank.
  - Coefficient shape mismatches.
  - Mismatched real and imaginary coefficient shapes.
  - Per-orbital lengths inconsistent with `n_mo`.
- Regenerate schema documentation and update the related tests.

## Reviewer Notes

- `MolecularOrbitals` intentionally does not inherit electronic-eigenvalue-derived quantities such as `highest_occupied`, `lowest_unoccupied`, or `band_gap`.
- Molecular systems do not acquire a synthetic k-point dimension; per-orbital arrays use `['n_mo']`.
[...]

PR #444, opened 2026-08-03
QC1: `MolecularOrbitals` semantic and scientific cleanup

EBB2675:
## Purpose

`MolecularOrbitals` had no canonical location in `Outputs` and inherited the k-resolved `ElectronicEigenvalues` schema while separately defining molecular-orbital quantities.

This PR makes molecular orbitals a first-class output and models them as an independent `PhysicalProperty` with MO-specific dimensions, quantities, and validation.

## Scope

**Included**

- Export `MolecularOrbitals` from `properties/__init__.py`.
- Add `Outputs.molecular_orbitals` as a repeating `SubSection`.
- Change `MolecularOrbitals` from `ElectronicEigenvalues` inheritance to an independent `PhysicalProperty`.
- Keep explicit, strictly positive dimensions:
  * `n_mo`: number of molecular orbitals.
  * `n_ao`: number of atomic orbitals.
- Define per-orbital quantities with shape `['n_mo']`:
  * `energies`
  * `occupations`
  * `role`
  * `symmetry`
- Classify the whole orbital set with `kind` (`canonical`, `natural`, `localized`).
- Represent spin-polarized results using one `MolecularOrbitals` section per `spin_channel`.
- Resolve frontier orbitals for `kind=canonical` during normalization: `homo`, `lumo`, and `homo_lumo_gap` (all in joule), computed from `energies` and `occupations`, and not overwritten if already set.
- Rename MO-specific quantities:
  * `mo_energies` -> `energies`
  * `mo_occupations` -> `occupations`
  * `mo_spin` -> section-level `spin_channel`
  * `mo_coefficients` -> `coefficients`
  * `mo_coefficients_im` -> `coefficients_im`
  * `mo_class` -> `role`
  * `mo_symmetry` -> `symmetry`
  * `mo_type` -> `kind`
- Store real and imaginary AO-to-MO coefficient matrices as `HDF5Dataset` quantities with expected shape `(n_mo, n_ao)`.
- Infer dimensions during normalization:
  * Infer `n_mo` from a valid coefficient matrix first, then from `occupations`, `role`, `symmetry`, or `energies`.
  * Infer `n_ao` from a valid coefficient matrix.
- Log normalization errors for:
  * Invalid coefficient rank.
  * Coefficient shape mismatches.
  * Mismatched real and imaginary coefficient shapes.
  * Per-orbital lengths inconsistent with `n_mo`.
  * `spin_channel` set to a value other than 0 or 1.
  * `occupations` outside the range allowed for the spin representation (0 to 1 with `spin_channel` set, 0 to 2 otherwise), including negative values.
  * `n_mo` greater than `n_ao`.
- Regenerate schema documentation and update the related tests.

## Reviewer Notes

- `MolecularOrbitals` intentionally does not inherit electronic-eigenvalue-derived quantities such as `highest_occupied`, `lowest_unoccupied`, or `band_gap`. It defines its own molecular `homo`, `lumo`, and `homo_lumo_gap`, resolved only for canonical orbitals.
- Molecular systems do not acquire a synthetic k-point dimension; per-orbital arrays use `['n_mo']`.
- Spin channels are represented by separate repeating sections rather than a per-orbital spin array.
- `n_mo` and `n_ao` must be greater than zero, and `n_mo` may not exceed `n_ao`.

## Status

- [x] Ready for review

## Breaking Changes

- [x] `MolecularOrbitals` is now stored under `Outputs.molecular_orbitals`.
- [x] `MolecularOrbitals` no longer inherits from `ElectronicEigenvalues`.
- [x] `n_mo` and `n_ao` no longer accept `0`.
- [x] Orbital energy units change from electron volts to joules.
- [x] Quantities were removed or renamed:

| Previous             | Replacement                    |
| -------------------- | ------------------------------ |
| `mo_energies`        | `energies`                     |
| `mo_occupations`     | `occupations`                  |
| `mo_spin`            | `spin_channel` on each section |
| `mo_coefficients`    | `coefficients`                 |
| `mo_coefficients_im` | `coefficients_im`              |
| `mo_class`           | `role`                         |
| `mo_symmetry`        | `symmetry`                     |
| `mo_type`            | `kind`                         |

## Dependencies / Blockers

- The ORCA parser is not merged yet and must be updated to use this schema.


src/nomad_simulations/schema_packages/properties/molecular_orbitals.py
```diff
@@ -1,220 +1,291 @@
...
+        if not occupied.any() or occupied.all():
+            return
+        homo = self.energies[occupied].max()
+        lumo = self.energies[~occupied].min()
+        if self.homo is None:
+            self.homo = homo
+        if self.lumo is None:
+            self.lumo = lumo
+        if self.homo_lumo_gap is None:
+            self.homo_lumo_gap = lumo - homo
```
ndaelman-hu: `homo_lumo_gap` is computed from the local derived `homo`/`lumo`, but `self.homo`/`self.lumo` are only filled when unset -- so a preset frontier value yields a stored triple where `gap != lumo - homo`. Example: a parser presets `homo=-3.0`; derived `homo=-1.0`/`lumo=0.5` -> stored `homo=-3.0, lumo=0.5, gap=1.5`, while `lumo - homo` of the stored values is `3.5`, contradicting this field's docstring. Derive the gap from the resolved `self.homo`/`self.lumo`, and only when the gap is itself being derived -- do not recompute a parser-provided gap, since a code may define it differently.

ndaelman-hu: Concretely: leave any parser-provided `homo`/`lumo`/`homo_lumo_gap` untouched, and emit the schema's own derivation as a separate `is_derived` result rather than filling these fields in place -- the way `ElectronicEigenvalues.extract_band_gap()` produces an `ElectronicBandGap(is_derived=True)`.

Why: as written, if a parser set `homo`/`lumo` but not the gap, `homo_lumo_gap` is computed from the freshly derived `homo`/`lumo`, not the stored ones, so it won't equal `self.lumo - self.homo`. And more generally, a code may report HOMO/LUMO/gap under its own definition -- the schema shouldn't silently overwrite them or force them to agree. Keeping the parsed values as-is and marking our derivation `is_derived` keeps the code's numbers and ours from contradicting each other.

src/nomad_simulations/schema_packages/properties/molecular_orbitals.py
```diff
@@ -1,220 +1,291 @@
...
+                'The real and imaginary coefficient matrices have different shapes and cannot be combined.'
             )
 
+        _TOL = 1e-6
+        if self.spin_channel is not None and self.spin_channel not in (0, 1):
+            logger.error('`spin_channel` must be 0 (alpha) or 1 (beta) when set.')
+        if self.occupations is not None:
+            occ = np.asarray(self.occupations)
+            upper = 1.0 if self.spin_channel is not None else 2.0
+            if np.nanmin(occ) < -_TOL:
```
ndaelman-hu: Empty `occupations` crashes instead of logging: `np.nanmin([])` raises (n_mo preset), or inferred `n_mo=0` at L191 trips `strictly_positive_int`. Guard zero-length arrays.

src/nomad_simulations/schema_packages/properties/molecular_orbitals.py
```diff
@@ -1,220 +1,291 @@
...
-from nomad.metainfo import URL, MEnum, Quantity, Reference, SectionProxy
+from nomad.metainfo import MEnum, Quantity, Reference, SectionProxy
 
-from nomad_simulations.schema_packages.properties import ElectronicEigenvalues
+from nomad_simulations.schema_packages.data_types import strictly_positive_int
+from nomad_simulations.schema_packages.physical_property import PhysicalProperty
 
+# Occupation below this threshold is treated as an unoccupied orbital when
+# resolving frontier (HOMO/LUMO) orbitals.
+_OCCUPATION_TOL = 1e-6
```
ndaelman-hu, line 23: `_OCCUPATION_TOL` and `_TOL` (L210) are both `1e-6` for the same concept. `ElectronicEigenvalues` uses the configurable `configuration.occupation_tolerance` -- reuse it.

src/nomad_simulations/schema_packages/properties/molecular_orbitals.py
```diff
@@ -1,220 +1,291 @@
...
-    provides auxiliary metadata needed by NOMAD tooling.
+    `spin_channel` selects the representation: set (0=alpha, 1=beta) gives spin
+    orbitals with occupations in [0,1]; unset gives spin-summed spatial orbitals
+    with occupations in [0,2]. `coefficients`, `energies`, `occupations`, `role`,
+    `symmetry` describe the same `n_mo` orbitals. `energies` defined only for
+    `kind=canonical`. `n_mo` is at most `n_ao`.
 
-    The TREXIO format:
-    Posenitsky et al., J. Chem. Phys. 158, 174801 (2023).
+    PhysicalProperty.value is intentionally unused here; pending refactor to a shared base class.
```
ndaelman-hu: Every other `PhysicalProperty` defines `value`; this class leaves it unset (its docstring calls that a stopgap). Any generic code that reads `value` -- plotting today, possibly search/table rendering later -- silently does nothing for molecular orbitals. And it's not obvious what `value` should be: `energies` is the intuitive pick but is canonical-only (absent for natural/localized), whereas `occupations` is defined for every `kind`. That needs a deliberate choice -- please resolve it now or open a tracking issue.

src/nomad_simulations/schema_packages/properties/molecular_orbitals.py
```diff
@@ -1,220 +1,291 @@
...
+            logger.error(
+                '`n_mo` exceeds `n_ao`, which is physically inconsistent: the MO space cannot be larger than the AO basis it is expanded in.'
+            )
+
+        self._resolve_homo_lumo()
+
+    def _resolve_homo_lumo(self) -> None:
+        # Frontier orbitals are well defined only for canonical orbitals with
+        # both energies and occupations present and consistent in length.
+        if self.kind not in (None, 'canonical'):
```
ndaelman-hu: Are you sure you want to treat unlabeled orbitals (`kind=None`) as canonical here? An untagged set could just as well be natural or localized, where HOMO/LUMO isn't defined. At least gate on the occupations -- canonical orbitals have (near-)integer fillings, so fractional occupations are a signal to skip rather than assume canonical.

tests/properties/test_molecular_orbitals.py
```diff
@@ -54,71 +65,191 @@ def archive_with_mo() -> Generator[
...
+        assert not isinstance(molecular_orbitals, ElectronicEigenvalues)
+
+    def test_stored_in_outputs(self):
+        """MolecularOrbitals can be stored through the canonical Outputs path."""
+        mo = MolecularOrbitals()
+        outputs = Outputs(molecular_orbitals=[mo])
+        simulation = Simulation(outputs=[outputs])
+        assert simulation.outputs[0].molecular_orbitals[0] is mo
+
+    def test_normalize_infers_dimensions_from_coefficients(self, archive_with_mo):
```
ndaelman-hu, line 82: No test exercises `homo`/`lumo`/`homo_lumo_gap` or the `n_mo > n_ao` validation. Most usefully, add a case that sets `homo`/`lumo` before `normalize()` but leaves the gap unset, then asserts the stored `homo_lumo_gap` equals `self.lumo - self.homo` -- that's the frontier-consistency issue flagged above, and it currently has zero coverage.

ndaelman-hu (changes requested):
Thank you for disentangling `MolecularOrbitals` from the eigenvalues lineage -- so the k-points don't mingle with the molecular orbitals. I'll use this as a foundation to build a better eigenvalues -> band-structure pipeline on top of it.

In that spirit, most of my comments are nudges toward a single shared way of deriving the frontier quantities (HOMO/LUMO/gap) that `MolecularOrbitals`, `ElectronicEigenvalues`, and DOS can all reuse, rather than three subtly different versions. A couple are concrete fixes I'd like before merge, and a few are open questions rather than asks -- all inline. Solid step in the right direction; glad to pair on the standardization follow-up.

ndaelman-hu:
The inheritance concern from my #399 review is addressed here — MOs are now a standalone `PhysicalProperty` rather than inheriting `ElectronicEigenvalues`. From my review on this PR, the blockers are small and shouldn't hold it up: the `homo_lumo_gap` consistency fix (derive from `self.homo`/`self.lumo`), the empty-`occupations` guard, and setting `value`. The rest of my notes are non-blocking. Won't merge until those three clear.

ndaelman-hu (approved):
The changes I requested have been worked in:

- `value` now holds the orbital energies (eigenvalues), mirroring `ElectronicEigenvalues` — no longer left unset.
- The HOMO-LUMO gap is derived from the stored frontier pair, so it can't disagree with its endpoints; the frontier fields are split into provenance `_parsed` and purely derived `_normalized`, with no fallback.
- `kind` must be explicitly `canonical` to resolve frontier orbitals; an unset `kind` is no longer assumed canonical.
- Occupation bounds moved onto an interval datatype, and the duplicated tolerance constant is gone.

Approving on that basis.

**Before merge:** resolve the merge conflict in `molecular_orbitals.py` — it's mechanical, from `develop` trimming unused imports (#438) while this branch rewrote the file. Take this branch's version; the import cleanup is already subsumed.

Issue #182, opened 2025-05-06
Assess code comments throughout


JFRudzinski:

ndaelman-hu:
I suppose that this is the issue targeting docstrings, as referred to here: https://github.com/nomad-coe/nomad-simulations/pull/170#discussion_r2076993067

Anyhow, as that review comment covered a more general aspect, I'll see to rephrase it here:

Methods with the return type `Optional[SomeOtherType]` where `SomeOtherType` also has a typical "failure" term (e.g. `List: []`, `Str: ""`, `Bool: False`), should clarify the distinction between `None` and the alternative failure mode. Distinction here means "the various cases that lead up to each".

In many cases, `None` seems to be default for anything that triggers a warning or error log call. If this covers its whole semantics, it may easily be removed in lieu of the alternative failure mode. **This will keep the code leaner and more manageable.**
Of course, there may be cases where the method directly writes to the archive. There `None` already has its own semantics.

PR #445, opened 2026-08-03
QC2: add `ModelSystem.total_multiplicity` and clarify spin terminology

### LINKED ISSUE #434: simplify & clarify spin-related quantities
Spin is represented inconsistently across the schema:

* `AtomsState.spin` is an `int32` described as total spin (S), defaults unknown spin to (S=0), and and has an unclear relationship to the structured spin information under `electronic_state.spin_orbit_state`.
* `ModelSystem.total_spin` is described as (S), but stores the doubled value (2S).
* `BSDFT.total_spin_projection` stores (2M_S) (This is a valid integer encoding for method dependent spin, not necessarily a problem)

### I suggest

- Remove `AtomsState.spin` and represent atomic spin states through `electronic_state.spin_orbit_state` 
- Either clarify that `ModelSystem.total_spin` stores (2S), or replace it with an explicitly named `total_spin_multiplicity` storing (2S+1) (I am leaning towards the latter)

EBB2675:


src/nomad_simulations/schema_packages/model_system.py
```diff
@@ -1478,13 +1479,15 @@ class ModelSystem(System, Representation):
         """,
     )
 
-    total_spin = Quantity(
-        type=np.int32,
+    total_multiplicity = Quantity(
```
JFRudzinski: I thought you said it was going to be total_spin_multiplicity?

EBB2675: it is already the case in `qc-super`

ndaelman-hu:
Needs a naming decision before merging. My earlier review points from #441 are in — `total_multiplicity` uses `strictly_positive_int` and the description is reworded — and the descriptor-coverage point I raised there is now tracked in #465. The open item is the name itself: QC2 adds `total_multiplicity`, but `qc-super` and the ORCA parser (#182) use `total_spin_multiplicity`, so if this merges as-is #182 breaks when it re-pins to `develop`. Plan is to fold `total_spin_multiplicity` into this PR so `develop` matches the parser; that also leaves room for the #465 sibling descriptors (`⟨S²⟩`, magnetization) to read as one family.

